PR #5, opened 2017-05-22
Initial checkin of REST Server

Terris-ScaleDB:


Terris-ScaleDB:
There is no connection to users in Postgres yet

src/python/foglamp/rest_server/README.md
```diff
@@ -0,0 +1,79 @@
+# Introduction
```
praveen-garg: move to `docs` as `.rst` (not .md)

Terris-ScaleDB: OK, we'll have README.md point to readthedocs. I don't know .rst

PR #11, opened 2017-05-24
Fogl 11 ashwin asyniounittest

ashwinscale:
https://scaledb.atlassian.net/browse/FOGL-11
Unit test for render_post.


ashwinscale:
I've removed Ori's changes, he is now operating under a separate branch. 
The test has been moved to  src\python\tests\coap\uri_handlers\test_sensor_values.py
The change to coap_service has not been made. 
> Configurator is light and has its own tests. This doesn't need to be changed but unless something uses a heavyweight resource, I wouldn't worry about mocking it.
Mocking is not only about heavy weight dependencies, it's about external dependencies in general. In this case, Configurator object need not be created, and sometimes CANNOT be created unless the file is accessible (an error I ran into while trying to develop the test).

PR #14, opened 2017-05-25
FOGL-76

praveen-garg:
- copy yaml file in build.sh; 
- use postgres db password and jwt secret from env var if set else from yaml; 
- provided export.sh use `source foglamp/export.sh -s` to set or `source foglamp/export.sh -u` to unset


Terris-ScaleDB:
@praveen-garg , You can merge if you don't have any change requests

praveen-garg:
@Terris-ScaleDB done!

PR #15, opened 2017-05-25
adding makefile for cleanup; added its use in build.sh

praveen-garg:
FOGL-83

run `make help` to see usage
`build.sh -c` calls `make clean`


praveen-garg:
who are interested in (or new to) .PHONY should read https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html

PR #21, opened 2017-05-30
FOGL-39: logging async via postgres

amarendra-ScaleDB:
[FOGL-39](https://scaledb.atlassian.net/browse/FOGL-39)
This PR contains first pass of below tasks:

- Log to a table in Postgres. Each logged message has a 'code', a severity, and a message
- Define logging levels
- Tests added for postgres log
- "test" env now enabled


src/python/foglamp/foglamp_logger.py
```diff
@@ -0,0 +1,42 @@
...
+
+logging.Logger.sensor1 = sensor1
+logging.Logger.sensor2 = sensor2
+
+foglamp_logger = logging.getLogger(__name__)
+foglamp_logger.setLevel(logging.DEBUG)
+
+foglamp_log_formatter = logging.Formatter('[%(asctime)s] %(levelname)s: %(name)s: %(message)s [%(filename)s/%(lineno)d %(module)s/%(funcName)s]')
+
+foglamp_filehandler = logging.FileHandler(FOGLAMP_DIR+'/fog.log')
```
praveen-garg: should be configurable.. i.e. we may want: coap.log; rest.log, admin.log etc..

src/python/foglamp/postgres_handler.py
```diff
@@ -0,0 +1,82 @@
...
+    , sa.Column('created_by', sa.types.VARCHAR(50)))
+
+def log_msg(record):
+    """Construct log message from logging record components"""
+
+    msg = record.msg
+    msg = msg.strip()
+    msg = msg.replace('\'', '\'\'')
+
+    return msg
```
praveen-garg: there must be some way to dump as json

src/python/foglamp/foglamp_logger.py
Terris-ScaleDB: Files under the foglamp package should not contain the word 'foglamp' because it is redundant.

Terris-ScaleDB: There will be potentially millions of sensors. We can't filter this way. The coap server would use some logger name that is hard-coded and would pass a dictionary to info/error/fatal/etc. which would be used by the logging module. The keys need to be defined.

src/python/foglamp/postgres_handler.py
Terris-ScaleDB: All postgres access should be moved to a module in foglamp/model package

See https://scaledb.atlassian.net/browse/FOGL-128

amarendra-ScaleDB:
Since logging requirements are still to be finalised and it seems that that more work will be needed in this task. The utility of this PR is more a kind of POC. Hence closing this PR.

PR #23, opened 2017-06-06
FOGL-81 - pip requirements are now separated with their versions and tests call via make

ashish-ScaleDB:
[FOGL-81](https://scaledb.atlassian.net/browse/FOGL-81)

Fixed items:
1) requirements - basic python dependencies
2) requirements_doc - only docs dependencies
3) requirements_test - only test dependencies
4) requirements_dev - python dependencies + test dependencies + docs dependencies
5) Now we should run our _doc and python_ tests by calling **make**
6) pip packages with their versions as standard process.


praveen-garg:
@Terris-ScaleDB LGTM!

PR #24, opened 2017-06-06
FOGL-41 adding template file and license.txt

praveen-garg:
[FOGL-41](https://scaledb.atlassian.net/browse/FOGL-41)

**TODO:** 

- [ ] ~$~**{**{FOGLAMP_LICENSE_DESCRIPTION}**}** in `__template__.py`
- [ ] setup.py  Need author_email  `?@osisoft.com`
- [ ] docs/conf.py

**author and version:** both can be set as template variable Or need to be put manually as needed.

`__author__ = ${FULL_NAME}`

`__version__ = ${VERSION}` **Not sure if we need this?!**


Terris-ScaleDB:
Please add Mark as reviewer. I am not familiar with ${ } .. what replaces those?

__template__.py
```diff
@@ -0,0 +1,311 @@
...
+    * You have to also use ``sphinx.ext.todo`` extension
+
+.. _Google Python Style Guide:
+   http://google.github.io/styleguide/pyguide.html
+
+"""
+
+import sys
+
+__author__ = ${FULL_NAME}
```
MarkRiddoch: Was your  thought to add a git checkin filter that substitutes these in first checkin?

__template__.py
```diff
@@ -0,0 +1,311 @@
+# -*- coding: utf-8 -*-
+"""
+${FOGLAMP_LICENSE_DESCRIPTION}
+
+Homepage and documentation: http://foglamp.readthedocs.io/
```
MarkRiddoch: I would be tempted to rephrase this as

See: http://fogloamp.readthedocs.io/

This gives us some flexibility if we want to make a separate homepage for the product

__template__.py
praveen-garg: no; I think I was not able to explain in PR description.

PyCharm allows File and Code Templates; so we can set these placeholders with template variables and have snippets to insert.

> author and version: both can be set as template variable 
> 
> __author__ = ${FULL_NAME}
> __version__ = ${VERSION}
> 

Or need to be put (replace) manually

praveen-garg:
@MarkRiddoch @izoratti @billehunt `{{FOGLAMP_LICENSE_DESCRIPTION}}` still need to be placed in this.

Terris-ScaleDB:
I don't think we're going to be able to specify the exact boilerplate now. The boilerplate's first and last lines should contain tags so a script can later change them en masse.  For example:

"""
BEGIN
(c) 2017 ..

END
"""

I would not actually use BEGIN and END.

Terris-ScaleDB:
Will the boilerplate appear in the generated docstring documentation for the module? That would not be ideal.

Terris-ScaleDB:
When a new is created, what is the process for updating the version strings en masse?

praveen-garg:
Fixed in https://github.com/foglamp/FogLAMP/pull/35

PR #25, opened 2017-06-08
Modified unit test for sensor values to use pytest-mock

vaibhav-ScaleDB:
Benefits of using pytest-mock over unittest mock:
1. No unnecessary use of decorators
2. No passing of arguments in the test method and hence reducing the complexity of remembering the order of parameters passed wrt mock definitions.
3. Return value can be specified in mock itself.

For further reading please refer: https://github.com/pytest-dev/pytest-mock#why-bother-with-a-plugin

### DIFF
--- a/src/python/requirements_test.txt
+++ b/src/python/requirements_test.txt
@@ -1,3 +1,4 @@
 pylint==1.7.1
 pytest==3.1.1
+pytest-mock==1.6.0
 pytest-allure-adaptor==1.7.7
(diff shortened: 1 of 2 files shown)


Terris-ScaleDB:
Please also see 
https://asvetlov.github.io/aiohttp-bilbao/#/6/7
https://asvetlov.github.io/aiohttp-bilbao/#/6/8

Terris-ScaleDB:
I don't know how to merge this into 'develop'. Need Praveen's help.

Terris-ScaleDB:
I think develop needs to be merged into FOGL-23_vaibhav and then pushed.

PR #30, opened 2017-06-13
default branch name changed to develop in Jenkinsfile

ashish-ScaleDB:
[FOGL-159](https://scaledb.atlassian.net/browse/FOGL-159)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -13,8 +13,8 @@ node {
          name: 'repo_url'
        ),
        stringParam(
-         defaultValue: 'master',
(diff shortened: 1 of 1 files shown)


praveen-garg:
@ashish-ScaleDB This should be  (I think this is ) configurable on Jenkins server. so we should be change their directly. However I did it, as we want this to be built daily with develop

ashish-ScaleDB:
This change is only as we need to built on daily basis with develop branch and yes you can configure with  build parameters on CI server and also be make sure to check **Branch Specifier** under configure pipeline settings.

PR #51, opened 2017-06-28
FOGL-180 

praveen-garg:
[FOGL-180](https://scaledb.atlassian.net/browse/FOGL-180) basic rest server with debugger + cors (both in commented blocks); endpoint: foglamp/ping;


praveen-garg:
@MarkRiddoch Please review. Added psutil to make  cross-platform ps check. Thanks @amarendra-dianomic

src/python/foglamp_daemon.py
```diff
@@ -58,6 +58,47 @@ def safe_makedirs(directory):
...
 
+def find_process_start_time(process_name):
+    """
+    Find Process start time
+
+    :param process_name: name of the process
+    :return: string start_time or None if no such process exists
+    """
+    f_ps = subprocess.Popen("ps h -C " + process_name + " -o lstart", shell=True, stdout=subprocess.PIPE)
+    start_time = f_ps.stdout.read().decode("utf-8")[:-1] # remove last \n character
```
MarkRiddoch: OK, this is one way to do it. I was thinking more along the lines of getting the time at startup and saving it. Then get the time now and do a simple subtraction.

src/python/foglamp/rest_service/routes.py
```diff
@@ -0,0 +1,44 @@
...
+    })
+
+    # Configure CORS on all routes.
+    for route in list(app.router.routes()):
+        cors.add(route)
+
+
+def enable_debugger(app):
+    """ provides a debug toolbar for server requests """
+    import aiohttp_debugtoolbar
```
MarkRiddoch: What does this do for us?

src/python/foglamp/rest_service/routes.py
```diff
@@ -0,0 +1,44 @@
...
+    app.router.add_route('GET', '/foglamp/ping', api.ping)
+
+    # enable cors only in dev mode ?!
+    enable_cors(app)
+    # a live debugger (watcher) for requests
+    # this will neutralize error middleware
+    enable_debugger(app)
+
+
+def enable_cors(app):
```
MarkRiddoch: This is needed because of the debugtoolbar stuff or is it for another reason?

praveen-garg: This allows us to access the api from the requests which are not from the same origin. https://en.wikipedia.org/wiki/Same-origin_policy 
For frontend, Once we agree that nginx is good to go for serving static files and as a reverse proxy to api calls. This can be made disabled (but we may may want to keep it in dev mode, as is.) For the api to api communication, we may want to keep this enabled in production mode too, but for specific ports and IP (not *).

src/python/foglamp/rest_service/routes.py
praveen-garg: This allows us to watch requests being made to the rest (aiohttp) server and profiling.. we need this in dev mode only.
aiohttp_debugtoolbar
ping_response

src/python/foglamp_daemon.py
praveen-garg: Yupe, that was my initial plan too. However this is linux specific, and does not work on darwin (mac). So we are using psutil py package to make life easier and that will allow to scale monitoring with ease.

praveen-garg:
Ready for review and merge! @MarkRiddoch @Terris-Dianomic

praveen-garg:
@Terris-Dianomic ready for review again!

PR #58, opened 2017-07-07
Fogl-240 Daemon/core/device service refactor

Terris-Dianomic:
* Remove foglamp/env.py and therefore the PyYAML dependency
"* build.sh start" starts the daemon (no more run or daemon commands)
* Add \_\_main\_\_.py files so that python -m foglamp, python -m foglamp.device, and python -m foglamp.core work
* "foglamp" script runs the daemon
* The foglamp daemon only runs the core service -- not device or the deprecated "admin api"
* Remove psutil dependency from core. Store start time in local variable instead.
* build.sh no longer says "Virtual environment is already active" when using multiple commands like --install -a


src/python/foglamp/core/core_daemon.py
```diff
@@ -179,44 +160,64 @@ def get_pid():
...
-                start()
-            elif 'stop' == sys.argv[1]:
-                stop()
-            elif 'restart' == sys.argv[1]:
-                restart()
-            elif 'status' == sys.argv[1]:
-                print(is_running())
-            elif 'info' == sys.argv[1]:
-                print(get_pid())
+        if 'start' == sys.argv[1]:
```
praveen-garg: out of this PR's scope, but just for note.. we need to check `sys.argv` vs `argparser`

src/python/foglamp/core/core_daemon.py
```diff
@@ -28,68 +32,38 @@
...
-def run():
-    """
-    Run foglamp server in foreground
-    """
-
-    if is_running():
-        message = "Daemon already running. Check PID %s.\n"
-        sys.stderr.write(message % get_pid())
-        return is_running()
+    global _logger_configured
```
praveen-garg: this should be just in other PR. skip the logging until we are decided 👎

src/python/foglamp/core/core_daemon.py
```diff
@@ -5,19 +5,23 @@
...
 import signal
 import sys
 import time
 import daemon
 from daemon import pidfile
-import psutil
 
 
-from foglamp.controller import start as start_controller
+import foglamp.core.server
```
praveen-garg: from  foglamp.core import server

src/python/foglamp/core/api.py
```diff
@@ -6,25 +6,21 @@
...
     """
 
-    # Since foglamp can be started in foreground or as a daemon,
-    # need to check for both foglampd and foglamp
-    process_info = server.find_process_info('foglampd') or server.find_process_info('foglamp')
-
-    since_started = 0
-    if process_info is not None:
-        since_started = time.time() - process_info['start_time']
+    since_started = time.time() - __start_time
```
praveen-garg: are we relying on our code start time then checking the real process?!

praveen-garg: we can keep this, but let's have real code to check named process from the system

src/python/foglamp/core/__main__.py
```diff
@@ -0,0 +1,12 @@
...
+
+# FOGLAMP_BEGIN
+# See: http://foglamp.readthedocs.io/
+# FOGLAMP_END
+
+import foglamp.core.server
+
+
+if __name__ == "__main__":
+    foglamp.core.server.start()
```
praveen-garg: missing a new line at EOF

src/python/foglamp/core/__main__.py
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+# FOGLAMP_BEGIN
+# See: http://foglamp.readthedocs.io/
+# FOGLAMP_END
+
+import foglamp.core.server
```
praveen-garg: from  foglamp.core import server.

again, either have it or just server.py

vaibhav-ScaleDB:
Fix tests:
  1. Since we don't have environment file now, its safe to delete `tests/test_db_config` which was just checking the postgreSQL connection url
  2. In `test_Sensor_values.py` replace `from foglamp.device_api.coap.sensor_values import SensorValues` with `from foglamp.device.coap.sensor_values import SensorValues`

Environmental fixes:
  1. On a clean environment, `source build-sh -s` fails, reason : `-bash: foglamp: command not found` . You need to register foglamp first before starting. Call `make install` before `foglamp start`

src/python/foglamp/admin_api/server.py
```diff
@@ -0,0 +1,15 @@
...
+
+import asyncio
+from foglamp.admin_api import controller as admin_api_controller
+ 
+ 
+def start():
+    """Starts the service"""
+    admin_api_controller.start()
+    asyncio.get_event_loop().run_forever()
+
```
amarendra-dianomic: "asyncio.get_event_loop().run_forever()" appears twice, once each in admin_api/server.py and device/server.py. I think, it should appear in device/server.py only.

src/python/foglamp/device/server.py
```diff
@@ -0,0 +1,25 @@
...
+__author__    = "Terris Linenbach"
+__copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
+__license__   = "Apache 2.0"
+__version__   = "${VERSION}"
+
+
+def start():
+    """Starts the device service"""
+    coap_controller.start()
+    asyncio.get_event_loop().run_forever()
```
amarendra-dianomic: "asyncio.get_event_loop().run_forever()" appears twice, once each in admin_api/server.py and device/server.py. I think, it should appear in device/server.py only.

src/python/foglamp/admin_api/__main__.py
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+# FOGLAMP_BEGIN
+# See: http://foglamp.readthedocs.io/
+# FOGLAMP_END
+
+import foglamp.admin_api.server
```
amarendra-dianomic: from foglamp.admin_api import server

amarendra-dianomic:
Do we still need admin_api/wsgi.py ?

src/python/foglamp/__main__.py
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+# FOGLAMP_BEGIN
+# See: http://foglamp.readthedocs.io/
+# FOGLAMP_END
+
+import foglamp.core.core_daemon
```
Terris-Dianomic: I tried core_daemon being renamed to daemon. It hides the third-party daemon package.

src/python/foglamp/__main__.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+if __name__ == "__main__":
+    foglamp.core.core_daemon.main()
```
Terris-Dianomic: I am following this convention:

main() takes not arguments and processes command-line arguments
start() only works with input parameters

src/python/foglamp/admin_api/server.py
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env python3
+
+import asyncio
+from foglamp.admin_api import controller as admin_api_controller
+ 
+ 
+def start():
+    """Starts the service"""
+    admin_api_controller.start()
```
Terris-Dianomic: First comment: The Python guide is currently not specific about how to import. I'll add to it.

Second comment: Yes, we should have both so that you can python -m foglamp.device instead of foglamp.device.server. server.py is for tests and other modules to use, __main__ is convenience for users.

src/python/setup.py
```diff
@@ -11,8 +11,7 @@
     packages=['foglamp'],
     entry_points={
         'console_scripts': [
-            'foglamp = foglamp.server:run',
-            'foglampd = foglamp.server:main',
+            'foglamp = foglamp.core.core_daemon:main'
```
Terris-Dianomic: I wish that worked.

src/python/foglamp/device/__main__.py
```diff
@@ -0,0 +1,25 @@
...
+
+def start():
+    """
+    Starts the device server
+    """
+    foglamp.device.server.start()
+
+
+if __name__ == "__main__":
+    start()
```
Terris-Dianomic: Let's remove the if from __main.py__ and server.py

src/python/foglamp/core/api.py
Terris-Dianomic: We currently haven't decided on what the process name would be. I'm not sure the entry point stuff will be used. Depends on how the snap will work. The process that's running the core REST server is the code that has __start_time, so there is no need to look for a process. That would be needed only if core_daemon had a function to report the uptime which thankfully it doesn't.

src/python/foglamp/admin_api/server.py
Terris-Dianomic: "admin api" and "device" run in different processes now. "admin api" will be deleted -- but the UI has a dependency on it.

src/python/foglamp/device/server.py
Terris-Dianomic: This is a duplicate request

src/python/foglamp/env.py
```diff
@@ -1,63 +0,0 @@
-import os
```
Terris-Dianomic: Let's address it when needed

Terris-Dianomic:
I'm working on the __template__ sync with all files and fixing the imports.

src/python/foglamp/core/__main__.py
Terris-Dianomic: Thanks. Adding to guidelines.

Terris-Dianomic:
@vaibhav-ScaleDB  I can't reply to your comment. I have no idea why. 

build.sh actually takes multiple parameters

I think build.sh should be run like this:

./build.sh --install --start

If that's what you want.

./build.sh --start 

Should just start. It's faster and there's less crud on the screen.

src/python/foglamp/__main__.py
praveen-garg: ah! I missed that. Yes this will shadow the python-daemon's daemon module.

Yeah, here we either have to keep it as `core_daemon` itself or just another name may be`daemonizer`

Terris-Dianomic: I chose server_daemon to match server.py. Not ideal but better that core_daemon

src/python/foglamp/device/server.py
Terris-Dianomic: "Do we still need admin_api/wsgi.py ?" If we're not using nginx then no. admin_api will be deleted from 'develop' eventually and we will leave it in some fork.

Terris-Dianomic:
ready for review

PR #61, opened 2017-07-07
FogLAMP, Admin and Device daemons

amarendra-dianomic:
@MarkRiddoch @Terris-Dianomic, Please have a look into this PR regd FOGL-240 per your convenience. I have forked this from your FOGL-240 branch only to my individual repo. The main features of this PR are:
- Now it is possible to have three running daemons - one each for rest, admin and device
- At startup, all three daemons can be started from a single command "foglamp". This fires three daemons, namely foglampd, admind and deviced
- Once fired, all three daemons can be separately managed from command line using start|stop|restart|status command
- Now we have three entry points into foglamp - 1) startup foglamp command, 2) python -m <script>, and 3) server start command where server is one of foglampd|admind|devicd

Is this what we intend to do?


Terris-Dianomic:
The desired sequence is for the core to start the scheduler which will then start the device service and eventually the storage service.

PR #65, opened 2017-07-11
FOGL-259 Start coap server in core as a separate process

Terris-Dianomic:


src/python/foglamp/core/server.py
```diff
@@ -17,13 +20,26 @@
...
+    for task in asyncio.Task.all_tasks():
+        task.cancel()
+    loop.stop()
+
+
 def start():
     """starts the server"""
-    # https://aiohttp.readthedocs.io/en/stable/_modules/aiohttp/web.html#run_app
+    loop = asyncio.get_event_loop()
+    scheduler.start(loop)
```
praveen-garg: remove `loop` arg.

**note:** when process is running in background, `foglamp start`. `lsof -i:5683` lists 1 pid on UDP port 5683.
we can still run `python -m foglamp.device` and it will also start scheduler at port 5683 in foreground  with new pid (i.e. `lsof -i:5683` lists 2 pid)

> For UDP (Multicasts), multiple applications can subscribe to the same port.

src/python/foglamp/core/scheduler.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def shutdown():
+    """Stops the scheduler"""
+    for process in _processes:
+        process.terminate()
+
+
+async def _start_device():
+    process = await asyncio.create_subprocess_exec(
```
Terris-Dianomic: to an internal function?

src/python/foglamp/core/server.py
Terris-Dianomic: https://scaledb.atlassian.net/browse/FOGL-266

Terris-Dianomic:
Please re-review

src/python/foglamp/core/server_daemon.py
```diff
@@ -75,38 +83,54 @@ def start():
             _start_server()
 
 
-def stop():
-    """Stops the FogLAMP process if it is running"""
+class StopProcessFailed(Exception):
+    pass
```
praveen-garg: great addition 👍 , eventually we will need to have exceptions.py module for each package for our definitions

src/python/foglamp/core/server_daemon.py
```diff
@@ -188,6 +214,6 @@ def main():
             # If the daemon package has been invoked, the following 'write' will
             # do nothing
             sys.stderr.write(format(str(e)) + "\n");
```
praveen-garg: we have got this `;` in pull/58, let's get rid of this

src/python/foglamp/core/server_daemon.py
```diff
@@ -75,38 +83,54 @@ def start():
...
-        return 
+        return
 
-    # Kill the daemon process
-    # TODO This should time out and throw an exception
+    stopped = False
+    
     try:
-        while True:
+        for retry_index in range(_MAX_STOP_RETRY):
```
praveen-garg: we can also use `for _ in range(_MAX_STOP_RETRY):` as retry_index is not used.

src/python/foglamp/core/server_daemon.py
```diff
@@ -75,38 +83,54 @@ def start():
...
             os.kill(pid, signal.SIGTERM)
-            time.sleep(0.1)
+
+            for seconds_index in range(_WAIT_TERM_SECONDS):
```
praveen-garg: as above `for _`

src/python/foglamp/core/server_daemon.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+# !/usr/bin/env python3
```
Terris-Dianomic: Yes, there can be a space but I don't know where it came from

src/python/foglamp/core/server_daemon.py
Terris-Dianomic: I'll move it.

src/python/foglamp/core/server_daemon.py
Terris-Dianomic: Python community doesn't agree on using _ for iterating. https://stackoverflow.com/questions/3685974/python-iterate-a-certain-number-of-times-without-storing-the-iteration-number-a

src/python/foglamp/core/server_daemon.py
```diff
@@ -75,38 +83,54 @@ def start():
...
+        for retry_index in range(_MAX_STOP_RETRY):
             os.kill(pid, signal.SIGTERM)
-            time.sleep(0.1)
+
+            for seconds_index in range(_WAIT_TERM_SECONDS):
+                os.kill(pid, 0)
+                time.sleep(1)
     except OSError:
-        pass
+        stopped = True
```
Terris-Dianomic: Pid file existence check happens before this code.

src/python/foglamp/core/server_daemon.py
```diff
@@ -75,38 +83,54 @@ def start():
...
-    pid = get_pid()
 
-    if pid is None:
+def stop(pid = None):
+    """Stops FogLAMP if it is running
+
+    :param pid: Optional pid of the daemon process
+    """
+
+    if not pid:
```
Terris-Dianomic: There is a popular idea among the python community to threat none and false together and not specifically check for none unless there is a difference between none and 0. pid 0 is not valid.

src/python/foglamp/core/server_daemon.py
Terris-Dianomic: Done

Terris-Dianomic:
server_daemon.py was sent through pylint and I learned a lot. Suffice it to say that our code will never pass lint unless we triple the amount of coding time at hand.

Ready for re-review.

src/python/foglamp/core/server_daemon.py
Terris-Dianomic: OK, pylint doesn't complain when _ is used. This goes in the py guide.

src/python/foglamp/core/scheduler.py
```diff
@@ -0,0 +1,55 @@
...
+            process.terminate()
+        except ProcessLookupError:
+            # This occurs when the process has terminated already
+            pass
+
+
+async def _start_device_server():
+    """Starts the device server (foglamp.device) as a subprocess"""
+    process = await asyncio.create_subprocess_exec(
+        'python3', '-m', 'foglamp.device')
```
MarkRiddoch: We are presumably assuming we are running with a particular current directory here. I don't have an issue with this, but maybe it is something we should think about for the future. i.e. setting the working directory during startup.

Terris-Dianomic: -m means find the package in the installed packages. The snap would set this up properly. The wd is set in the daemon. It's currently hard coded. It was a parameter but that was removed by Amarenda because the python package to parse parameters was removed.

Terris-Dianomic: Also that hard coding will occur for he storage service only. The device service could be started by reading the database via the storage service (interval=0 no repeat).

PR #66, opened 2017-07-11
FOGL-155_vaibhav Generate HTML from docstrings for all code under src/python

vaibhav-ScaleDB:
**Changes:**

1. API docs are created on the fly now.
2. Added module files for current code.
3. Added support to run api docs from makefile. 
4. Added support to see changes in docstrings through livedoc (autodoc)


vaibhav-ScaleDB:
Running doc-test over this PR will fail, this is not a regression and not caused by this PR. 
There were always docstring issues in our code but since this PR now adds source code docstrings in .rst files and these docstrings are not as per PEP8 guidelines, doc tests fails. In order to pass them, we need to fix the docstrings in our source code, issue https://scaledb.atlassian.net/browse/FOGL-260

Terris-Dianomic:
This is awesome!!

Terris-Dianomic:
@praveen Is this ready to go?

docs/index.rst
```diff
@@ -10,6 +10,8 @@ Welcome to FogLAMP's documentation!
 
    build-tools
 
+   api/foglamp
```
praveen-garg: @Terris-Dianomic suggest the correct name `api/foglamp`

Terris-Dianomic:
It's too bad github won't let me reply above. "FogLAMP API" sounds good to me.. Will that work?

src/python/requirements-dev.txt
```diff
@@ -1,2 +1,3 @@
 -r requirements.txt
 -r requirements-test.txt
+-r requirements-doc.txt
```
vaibhav-ScaleDB: Actually Yes, for documenting the package, we need to parse all modules (.py files) and some modules are importing foglamp

docs/index.rst
Terris-Dianomic: FogLAMP API?

src/python/requirements-dev.txt
Terris-Dianomic: I mean requirements-dev.txt includes everything, even doc, which is fine

PR #70, opened 2017-07-12
Fogl-217 Configuration Manager

ashwinscale:
Please review and provide feedback.


src/python/foglamp/configuration_manager.py
```diff
@@ -0,0 +1,370 @@
...
+
+async def _read_category_value(category_key):
+    try:
+        async with aiopg.sa.create_engine('postgresql://foglamp:foglamp@localhost:5432/foglamp') as engine:
+            async with engine.acquire() as conn:
+                async for row in conn.execute(
+                        _configuration_tbl.select().where(_configuration_tbl.c.key == category_key)):
+                    return row.value
+    except Exception:
+        logging.getLogger('configuration-manager').exception('Unable to read all category names')
```
MarkRiddoch: Wrong message in exception

src/python/foglamp/configuration_manager.py
```diff
@@ -0,0 +1,370 @@
...
+            configuration.key='{}'
+    """
+    query_full = query_template.format(category_item, category_key)
+    try:
+        async with aiopg.sa.create_engine('postgresql://foglamp:foglamp@localhost:5432/foglamp') as engine:
+            async with engine.acquire() as conn:
+                async for row in conn.execute(text(query_full).columns(_configuration_tbl.c.value)):
+                    return row.value
+    except Exception:
+        logging.getLogger('configuration-manager').exception('Unable to read all category items')
```
MarkRiddoch: Again, message has been copied and needs to be changed

src/python/foglamp/configuration_manager.py
```diff
@@ -0,0 +1,370 @@
...
+    Unknown
+
+    Restrictions and Usage:
+    A FogLAMP component calls this method to obtain all category names in the FogLamp system
+    """
+    try:
+        return await _read_all_category_keys()
+    except:
+        logging.getLogger('configuration-manager').exception(
+            'Unable to read all category names')
```
MarkRiddoch: Logging already happens in _read_all_category_keys, we should not repeat logging at all levels.

ashwinscale:
> A minor point since it is temporary, but I would suggest make the connection string for the database a constant near the top of the file somewhere. If we were keeping the direct database code for longer I would suggest something different.

Done

> Likewise the 'configuration-manager' string, used in the logger, should be defined once at the top.

Done

> The 'print("error")' stuff should go before merging.

Removed prints altogether. Logging only.

> We log at the _xxx and xxx level for each entry point. This seems a little unnecessary.

We are now only logging at xxx level.

> We use await in the wrappers and in the underlying routines, this seems like duplication.
Await is necessary for calling all async functions. So in this call chain, it is expected that we call await at each level. Example:
await f1() -> await f2() -> await f3()

> create_new_category needs to merge any existing database content for the category. Remember every time we start the code we will create the category, if the category was already in existence we need to preserve the values that the user may have changed.
Understood. I'll make the change now and provide feedback when it's complete. 

> Now we have descriptions for the categories we should probably return them in the get_all_categories call. This is a change to the original requirement based on the schema change, but probably a good thing to do.
Understood. I'll make the change after merge functionality and provide feedback when it's complete. 

> We will need to add some verification at some point, e.g. checking types and that values that are set are valid for the types etc.
Understood. I'll do this last

src/python/foglamp/configuration_manager.py
```diff
@@ -0,0 +1,330 @@
...
+    sa.MetaData(),
+    sa.Column('key', sa.types.CHAR(10)),
+    sa.Column('description', sa.types.VARCHAR(255)),
+    sa.Column('value', JSONB),
+    sa.Column('ts', sa.types.TIMESTAMP)
+)
+"""Defines the table that data will be inserted into"""
+
+_connection_string = 'postgresql://foglamp:foglamp@localhost:5432/foglamp'
+_logger_name = 'configuration-manager'
```
praveen-garg: logger = logging.getLogger(__name__) and use `logger.exception(.....` as needed.

src/python/foglamp/configuration_manager.py
```diff
@@ -0,0 +1,330 @@
...
+"""FOGLAMP_PRELUDE_BEGIN
+{{FOGLAMP_LICENSE_DESCRIPTION}}
+
+See: http://foglamp.readthedocs.io/
+
+Copyright (c) 2017 OSIsoft, LLC
+License: Apache 2.0
+
+FOGLAMP_PRELUDE_END
+"""
```
praveen-garg:
```
# -*- coding: utf-8 -*-

# FOGLAMP_BEGIN
# See: http://foglamp.readthedocs.io/
# FOGLAMP_END
```


copyright license will go post import block, refer `__template__.py`

src/python/foglamp/configuration_manager.py
```diff
@@ -0,0 +1,330 @@
...
+
+import logging
+import psycopg2
+import aiopg.sa
+import sqlalchemy as sa
+import asyncio
```
praveen-garg: move to main itself, so that when plugged you just remove that block and you don't need to revisit this..

src/python/foglamp/configuration_manager.py
```diff
@@ -0,0 +1,330 @@
...
+See: http://foglamp.readthedocs.io/
+
+Copyright (c) 2017 OSIsoft, LLC
+License: Apache 2.0
+
+FOGLAMP_PRELUDE_END
+"""
+
+import logging
+import psycopg2
```
praveen-garg: unused?

src/python/foglamp/configuration_manager.py
```diff
@@ -0,0 +1,330 @@
...
+import asyncio
+from sqlalchemy.dialects.postgresql import JSONB
+from sqlalchemy.sql import text
+
+"""Configuration Manager
+"""
+
+__author__ = 'Ashwin Gopalakrishnan'
+__version__ = '${VERSION}'
+
```
praveen-garg: add copyright and license here; refer `__template__.py`

src/python/foglamp/configuration_manager.py
```diff
@@ -0,0 +1,330 @@
...
+_configuration_tbl = sa.Table(
+    'configuration',
+    sa.MetaData(),
+    sa.Column('key', sa.types.CHAR(10)),
+    sa.Column('description', sa.types.VARCHAR(255)),
+    sa.Column('value', JSONB),
+    sa.Column('ts', sa.types.TIMESTAMP)
+)
+"""Defines the table that data will be inserted into"""
```
praveen-garg: inserted only? per comment

src/python/foglamp/configuration_manager.py
```diff
@@ -0,0 +1,330 @@
...
+        """
+    query_full = query_template.format(category_item, item_value_replacement, category_key)
+    async with aiopg.sa.create_engine(_connection_string) as engine:
+        async with engine.acquire() as conn:
+            await conn.execute(query_full)
+
+async def get_all_category_names():
+    """Get all category names in the FogLAMP system
+
+    Keyword Arguments:
```
praveen-garg: per `__template__.py`

src/python/foglamp/configuration_manager.py
```diff
@@ -0,0 +1,330 @@
...
+    Unknown
+
+    Restrictions and Usage:
+    A FogLAMP component calls this method to obtain a category's configuration
+    """
+    try:
+        return await _read_category_value(category_name)
+    except:
+        logging.getLogger(_logger_name).exception(
+            'Unable to get all category names based on category_name {}'.format(category_name))
```
praveen-garg: let interpolation overhead be handled by logging code itself? 

`("Unable to get all category names based on category_name : %s", category_name)`

NOTE: this comment is for logger string format only (vs .format which is our best practice)

src/python/foglamp/configuration_manager.py
```diff
@@ -0,0 +1,330 @@
...
+#     json = await set_item_value('CATEG', "url", "blablabla")
+#
+#     print("test get_item_value")
+#     json = await get_item_value('CATEG', "url")
+#     print(json)
+#     print(type(json))
+#
+# if __name__ == '__main__':
+#     loop = asyncio.get_event_loop()
+#     loop.run_until_complete(main_test())
```
praveen-garg: 1 and only 1 new line at EOF

src/python/foglamp/configuration_manager.py
ashwinscale: done

src/python/foglamp/configuration_manager.py
ashwinscale: psycopg2 is unused. removed

src/python/foglamp/configuration_manager.py
ashwinscale: Need help with this... I copy and pasted the header.

src/python/foglamp/configuration_manager.py
ashwinscale: Done. May have made a mistake. please review.

src/python/foglamp/configuration_manager.py
ashwinscale: updated

src/python/foglamp/configuration_manager.py
ashwinscale: done

src/python/foglamp/configuration_manager.py
ashwinscale: done

src/python/foglamp/configuration_manager.py
ashwinscale: need more info

ashwinscale:
Note that I am not using a JSON schema package of any sort to do JSON validation...

src/python/foglamp/configuration_manager.py
```diff
@@ -32,33 +21,106 @@
...
+            type_name - string (fixed - 'type')
+                type_val - string (dynamic - ('boolean', 'integer', 'string', 'IPv4', 'IPv6', 'X509 certificate'))
+            default_name - string (fixed - 'default')
+                default_val - string (dynamic)
+            value_name - string (fixed - 'value')
+                value_val - string (dynamic)
+"""
 
-async def _create_new_category(category_name, category_description, category_json_schema):
+async def _merge_category_vals(category_val_new, category_val_storage):
```
MarkRiddoch: Does this need to be async? It's not doing any IO is it?

praveen-garg: Yes. `category_val_prepared = await _merge_category_vals...`  awaits in create_category def

src/python/foglamp/configuration_manager.py
```diff
@@ -0,0 +1,415 @@
...
+#     print(type(json))
+#
+#     print("test get_item")
+#     json = await get_item('CATEG', "url")
+#     print(json)
+#     print(type(json))
+#
+#     print("test get_value")
+#     json = await get_value('CATEG', "url")
+#     print(json)
```
praveen-garg: return is not json but just string. for all get_value

you may want to store in `result` as you are printing and doing type checking

> Understood that its test code only. you may want to add assert type(json) == dict or assert type(json) == str instead of print

src/python/foglamp/configuration_manager.py
```diff
@@ -0,0 +1,415 @@
...
+    'configuration',
+    sa.MetaData(),
+    sa.Column('key', sa.types.CHAR(10)),
+    sa.Column('description', sa.types.VARCHAR(255)),
+    sa.Column('value', JSONB),
+    sa.Column('ts', sa.types.TIMESTAMP)
+)
+"""Defines the table that data will be used for CRUD operations"""
+
+_valid_type_strings = ['boolean', 'integer', 'string', 'IPv4', 'IPv6', 'X509 certificate']
```
praveen-garg: password is a new addition in doc?

ashwinscale: password added.

src/python/foglamp/configuration_manager.py
ashwinscale: updated tests but not using assertions yet

PR #74, opened 2017-07-15
Fogl 202 scheduler

Terris-Dianomic:
- [x] daemon and core converted to classes because they have mutable globals
- [x] new scheduler.py supporting interval schedules
- [x] An interval schedule starts the device server at start-up time (no more hard-coded tasks). Start-up tasks do not repeat and they are not entered into the tasks table. The are start-up tasks by virtue of having type = 1.
- [x] Stopping the core daemon waits for tasks to stop

I am working on timed schedules. Methods will be docstring'ed and modules will be sent through lint. As you will see, there are many TODOs to handle errors (there are some JIRAs for this). This won't be ready for merge until Monday.


src/python/foglamp/core/scheduler.py
```diff
@@ -6,50 +6,349 @@
...
 __author__ = "Terris Linenbach"
 __copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
 __license__ = "Apache 2.0"
 __version__ = "${VERSION}"
 
-# For Process methods, see https://docs.python.org/3/library/asyncio-subprocess.html#asyncio.asyncio.subprocess.Process
-# or upgrade to a version of Python that uses type annotations
-_processes = []  # type: List[Process]
 
+class Scheduler(object):
```
praveen-garg:
``` 
  scheduler = Scheduler() 
   # call it TaskScheduler? Scheduler is fine too.. 
   # consider to have BaseScheduler and subclass it?

   # Move Db stuff to TaskSchedulerStore or task_scheduler_store.py ?

   scheduler.add_task(task0, _ScheduleType.<TIMED/ INTERVAL>, time=?, interval=?) 

   # scheduler.remove_task(task0)

   # I see that we are processing db objects array for it.. but read db and add to scheduler.. let's process in this way? Atleast this will be the way for REST API to tell db that I want this task to be get started/ stopped... individual task  and then this code will process those db objects 

   # task0 is instance/ def of Task
   # _ScheduleType.TIMED will need kwarg time else interval (with unit?), defaults to time?

    `# For "__main__"`
    scheduler.start()
    print('Press Ctrl+C to exit')

    try:
        # process tasks

    except (KeyboardInterrupt, SystemExit):
        # kill the instance




core server, cls.__scheduler.start()
it reads from storage, and do the stuff for tasks

but..

__scheduler.start only once and read storage once? or it should periodically read storage?
````

Comment (Terris-Dianomic) [inline src/python/foglamp/core/scheduler.py]: 1. "Control+C to quit" is already output by aiohttp
2. I think KeyboardInterrupt is already handled by asyncio (SIGINT handler) 
3. SystemExit, which I haven't heard about until now, needs more thought if it is different from SIGTERM or SIGINT. That is best handled via an asyncio exception handler. But I'm not sure if there is a good way to handle this case. Wait for tasks to exit? Proper shutdown requires writing end times to the database. There might not be time to do that. This might not be a case where a graceful shutdown is possible.
4. With SchedulerBase you might be thinking of the different types of schedules. That would be justified in the API layer which I haven't thought about yet. Internally I'll keep it a union-tuple matching the table.

**Comment by Terris-Dianomic** [comment]
I will create a jira to avoid the frequent rechecking of schedules. The scheduler will compute the next task start time and wait until then. Currently it evaluates the schedules every 20 seconds (configurable, but I can do better)

**Comment by Terris-Dianomic** [comment]
Sorry @izoratti-dianomic I made another change to the schema.

File: src/sql/foglamp_ddl.sql [hunk 2]
```diff
@@ -877,10 +878,16 @@ ALTER TABLE foglamp.schedules OWNER to foglamp;
 CREATE TABLE foglamp.tasks (
   id           uuid                        UNIQUE,                 -- Unique uuid, PK
   process_name character varying(20)       NOT NULL,               -- Name of the task
-  state        smallint                    NOT NULL,               -- State of the task: 1-Running, 2-Complete, 3-Cancelled
+  state        smallint                    NOT NULL,               -- State of the task: 1-Running, 2-Complete, 3-Cancelled Terris:
+                                                                   -- Terris: should there be a "didn't start" state
+                                                                   -- or is this only for tasks that started?
```
MarkRiddoch: By didn't start do you mean a task execution was attempted by it failed to run at all? This is a special case of failed, but I see no reason not to have it.

src/sql/foglamp_ddl.sql
```diff
@@ -877,10 +878,16 @@ ALTER TABLE foglamp.schedules OWNER to foglamp;
...
   start_time   timestamp(6) with time zone NOT NULL DEFAULT now(), -- The date and time the task started
   end_time     timestamp(6) with time zone,                        -- The date and time the task ended
   reason       character varying(20),                              -- The reason why the task ended
+                                                                   -- Terris: Please enumerate reason codes here
```
MarkRiddoch: I would say character 20 is too small. This was meant to be the place that a task could write a message that was task specific, not a code as such. Probably not well explained in the document.

src/python/foglamp/core/scheduler.py
```diff
@@ -4,52 +4,419 @@
...
+
+        return True
+
+    async def _start_startup_task(self, *args):
+        """Startup tasks are not tracked in the tasks table"""
+        task_id = str(uuid.uuid4())
+        logging.getLogger(__name__).info("Starting task %s %s", task_id, args)
+
+        # TODO: what if this fails?
+        process = await asyncio.create_subprocess_exec(*args)
```
MarkRiddoch: Log a fatal error in the short term. We should probably do better longer term. Looking at why it would not start, maybe retrying with a back-off etc. But to get the system running let's keep it simple.

src/python/foglamp/core/scheduler.py
```diff
@@ -4,52 +4,610 @@
...
+                sa.Column('script', pg_types.JSONB))
+        # Class variables (end)
+
+        # Instance variables (begin)
+        self._start_time = None
+        """When the scheduler started"""
+        self._paused = False
+        """When True, the scheduler will not start any new tasks"""
+        self._process_scripts = dict()
+        """Dictionary of scheduled_processes.id to script. Immutable."""
```
praveen-garg: what is `scheduled_processes.id`  ?

PR #76, opened 2017-07-17
DB Model Layer

amarendra-dianomic:


amarendra-dianomic:
@MarkRiddoch Please review this approach for Storage layer. If you consider it ok, then we can further develop upon this idea else it will be dropped.

amarendra-dianomic:
@MarkRiddoch , I agree to your views that this is an interim step towards designing a full fledged independent Storage layer. Please note that,
- As of now, Storage layer is embedded into different processes and is not a separate process like Scheduler or Device service. Every process will be talking to Storage via import rather than some API etc.
- My understanding per Architecture design is that when we will be running a Foglamp cluster, then for high availibility sake, DB will be replicated amongst all the cluster members. So it is not clear to me as to what is meant by running Storage layer as a detached process possibly on a different box.
- Currently, four components viz Scheduler, Configuration, Device and OMF translation, are directly accessing DB. If we move DB accesses from this components to Storage layer, then we will benefit greatly from this interim design in terms of less duplication and less chances of errors. 
- However, if we do not want to revisit the existing code, then we should at least do three tasks: 1) Remove Postgres connection info and table definitions from modules and use the same from tables.py as a lot of DB code e.g. readings table handling, is getting repeated in the modules. 2) Modify DB code in modules in order to add metrics records to statistics table, wherever applicable, and 3) Any further DB access code should be directly added to Storage layer in concerned Services class.

amarendra-dianomic:
@MarkRiddoch For now, we can use this PR for consolidating DB connection string definition and Schema definitions/changes at one place. We can start with this and may gradually move DB code from modules to this layer gradually, in small increments with proper unit and integration tests.

Terris-Dianomic:
If we're not going to use sqlalchemy in the storage layer, I don't think this PR should go into develop.

amarendra-dianomic:
In view of above evaluation, this PR is closed.

src/python/foglamp/storage/tables.py
```diff
@@ -0,0 +1,289 @@
...
+from sqlalchemy.dialects.postgresql.base import INET, UUID
+from sqlalchemy.dialects.postgresql import JSONB
+
+
+__author__ = "Amarendra Kumar Sinha"
+__copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
+__license__ = "Apache 2.0"
+__version__ = "${VERSION}"
+
+
```
Terris-Dianomic: I can understand the usefulness of this file for the storage engine which accesses all tables. If this was used by anything else, I would not wish to take the module import overhead of initializing all the variables when my code only uses one of them. Complex module initialization is problematic in general. For example, the 'metadata' variable below is really a temporary but it becomes a global because module initialization can only use globals. It's a minor thing in this particular case but in general becomes a snowball.

PR #77, opened 2017-07-18
FOGL-203 the simplest syslog logging system

praveen-garg:
**LEVELS:**  warn to exception

**address:**  

- /dev/log for Linux 

**In Ubuntu**, you can see syslogs using System Log app.

Log file: `/var/log/syslog`

`tail -f /var/log/syslog` to monitor logs

### DIFF
--- a/src/python/foglamp/configuration_manager.py
+++ b/src/python/foglamp/configuration_manager.py
@@ -6,11 +6,13 @@
 
 """ Configuration Manager """
 
-import logging
+# import logging
 import aiopg.sa
 import sqlalchemy as sa
 from sqlalchemy.dialects.postgresql import JSONB
 from sqlalchemy.sql import text
+
+from foglamp import logger
 
 __author__ = "Ashwin Gopalakrishnan"
 __copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
@@ -29,7 +31,8 @@ _configuration_tbl = sa.Table(
 
 _valid_type_strings = ['boolean', 'integer', 'string', 'IPv4', 'IPv6', 'X509 certificate', 'password']
 _connection_string = 'postgresql://foglamp:foglamp@localhost:5432/foglamp'
-_logger = logging.getLogger(__name__)
+# _logger = logging.getLogger(__name__)
+_logger = logger.setup(__name__)
 
 """
 General naming convention:
--- a/src/python/foglamp/logger.py
+++ b/src/python/foglamp/logger.py
@@ -1,0 +1,36 @@
+# -*- coding: utf-8 -*-
+
+# FOGLAMP_BEGIN
+# See: http://foglamp.readthedocs.io/
+# FOGLAMP_END
+
+""" Syslog logger """
+
+
+import logging
+from logging import handlers
+
+__author__ = "Praveen Garg"
+__copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
+__license__ = "Apache 2.0"
+__version__ = "${VERSION}"
+
+
+def setup(name):
+    """ Configures the log mechanism """
+
+    try:
+        handler = handlers.SysLogHandler(address='/dev/log')  # OSX: /var/run/syslog, Ubuntu: /dev/log
+        formatter = logging.Formatter(
+            fmt='[FOGLAMP] %(asctime)s - %(levelname)s :: %(module)s: %(message)s',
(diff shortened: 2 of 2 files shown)


src/python/foglamp/configuration_manager.py
```diff
@@ -29,7 +31,8 @@
 
 _valid_type_strings = ['boolean', 'integer', 'string', 'IPv4', 'IPv6', 'X509 certificate', 'password']
```
izoratti-dianomic: there should not be a password.

src/python/foglamp/logger.py
```diff
@@ -0,0 +1,36 @@
...
+__copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
+__license__ = "Apache 2.0"
+__version__ = "${VERSION}"
+
+
+def setup(name):
+    """ Configures the log mechanism """
+
+    try:
+        handler = handlers.SysLogHandler(address='/dev/log')  # OSX: /var/run/syslog, Ubuntu: /dev/log
```
izoratti-dianomic: Please remove the comment, it is misleading.

src/python/foglamp/configuration_manager.py
praveen-garg: This is what I requested to @ashwinscale after seeing in requirement doc for https://github.com/foglamp/FogLAMP/pull/70:

types
 please confirm if we should fix here (and update the doc)?

src/python/foglamp/logger.py
praveen-garg: you mean `# OSX: /var/run/syslog, Ubuntu: /dev/log` ?

izoratti-dianomic: Yes, I mean, let's not refer to OSX because FogLAMP does not meant to run on OSX. This only confuses a developer who may get the code. By reading the comment, the developer may think that FogLAMP works on OSX.

src/python/foglamp/configuration_manager.py
izoratti-dianomic: There is a misunderstanding: this is the list of possible configuration items, NOT the connection to the database. 
As described [here](https://docs.google.com/document/d/1QDkwb1l3XPZCTZfBVwmchgI9ay0vmfKNXUEwu1KcuOA), the database is accessible via socket with peer authentication. The connections string must be _"dbname='foglamp'"_

src/python/foglamp/logger.py
praveen-garg: Fixed. thanks

src/python/foglamp/configuration_manager.py
praveen-garg: so it is okay to keep `password` in `_valid_type_strings` or should be removed?

ashwinscale: password is fine to keep in _valid_type_strings. 
Ivan's concern was regarding this line:
_connection_string = 'postgresql://foglamp:foglamp@localhost:5432/foglamp'

We need to change this and all other connection strings to connect to the DB via socket connection. See:
https://docs.google.com/document/d/1QDkwb1l3XPZCTZfBVwmchgI9ay0vmfKNXUEwu1KcuOA/edit#heading=h.omckl7aauz2g

Terris-Dianomic: I tried "dbname='foglamp'"  in scheduler and device server .. it works

praveen-garg: Got it, I was tuned to _valid_type_strings password context only. @ashwinscale  could you please fix and test that `connection_string = "dbname='foglamp'"` in a new PR for `configuration_manager.py`? I will merge that and rebase this PR with develop.

> This is where  DB model  (interim solution for storage) layer https://github.com/foglamp/FogLAMP/pull/76/files shines, BTW. We are using conn_string in sensor_value, omf_translator and configuration_manager in develop branch; and it will be in scheduler, purge and other places as needed, so is for the same tables meta info e.g. readings. So having that at single place, will allow schema update (ddl match) etc. easy to maintain/ change. Just import that and later this will be easy to plug out this *interim* layer and plug in storage_engine methods which itself has an abstraction for other services for DB.

praveen-garg:
this syslog stuff will be to log only the errors and exception which may occur while connecting to DB/ storage engine, bootstrapping problems etc.; **AND/ OR until that db logger is in place.**

PR #79, opened 2017-07-19
FOGL-219 Configuration manager REST API

ashish-ScaleDB:
[FOGL-219](https://scaledb.atlassian.net/browse/FOGL-219) - Build the Configuration REST API

**How to run?**
1) install requirements.txt  or "source build.sh -i" or "source build.sh -a " and make install
2) start core service "python -m foglamp.core" or just foglamp start
3) Import collection and use configuration_manager environment  in postman and run the stuff with given or other test values.
4) for #3 you can use Curl too

-PFA
[collection.zip](https://github.com/foglamp/FogLAMP/files/1160328/collection.zip)

### DIFF
--- a/src/python/foglamp/core/api.py
+++ b/src/python/foglamp/core/api.py
@@ -13,16 +13,16 @@ __copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
 __license__ = "Apache 2.0"
 __version__ = "${VERSION}"
 
-START_TIME = time.time()
+__start_time = time.time()
 
-HELP = """
+_help = """
     -----------------------------------------------------------------------
     | GET             | /ping                                             |
 
     | GET             | /categories                                       |
     | GET             | /category/{category_name}                         |
-    | GET             | /category/{category_name}/{config_item}           |
-    | PUT DELETE      | /category/{category_name}/{config_item}/{value}   |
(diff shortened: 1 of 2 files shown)


src/python/foglamp/core/api.py
```diff
@@ -6,21 +6,117 @@
...
+    return web.json_response(result)
+
+
+async def create_category(request):
+    """
+
+    :param request:
+    :return:
+    """
+    # TODO: Missing in doc itself. Should we need to add create_category or
```
MarkRiddoch: No, there is no need for a create category REST endpoint, it was deliberately not included in the document

src/python/foglamp/core/api.py
```diff
@@ -6,21 +6,117 @@
...
+
+async def get_category(request):
+    """
+
+    :param request:  category_name is required
+    :return: the configuration items in the given category.
+    """
+    category_name = request.match_info.get('category_name', None)
+    category = await configuration_manager.get_category_all_items(category_name)
+    # TODO: If category is None from configuration manager. Should we send category
```
MarkRiddoch: Good point, error handling is not covered in the document. I should have defined the error mechanism. My thought for now would be to say we return None for the first pass and add better error handling.

PR #81, opened 2017-07-21
Fogl-291 core logging

Terris-Dianomic:
Log to syslog instead of console. Enhance foglamp.lgogger.


src/python/__template__.py
```diff
@@ -62,22 +62,23 @@
 """
 
 
-def function_with_types_in_docstring(param1, param2):
-    """Example function with types documented in the docstring.
+def function_with_types_in_docstring(param1: int, param2: str) -> bool:
```
praveen-garg: then write 2 def:
retain function_with_types_in_docstring
and add this to `function_with_types`

src/python/foglamp/core/scheduler.py
```diff
@@ -404,14 +409,14 @@ def _schedule_next_timed_task(self, schedule, schedule_execution, current_dt):
...
-        # Advance to the correct day if specified
+        """Advance to the correct day if specified"""
         if schedule.day:
             while dt.isoweekday() != schedule.day:
                 dt += self._ONE_DAY
 
         schedule_execution.next_start_time = time.mktime(dt.timetuple())
 
-    def _schedule_first_task(self, schedule, current_time):
+    def _schedule_first_task(self, schedule: _Schedule, current_time: int) -> None:
```
praveen-garg: want to use type here only? what about other defs?!  if this is the plan we need to prepare the team to use it to keep things consistent.

praveen-garg:
77 commits?!

Terris-Dianomic:
Not sure why there are 77 commits. Rebase issue? I will add to the python guide to recommend using types in method defs and inform the team. They are helpful if the type is static which is mostly the case.

docs/dev-tools.rst
```diff
@@ -0,0 +1,15 @@
+Developer Tools
+===============
+
+Python Template
+---------------
+See the file src/python/__template__.py
+
+include:: ../src/python/__template__.py
```
praveen-garg: include is for rst only; I think we need .. literalinclude::

docs/dev-tools.rst
```diff
@@ -0,0 +1,15 @@
...
+
+Bash Script Template
+--------------------
+See the file src/__template__.sh
+
+include:: ../src/__template__.sh
```
praveen-garg: include is for rst only; I think we need .. literalinclude::

docs/dev-tools.rst
praveen-garg: Fixed in https://github.com/foglamp/FogLAMP/pull/81/commits/0e632b748bcc4a9646cc29c8115f94e82bdffafd

docs/dev-tools.rst
praveen-garg: Fixed in https://github.com/foglamp/FogLAMP/pull/81/commits/0e632b748bcc4a9646cc29c8115f94e82bdffafd

PR #83, opened 2017-07-21
Fogl 261/262 - Rest api support for Schedules and Tasks - Read function only

amarendra-dianomic:
With reference to [Rest API Architecture Doc](https://docs.google.com/docuement/d/1JJDP7g25SWerNVCxgff02qp9msHbqA9nt3RAFx8-Qng/edit#heading=h.al9srgd46ohj), Rest api support for read functions for Schedules and Tasks provided. Create/Update/Delete api will be provided when required support becomes available from Scheduler module.

To test this PR, please proceed as follows:
- Create DB and populate the tables per SQL provided. Also, please do incorporate scheduler_demo_data.sql. ~~You may be required to do some schedule_time adjustments in order to run the Scheduler.~~
- If required, modify DB connection string, per your local setting, in scheduler.py and scheduler_db_services.py.
- Generate Tasks data by invoking the Scheduler.
- Now you can check GET urls from the above Architecture doc for Schedules and Tasks e.g.
http://localhost:8082/foglamp/tasks
http://localhost:8082/foglamp/tasks?name=xxxx&state=yyyy
http://localhost:8082/foglamp/tasks?name=xxxx
http://localhost:8082/foglamp/tasks?state=yyyy
http://localhost:8082/foglamp/tasks/latest
http://localhost:8082/foglamp/tasks/latest?name=xxxx
http://localhost:8082/foglamp/task/<task_id>
http://localhost:8082/foglamp/schedules
http://localhost:8082/foglamp/schedule/<schedule_id>


src/python/foglamp/core/routes.py
```diff
@@ -16,12 +16,30 @@
 def setup(app):
     # app.router.add_route('POST', '/foglamp/a-post-req', api.a_post_req, expect_handler = aiohttp.web.Request.json)
     app.router.add_route('GET', '/foglamp/ping', api.ping)
+
+    # Configuration
     app.router.add_route('GET', '/foglamp/categories', api.get_categories)
     app.router.add_route('GET', '/foglamp/category/{category_name}', api.get_category)
     app.router.add_route('GET', '/foglamp/category/{category_name}/{config_item}', api.get_category_item)
     app.router.add_route('PUT', '/foglamp/category/{category_name}/{config_item}/{value}', api.set_configuration_item)
```
MarkRiddoch: This is not as per the specification. In the specification the value is in the payload, not a parameter in the URL

src/python/foglamp/core/routes.py
```diff
@@ -16,12 +16,30 @@
...
+    app.router.add_route('GET', '/foglamp/schedule/{schedule_id}', api.get_schedule)
+    app.router.add_route('PUT', '/foglamp/schedule/{schedule_id}', api.update_schedule)
+    app.router.add_route('DELETE', '/foglamp/schedule/{schedule_id}', api.delete_schedule)
+
+    # Tasks - As per doc
+    app.router.add_route('GET', '/foglamp/tasks', api.get_tasks)
+    app.router.add_route('POST', '/foglamp/task', api.post_task)
+    app.router.add_route('GET', '/foglamp/tasks/latest', api.get_tasks_latest)
+    app.router.add_route('GET', '/foglamp/task/{task_id}', api.get_task)
+    # TODO: Find out why not DELETE/PUT a task? Cancel flag can be handled in database update.
```
MarkRiddoch: This can not be a database update, it must interact with the scheduler to make the scheduler cancel the task. The scheduler should not have to poll the database. As per PUT or DELETE, I am flexible on that, it's borderline that it should be delete as it does not remove the entity.

src/python/foglamp/core/scheduler_db_services.py
```diff
@@ -0,0 +1,285 @@
...
+"""
+
+import asyncpg
+
+__author__ = "Amarendra Kumar Sinha"
+__copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
+__license__ = "Apache 2.0"
+__version__ = "${VERSION}"
+
+__DSN = "postgresql://foglamp:foglamp@localhost:5432/foglamp"
```
MarkRiddoch: As per latest recommendation from Ivan we should not have this

src/python/foglamp/core/scheduler_db_services.py
```diff
@@ -0,0 +1,285 @@
...
+    # stmt = await conn.prepare(query)
+    # if schedule_id is None:
+    #     rows = await stmt.fetch()
+    # else:
+    #     rows = await stmt.fetch(schedule_id)
+
+    if not schedule_id:
+        _where_clause = ''
+    else:
+        _where_clause = ' WHERE id = \'{0}\''
```
MarkRiddoch: Putting the {0} here but the .format on the query seems a little risky to me. It is prone to error if the code is modified and new parameters introduced. I would prefer the format and placeholder to be together.

src/python/foglamp/core/scheduler_db_services.py
```diff
@@ -0,0 +1,285 @@
...
+
+    if not task_id:
+        _where_clause = ''
+    else:
+        _where_clause = ' WHERE id = \'{0}\''
+
+    query += _where_clause
+
+    if task_id is not None:
+        query = query.format(task_id)
```
MarkRiddoch: As above with the query formatting

src/python/foglamp/core/scheduler_db_services.py
```diff
@@ -0,0 +1,285 @@
...
+    else:
+        if state is None and name is None:
```
MarkRiddoch: This seems a little cumbersome, you could reduce the number of if clauses buy always have "WHERE 1 = 1", that way for each condition you want to add you simply say " AND <column> = <value)". Doing it this way as you add more filters the number of if .. elif clauses just explodes.

src/python/foglamp/core/scheduler_db_services.py
```diff
@@ -0,0 +1,285 @@
...
+            _where_clause = ''
+        elif state is None and name is not None:
+            _where_clause = ' WHERE process_name = \'{0}\''
+        elif state is not None and name is None:
+            _where_clause = ' WHERE state = \'{0}\''
+        else:
+            _where_clause = ' WHERE process_name = \'{0}\' and state = \'{1}\''
+        query += _where_clause
+
+        if state is None and name is None:
```
MarkRiddoch: Likewise by having the .format here you have to duplicate the if...elif chain. Better to keep them together.

src/python/foglamp/core/scheduler_db_services.py
```diff
@@ -0,0 +1,285 @@
...
+    # if state is None and name is None:
+    #     rows = await stmt.fetch()
+    # elif state is None and name is not None:
+    #     rows = await stmt.fetch(name)
+    # elif state is not None and name is None:
+    #     rows = await stmt.fetch(state)
+    # else:
+    #     rows = await stmt.fetch(state, name)
+
+    if state is None and name is None:
```
MarkRiddoch: As above with the if...elif and the duplication needed for the .format

src/python/foglamp/core/routes.py
amarendra-dianomic: @ashish-ScaleDB Please address this concern as it belongs to Configuration Manager Rest api.

amarendra-dianomic:
@MarkRiddoch please note that, 
- I have incorporated prepared statements in all DB queries, eliminating the need and risk of string format operators. 
- @ashish-ScaleDB will be taking care of Configuration Manager related observation above.
- As discussed earlier, all Scheduler related create/update/delete operations never be attempted directly. They will always be performed vide Schdeuler module provided methods. Once the support becomes available via Scheduler module, their implementation will be taken up vide a separate PR.

src/python/foglamp/core/routes.py
ashish-ScaleDB: Fixed in https://github.com/foglamp/FogLAMP/pull/86/commits/417ce2b118d06e1c17cc3ddd3333c68480e6a6a4 . Please rebase it.

praveen-garg:
1) **readonly** /foglamp/schedule/process should be /foglamp/scheduled_processes [IMP]
2) docstring can be better with :Example: `Curl -X {verb} -H {header if needed} -d {data if needed} url as in routes
3) avoid unnecessary new lines those just make file too lengthy to scroll and read e.g. `stmt= ..` newline `row = ...`
4) cancel Task todo clarity is already there

^^ so called trivial so merged; and also new work will be in with scheduler object caller so above can be taken care of in that PR or just a new refactoring  PR

PR #85, opened 2017-07-21
FOGL-192: Build Scheduled Metrics Collector - successfuly calculate and maintain deltas in statistics_history table

oshadmon:
The following code has an additional column (previous_value) in statistics table. In addition, it uses the information in statistics table to calculate delta of his row, and store that value in the statistics_history.


billehunt:
crap, I did not mean to approve this yet.

src/python/foglamp/update_statistics_history.py
```diff
@@ -0,0 +1,139 @@
+"""
```
Terris-Dianomic: The first line of a python file should not be the module docstring. Please see \_\_template\_\_.py https://foglamp.readthedocs.io/en/develop/dev-tools.html#python-template

It should start with: 

    # -*- coding: utf-8 -*-
    
    # FOGLAMP_BEGIN
    # See: http://foglamp.readthedocs.io/
    # FOGLAMP_END

src/python/foglamp/update_statistics_history.py
```diff
@@ -0,0 +1,139 @@
...
+__license__ = "Apache 2.0"
+__version__ = "${VERSION}"
+
+
+# Set variables for connecting to database
+_db_type = "postgres"
+_user = "foglamp"
+_db_user = "foglamp"
+_host = "192.168.0.182"
+_db = "foglamp"
```
Terris-Dianomic: All of the above variables should be deleted. The connection string is dbname='foglamp'

Please change purge.py too and include it in this branch.

src/python/foglamp/update_statistics_history.py
```diff
@@ -0,0 +1,139 @@
...
+
+# Create Connection
+_ENGINE = sqlalchemy.create_engine('%s://%s:%s@%s/%s' % (_db_type, _db_user, _user, _host, _db),  pool_size=20,
+                                   max_overflow=0)
+_CONN = _ENGINE.connect()
```
Terris-Dianomic: Globals like _CONN are initialized at module import time. This is not the time to do complex things like I/O because, if the guidelines are followed, there is no way to handle exceptions via import statements (it is possible to do this in Python but should be avoided for simplicity). Complex initialization should be performed in application logic. If you made this change and ran pylint against it, it would complain that you are using the 'global' keyword. Use a class attribute instead.

src/python/tests/test_metric_collection.py
```diff
@@ -0,0 +1,75 @@
...
+import sqlalchemy
+import time
+
+# Set variables for connecting to database
+_db_type = "postgres"
+_user = "foglamp"
+_db_user = "foglamp"
+_host = "127.0.0.1"
+_db = "foglamp"
+
```
Terris-Dianomic: Change the connection string

src/python/tests/update_statistics_table.py
```diff
@@ -0,0 +1,77 @@
...
+__version__ = "${VERSION}"
+
+
+# Set variables for connecting to database
+_db_type = "postgres"
+_user = "foglamp"
+_db_user = "foglamp"
+_host = "192.168.0.182"
+_db = "foglamp"
+
```
Terris-Dianomic: Change the connection string

src/python/foglamp/update_statistics_history.py
```diff
@@ -0,0 +1,139 @@
...
+    3. UPDATE the stats table with new values
+    Returns:
+
+    """
+    for key in _STATS_KEY_VALUE_LIST:
+        stats_select_result = __select_from_statistics(key=key)
+        value = int(stats_select_result[0][0])
+        previous_value = int(stats_select_result[0][1])
+        history_ts = time.mktime(stats_select_result[0][2].timetuple())
+        history_ts = time.strftime('%Y-%m-%d %H:%M:%S', time.localtime(history_ts))
```
ashwinscale: use now() or current_time() rather than the ts value from the statistics table

src/python/foglamp/update_statistics_history.py
```diff
@@ -0,0 +1,139 @@
...
+
+        __insert_into_stats_history(key=key, history_ts=history_ts, value=value-previous_value)
+        __update_previous_value(key=key)
```
ashwinscale: update_previous_value should have the value passed into it as an argument
__update_previous_value(key=key, value=value)

src/python/foglamp/update_statistics_history.py
```diff
@@ -0,0 +1,139 @@
...
+
+def __update_previous_value(key=''):
+    """Query: 
+    UPDATE 
+        statistics 
+    SET 
+        previous_value = (SELECT value FROM statistics WHERE key='READINGS')
+    WHERE 
+        value = (SELECT value FROM statistics WHERE key='READINGS'); 
+    """
```
ashwinscale: suppose key_arg and value_arg are parameters passed in to __update_previous_value, the logic should be:
UPDATE statistics SET previous_value = value_arg WHERE key = key_arg

src/python/foglamp/update_statistics_history.py
```diff
@@ -0,0 +1,144 @@
...
+    Execute query and return result
+    Args:
+        stmt: Query being executed
+
+    Returns:
+        Result of the query 
+    """
+    
+    engine = sqlalchemy.create_engine('postgres:///foglamp',  pool_size=20, max_overflow=0)
```
ashwinscale: Update _CONNECTION_STRING to 'postgres:///foglamp' and use _CONNECTION_STRING in call.

src/python/foglamp/update_statistics_history.py
```diff
@@ -0,0 +1,144 @@
...
+    Returns:
+        list of distinct keys
+    """
+    key_list = []
+    stmt = sqlalchemy.select([_STATS_TABLE.c.key.distinct()]).select_from(_STATS_TABLE)
+    result = __query_execution(stmt)
+
+    result = result.fetchall()
+    for i in range(len(result)):
+        key_list.append(result[i][0].replace(" ", ""))
```
ashwinscale: please trim whitespace using strip(),
https://stackoverflow.com/questions/1185524/how-to-trim-whitespace-including-tabs

src/python/foglamp/update_statistics_history.py
```diff
@@ -0,0 +1,144 @@
...
+    Returns:
+        list of distinct keys
+    """
+    key_list = []
+    stmt = sqlalchemy.select([_STATS_TABLE.c.key.distinct()]).select_from(_STATS_TABLE)
+    result = __query_execution(stmt)
+
+    result = result.fetchall()
+    for i in range(len(result)):
+        key_list.append(str(result[i][0]).rstrip().lstrip())
```
ashwinscale: you can just use strip(), it strips whitespace on both sides

src/python/tests/test_metric_collection.py
```diff
@@ -0,0 +1,75 @@
+"""
```
ashwinscale: remove these test files for now. possible route: copy the test code over to the main file and comment it out before check in. delete the test files. then push.

PR #91, opened 2017-07-26
Fogl 278 omf translator to update statistics

ashwinscale:


ashwinscale:
Please note that this is not to review the entire file, just the additions regarding statistics. Use of globals are not to be discussed here. The work required was to have omf translator using statistics. More modifications may need to be done by Stefano.

src/python/foglamp/translators/omf_translator.py
```diff
@@ -178,7 +182,7 @@ def initialize_plugin():
 
     try:
         # URL
-        _server_name = "WIN-4M7ODKB0RH2"
+        _server_name = "192.168.0.221"
```
vaibhav-ScaleDB: Any reason to change that? its just one hard-coded value that works for someone and not for other. Ultimately all these values will be read from a foglamp system config file or configuration manager

src/python/foglamp/translators/omf_translator.py
```diff
@@ -936,6 +941,13 @@ def omf_object_creation():
         _log.error(message)
         raise Exception(message)
 
+async def _update_statistics():
+    global _num_readings
+    global _num_discarded_readings
+    await statistics.update_statistics_value('SENT', _num_sent)
+    _num_readings = 0
+    await statistics.update_statistics_value('UNSENT', _num_unsent)
```
vaibhav-ScaleDB: Where are we computing _num_unsent value? Shouldn't it be inside the exception of the method send_omf_message_to_end_point?

src/python/foglamp/translators/omf_translator.py
```diff
@@ -936,6 +941,13 @@ def omf_object_creation():
         _log.error(message)
         raise Exception(message)
 
+async def _update_statistics():
+    global _num_readings
+    global _num_discarded_readings
```
vaibhav-ScaleDB: Where are the variables _num_readings and _num_discarded_readings used? In the next few lines we are using _num_sent and _num_unsent.

PR #92, opened 2017-07-27
FOGL-269 Scheduler bug fixes, refactoring, and API methods

Terris-Dianomic:
Please see the test for example usage. The API is documented in the scheduler architecture document.

### DIFF
--- a/src/python/foglamp/core/scheduler.py
+++ b/src/python/foglamp/core/scheduler.py
@@ -65,7 +65,7 @@ class ScheduleNotFoundError(ValueError):
 class Task(object):
     """A task represents an operating system process"""
 
-    class TaskState(IntEnum):
+    class State(IntEnum):
         """Enumeration for tasks.task_state"""
         RUNNING = 1
         COMPLETE = 2
@@ -73,17 +73,18 @@ class Task(object):
         INTERRUPTED = 4
 
     __slots__ = ['task_id', 'process_name', 'state', 'cancel_requested', 'start_time',
-                 'end_time', 'state']
+                 'end_time', 'state', 'exit_code', 'reason']
 
     def __init__(self):
         self.task_id = None  # type: uuid.UUID
         """Unique identifier"""
         self.process_name = None  # type: str
-        """From the scheduled_processes table"""
-        self.state = None  # type: Task.TaskState
+        self.reason = None  # type: str
+        self.state = None  # type: Task.State
         self.cancel_requested = None  # type: datetime.datetime
         self.start_time = None  # type: datetime.datetime
         self.end_time = None  # type: datetime.datetime
+        self.exit_code = None  # type: int
 
 
 class ScheduledProcess(object):
@@ -409,7 +410,7 @@ class Scheduler(object):
                         pid=(self._schedule_executions[schedule.id].
                              task_processes[task_id].process.pid),
                         process_name=schedule.process_name,
-                        state=int(Task.TaskState.RUNNING),
+                        state=int(Task.State.RUNNING),
                         start_time=datetime.datetime.now()))
 
             asyncio.ensure_future(self._wait_for_task_completion(task_process))
@@ -459,9 +460,9 @@ class Scheduler(object):
 
         if schedule.type != self._ScheduleType.STARTUP:
             if exit_code < 0 and task_process.cancel_requested:
-                state = Task.TaskState.CANCELED
+                state = Task.State.CANCELED
             else:
-                state = Task.TaskState.COMPLETE
+                state = Task.State.COMPLETE
 
             # Update the task's status
             # TODO if no row updated output a WARN row
@@ -470,9 +471,9 @@ class Scheduler(object):
                     result = await conn.execute(
                         self._tasks_tbl.update().where(
                             self._tasks_tbl.c.id == str(task_process.task_id)).values(
-                                    exit_code=exit_code,
-                                    state=int(state),
-                                    end_time=datetime.datetime.now()))
+                            exit_code=exit_code,
+                            state=int(state),
+                            end_time=datetime.datetime.now()))
 
                     if result.rowcount == 0:
                         self._logger.warning("Task %s not found. Unable to update its status",
@@ -740,7 +741,7 @@ class Scheduler(object):
                 await conn.execute(
                     self._tasks_tbl.update().where(
                         self._tasks_tbl.c.end_time is None).values(
-                            state=int(Task.TaskState.INTERRUPTED),
+                            state=int(Task.State.INTERRUPTED),
                             end_time=datetime.datetime.now()))
 
     async def _get_schedules(self):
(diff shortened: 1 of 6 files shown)


Terris-Dianomic:
There are bugs. Changing an existing schedule doesn't necessarily affect the next start time of one of its tasks. I'm working on it. This change should be sufficient for integration purposes however. Please merge it into your branch.

amarendra-dianomic:
@Terris-Dianomic my preliminary feedback is as under:
- I was able to run the test at command prompt only after I did **'pip install pytest_asyncio'**. So I think pytest_asyncio should be added to requirements.txt.
- After the test ended, I checked DB tables. I could find entries in scheduled_processes, schedules but no entry was found in tasks table, even for the task before modification, meaning that Scheduler did not execute the task, either before or after modifcation.
- AFAIK, this modification of the schedule should be carried out by another task itself. Meaning that calling scheduler.save_schedule() should be done via another task which should be scheduled to be executed  via the event loop.

Terris-Dianomic:
pytest_asyncio is in requirements_test.txt. You should be using a virtual environment (I thought you already were?).

You can prepare the virtual environment for development and testing via:

build.sh --install-dev-dep

Praveen can provide additional details.

(build.sh --install only installs requirements.txt)

The test creates a startup schedule. Startup tasks don't write to the tasks table. The next push will create other types of schedules.

save_schedule is an async function. You can put the coroutine on the event loop via ensure_future but I don't know why you would do that. You must "await" on the function to receive any exceptions, which is important for users.

amarendra-dianomic:
@Terris-Dianomic yes, you do not write startup tasks to tasks table. I actually assumed that a 'sleep'  could never be a startup task and could not notice the task type!

I did miss on dev environment thing. Thx for reminding about that.

Regarding the third point, please note below excerpts from Rest Api Architecture doc:

> **Task creation is done via the scheduler using the “schedule now” operation.**

> **Task cancellation is done via the scheduler using the entry point to the scheduler that cancels a task.**

Hence my understanding that of the need for carrying out create/update/delete tasks/schedules via another task. Please correct me if wrong.

Terris-Dianomic:
Creating and modifying a schedule via put() is not the same as starting a task. If a startup schedule is created, or if a manual schedule is started, a task will not start automatically and must be started manually. If an interval or a timed schedule is created, it will run according to the specs (e.g., run every 5 minutes). Starting and canceling tasks requires additional methods that I'm working on.

Terris-Dianomic:
Scheduler.queue_task is implemented with a bunch of bug fixes and refactoring. delete_schedule isn't ready yet.

Terris-Dianomic:
The get_task and get_all_tasks methods will be implemented in a different branch. This branch won't have additional work except for bug fixes

Terris-Dianomic:
Code complete

src/python/foglamp/core/scheduler.py
```diff
@@ -547,60 +791,431 @@ def _schedule_next_task(self, schedule):
...
+                self._main_sleep_task = asyncio.ensure_future(asyncio.sleep(sleep_seconds))
+
+                try:
+                    await self._main_sleep_task
+                    self._main_sleep_task = None
+                except asyncio.CancelledError:
+                    self._logger.debug("Main loop awakened")
+
+    @staticmethod
+    async def populate_test_data():
```
amarendra-dianomic: Can we move this code to test class setup_method?

src/python/foglamp/core/scheduler.py
```diff
@@ -547,60 +791,431 @@ def _schedule_next_task(self, schedule):
...
+
+        processes = []
+
+        for (name, script) in self._process_scripts.items():
+            process = ScheduledProcess()
+            process.name = name
+            process.script = script
+            processes.append(process)
+
+        return processes
```
amarendra-dianomic: Instead of list of objects, we need list of dictionary of attr:value pairs for each process. Otherwise, we will need to make this conversion at rest api end in order to send the json response.

amarendra-dianomic:
@Terris-Dianomic I have examined this PR purely in terms of Rest API requirements for which I checked out a separate branch from this branch. The observations need minor adjustments/modifications.

src/python/foglamp/core/scheduler.py
Terris-Dianomic: There are two reasons why not. First, the table metadata should not be duplicated. Secondly, database access will be moved to the storage layer.

src/python/foglamp/core/scheduler.py
Terris-Dianomic: There won't be a straight-forward conversion between the objects in the schedule module to JSON. Secondly, having an ordered list lets me write better tests. In fact because of this I found bugs that I wouldn't have found otherwise.

src/python/foglamp/core/scheduler.py
praveen-garg: let's add a TODO for its removal?  
second reason, compels me to think that we will be receiving things in data objects format (from storage engine) and we shall be using test doubles, to test scheduler at unit level.  Rest api interface test should take care of its integration tests.

src/python/tests/core/test_scheduler.py
```diff
@@ -0,0 +1,212 @@
...
+# See: http://foglamp.readthedocs.io/
+# FOGLAMP_END
+
+import asyncio
+import datetime
+
+import pytest
+import uuid
+
+from foglamp.core.scheduler import Scheduler, IntervalSchedule, ScheduleNotFoundError, Task
```
praveen-garg: we are testing IntervalSchedule only.

src/python/tests/core/test_scheduler.py
```diff
@@ -0,0 +1,212 @@
...
+        interval_schedule = IntervalSchedule()
+        interval_schedule.exclusive = False
+        interval_schedule.name = 'sleep1'
+        interval_schedule.process_name = "sleep1"
+        interval_schedule.repeat = datetime.timedelta(seconds=1)
+
+        await scheduler.save_schedule(interval_schedule)
+        await asyncio.sleep(10)
+
+        await self.stop_scheduler(scheduler)
```
praveen-garg: I think test has no assert, as we are just expecting that it can be considered as PASSED if NO ERROR occurs.

src/python/tests/core/test_scheduler.py
```diff
@@ -0,0 +1,212 @@
...
+        interval_schedule.process_name = "sleep1"
+        interval_schedule.repeat = datetime.timedelta(seconds=1)
+
+        await scheduler.save_schedule(interval_schedule)
+
+        # TODO: Re-read the task and check the type (need API support)
+
+        await asyncio.sleep(10)
+
+        # TODO: check for task created (need API support)
```
praveen-garg: now we have API support; still want to keep this as TODO?! for assert. Just an ask!

src/python/tests/core/test_scheduler.py
```diff
@@ -0,0 +1,212 @@
...
+        await scheduler.save_schedule(interval_schedule)
+        await asyncio.sleep(10)
+
+        await self.stop_scheduler(scheduler)
+
+    @pytest.mark.asyncio
+    async def test_save(self):
+        scheduler = Scheduler()
+
+        await scheduler.populate_test_data()
```
praveen-garg: populate test data in the same DB ? ultimately (asap) we should have a replica as foglamp_test [OR use storage engine test data objects, when storage engine will be available]

PR #94, opened 2017-07-27
categories with value

praveen-garg:
if ?values=1 then return vaule: {} too else work as earlier

Frontend, can avoid looping and mapping back to this collection for each cat/name request


ashwinscale:
This is fine with me if it is still needed.

praveen-garg:
it will be great to have it in. to avoid 3+ calls (assuming 3+ config categories) to server from UI **vs** small payload

Terris-Dianomic:
I thought one of the design goals was to reduce the number of REST calls.

billehunt:
Like I said on the call yesterday: I think we should do one call to get the
list of keys, then do one call for each of the keys. I don't understand why
that is more work than doing a single call.

The most important thing is that we get this built ASAP, and it sounds like
we've wasted 24 hours discussing this.

On Fri, Jul 28, 2017 at 10:06 AM, Terris Linenbach <notifications@github.com

praveen-garg:
On UI side, Configuration  Manager has already been done the way, the api code exists today in develop (without this change).. i.e. make call for each category (for its value) once categories (name + descriptions) are received. 

> why this PR?

The purpose of this PR was just to elaborate (if I was not clear) -- it will help to reduce http calls to server, hence db queries. And avoid little more work on ui side to loop through those calls and map received collection to original key object [chained calls]


> should we merge this PR or discard.

**Does not matter** 
original code will continue to work the old way, GET `?values=1` is just optional. it can be discarded as we are now doing chained calls on UI side (which we saw in demo today)

MarkRiddoch:
The original concept was that there would be a call per category rather than getting all the details of all the categories in a single call. The reason being that the assumption is there will be a lot of categories eventually and that this would be more than single screens worth of data, hence there being no requirement to get all the data in one go. 

I note now that chained calls are being employed. Does this mean this PR is no longer required? If so can it be deleted please.

PR #99, opened 2017-07-28
FOGL-321: Updated purge process to utilize streams.last_object as the last readings row id sent to Pi/OMF 

oshadmon:
@sstefano69 - Please take a look the get_last_id() method, and let me know if that's the proper way of retriving the last readings table row id sent Pi.


src/python/foglamp/data_purge/purge.py
```diff
@@ -107,57 +107,14 @@
...
-            time_in_sec = datetime.timedelta(seconds=time_dict[key])
-        elif 'min' in key:
-            time_in_min = datetime.timedelta(minutes=time_dict[key])
-        elif ('hr' in key) or ('hour' in key):
-            time_in_hr = datetime.timedelta(hours=time_dict[key])
-        elif ('day' in key) or ('dy' in key):
-            time_in_day = datetime.timedelta(days=time_dict[key])
-    return time_in_sec+time_in_min+time_in_hr+time_in_day
-
+   stmt = "SELECT MAX(last_object) FROM foglamp.streams;"
```
M1118: Using MAX(last_object) FROM fog lamp.steams will mean that we get the latest not sent to any channel. Whereas if we use MIN(last_object) then we get the latest not sent to all channels. There is a subtle but important difference.

However I do not understand why we need a Python function at all. Se can simply do it with the delete statement. If the flag to **not** delete unsent data is true then we add an extra where clause to the delete statement...

and id < (select MIN(last_object) FROM fog lamp.steams) 

So the statement is something like...

deleteunsent = ''
if ! deleteRowsNotSent: 
    deleteuenset = ' AND id < (SELECT MIN(last_object) FROM foglamp.steams)'
stmt = 'DELETE FROM foglamp.readings where DATE_PART(\'day\', now() - user_ts) * 24 + 
              DATE_PART(\'hour\',now() - user_ts) < {0} {1}'.format(age, deleteunsent)


The above Python may not be 100% correct, I'm making it up on the spur of the moment. However the idea is this should be really simple and just come down to creating a SQL statement that does all the work in the database. If Postgres had datediff it would be even easier.

sstefano69:
an explanation of the situation :

- the table foglamp.streams could contains multiple rows (one for each stream)
	SELECT * FROM foglamp.streams;
	
  so the proper one should be identified, 
  the OMF translator is currently using the id=1, so the query 
  	
  	SELECT last_object FROM foglamp.streams WHERE id=1;
  	
  will return the latest row of the readings table already sent (by the OMF translator).

- example :
	considering : 
		SELECT last_object FROM foglamp.streams WHERE id=1;
  
	will return 294, the query :
	
		SELECT *  FROM foglamp.readings WHERE id < 294;
	will identify rows already sent.
	
- following this, I think that :

	- considering the value returned by :
		SELECT last_object FROM foglamp.streams WHERE id=1;
	
	- and subtracting N (I'm doing it easily/quickly, maybe a more complex way should be used) it could be the X value usable for the purge, sample :
	
		DELETE FROM foglamp.readings WHERE id < X;

src/python/foglamp/data_purge/purge.py
oshadmon: @M1118 & @izoratti - I'm updating the code per this review. 
However, I'm not entirely sure why we are doing the `MIN(last_object)`. Given that `foglamp.streams` gets updated, why don't not do `SELECT last_object FROM `foglamp.streams WHERE description = 'OMF'`? Otherwise, if `foglamp.streams` has multiple rows regarding different historians, we could end up with a `MINI(last_object)` that was not sent to OMF.

oshadmon:
Code has been updated to have last_id as a sql query rather than value, and am waiting for review. 
For whatever, reason, age_timestamp didn't play along when trying to replace the resulting value with query instead.

src/python/foglamp/data_purge/purge.py
```diff
@@ -176,115 +128,74 @@ def execute_command(stmt: str):
...
-    if unsent_rows_removed < 0: 
-        unsent_rows_removed = 0
-    purge_set = {'start_time': start_time, 'end_time': end_time, 'rows_removed': total_rows_removed,
-                 'rows_remaining': total_count_after, 'unsent_rows_removed': unsent_rows_removed,
-                 'total_failed_to_remove': failed_removal_count}
+    start_time = time.strftime('%Y-%m-%d %H:%M:%S.%s', time.localtime(time.time()))
+
+    unsent_rows_removed = 0
+    total_rows_removed = 0
+    last_id = sqlalchemy.select([sqlalchemy.func.max(_STREAMS_TABLE.c.last_object)]).select_from(_STREAMS_TABLE)
```
MarkRiddoch: All the review comments say to use the MIN() of the objects ids in the streams table, so why is the sqlalchemy.func.max()?

oshadmon: fixed.

PR #104, opened 2017-08-02
psutil info added to ping api method

ashish-ScaleDB:


src/python/foglamp/core/api.py
```diff
@@ -38,20 +39,83 @@
...
-    :return: basic health information json payload
-    {'uptime': 32892} Time in seconds since FogLAMP started
+   Returns:
+        basic health information json payload
+
+   :Example:
+
+       For more detailed info use details query param
+
+       curl -X GET http://localhost:8082/foglamp/ping?details=1
```
praveen-garg: curl -X GET http://localhost:8082/foglamp/ping
curl -X GET http://localhost:8082/foglamp/ping?details=1

src/python/foglamp/core/api.py
```diff
@@ -38,20 +39,83 @@
...
+                }
+            }}
+
+    return result
+
+
 async def ping(request):
     """
+    Args:
+        request: details query param set to 1 for fetching more info
```
praveen-garg: request: accepts ?details=1 as query param to return process and system info

src/python/foglamp/core/api.py
```diff
@@ -38,20 +39,83 @@
...
+        "pid": process[0].pid,
+        "status": process[0].status(),
+        "start_time": process[0].create_time(),
+        "cpu_percent": process[0].cpu_percent(interval=1.0),
+        "memory_percent": process[0].memory_percent(),
+        "system_disk_usage": psutil.disk_usage('/')
+    })
+
+
+def _prepare_process_info(_now, since_started):
```
praveen-garg: should be just _system_usage_info

src/python/foglamp/core/api.py
```diff
@@ -38,20 +39,83 @@
...
+    if len(process) == 0:
+        return None
+
+    return dict({
+        "pid": process[0].pid,
+        "status": process[0].status(),
+        "start_time": process[0].create_time(),
+        "cpu_percent": process[0].cpu_percent(interval=1.0),
+        "memory_percent": process[0].memory_percent(),
+        "system_disk_usage": psutil.disk_usage('/')
```
praveen-garg: not here!

src/python/foglamp/core/api.py
```diff
@@ -38,20 +39,83 @@
...
+
+    process = [proc for proc in psutil.process_iter() if proc.name() == process_name]
+
+    if len(process) == 0:
+        return None
+
+    return dict({
+        "pid": process[0].pid,
+        "status": process[0].status(),
+        "start_time": process[0].create_time(),
```
praveen-garg: `"uptime": _now - process[0].create_time()`, even okay if `__start_time - process[0].create_time()`

src/python/foglamp/core/api.py
```diff
@@ -38,20 +39,83 @@
...
-    since_started = time.time() - __start_time
+    _now = time.time()
+    since_started = _now - __start_time
+    result = {'uptime': since_started}
 
-    return web.json_response({'uptime': since_started})
+    if 'details' in request.query:
+        detail_query_param = request.query['details']
+        if detail_query_param.isdigit() and detail_query_param == '1':
+            result = _prepare_process_info(_now, since_started)
```
praveen-garg: `result.update(_process_info(_now, "foglamp"))`
`result.update(_system_usage_info())` # should have 3 key-value pairs disk, memory and cpu under system_usage key

src/python/foglamp/core/api.py
```diff
@@ -9,6 +9,7 @@
 from aiohttp import web
 from foglamp import configuration_manager
 from foglamp.core import scheduler_db_services, statistics_db_services
+import psutil
```
izoratti: We must not add another dependency.

izoratti-dianomic:
I did not follow much these requirements, but it can't be done adding more dependencies. 
The merge cannot go through.

Also, I need to understand the meaning of the process info, since we have multiple processes, I do not understand what we are trying to achieve here.

src/python/foglamp/core/api.py
izoratti-dianomic: It should not be used.

src/python/foglamp/core/api.py
```diff
@@ -38,20 +39,83 @@
 
     | GET             | /foglamp/statistics                                       |
     | GET             | /foglamp/statistics/history                               |
-    ------------------------------------------------------------------------------
+    -------------------------------------------------------------------------------
 """
 
 
+def _find_process_info(process_name):
```
izoratti-dianomic: Which function/method does consume this?

src/python/foglamp/core/api.py
```diff
@@ -38,20 +39,83 @@
...
+        "cpu_percent": process[0].cpu_percent(interval=1.0),
+        "memory_percent": process[0].memory_percent(),
+        "system_disk_usage": psutil.disk_usage('/')
+    })
+
+
+def _prepare_process_info(_now, since_started):
+    """Prepare Json for detailed info"""
+
+    process_info = _find_process_info('foglamp')
```
izoratti-dianomic: Why do we retrieve info for this process name? We have multiple services, either we provide a full set of info, this one alone would be irrelevant.

src/python/requirements.txt
```diff
@@ -22,3 +22,5 @@ PyJWT==1.5.0
 # service
 python-daemon==2.1.2
 
+# process
+psutil==5.2.2
```
izoratti-dianomic: I cannot be added.

src/python/foglamp/core/api.py
praveen-garg: This is the parent process. And yes, good point. later or sooner, we should collect child processes (for each services) info too.

src/python/requirements.txt
praveen-garg: there are other way too, issue command, read files/ output and parse etc..., but that will take time.

src/python/foglamp/core/api.py
izoratti: Ok, so let's put this on hold, since it is not relevant for the demo.

izoratti-dianomic:
Not merged.

PR #108, opened 2017-08-07
Fogl:359 - source, severity, offset filter support added to audit log entries

ashish-ScaleDB:
[FOGL-359](https://scaledb.atlassian.net/browse/FOGL-359) - audit trail (log) rest api > GET endpoint

Fixed items:
- [x] source, severity, offset filter added to audit log
- [x] assets_readings_demo_data.sql file added
- [x] aiohttp_cors enabled


src/python/foglamp/core/audit_trail_db_services.py
```diff
@@ -24,27 +24,39 @@ class _Severity(IntEnum):
...
     """
     conn = await asyncpg.connect(database=__DB_NAME)
-    _limit_clause = " LIMIT $1" if limit else " "
+    
+    _limit_clause = " LIMIT {0}".format(limit) if limit else " "
+    _offset_clause = " "
+    if limit:
+        _offset_clause = " OFFSET {0}".format(offset) if offset else " "
+
+    _where_clause = " "
```
MarkRiddoch: I'd rather we built where clauses based on individual items so we don't have to enumerate all of the combinations. 

You can use
   _where_clause = " WHERE 1 = 1 "
  if source:
    _where_clause = _where_clause + "AND source = {0}'.format(source)
 if severity:
    _where_clause = _where_clause + "AND severity = {0}'.format(_Severity[severity].value)

This way when we can more in the future we do not get an exponential explosion of if statements

ashish-ScaleDB:
@MarkRiddoch changes done as you suggested. Thanks!

PR #109, opened 2017-08-07
Fogl 366 - 1 OMF message for the block size / other improvements

sstefano69:
- send_omf_message_to_end_point changed 
- performance trk as debug using - omf_translator_perf as

### DIFF
--- a/src/python/foglamp/translators/omf_translator.py
+++ b/src/python/foglamp/translators/omf_translator.py
@@ -9,36 +9,7 @@
 The information are sent in chunks,
 the table foglamp.streams and block_size are used for this handling
 
-Note :
-    - how to run :
-        - it could be executed as is without parameters
-
-    - consider the constant LOG_SCREEN : Enable/Disable screen messages
-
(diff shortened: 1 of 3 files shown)


src/python/foglamp/translators/omf_translator.py
```diff
@@ -114,26 +88,29 @@
...
 _producer_token = ""
+_channel_id = 1
+"""Channel Id for the OMF translator"""
+_omf_max_retry = 5
+_omf_retry_sleep_time = 1
+
 
 # The size of a block of readings to send in each transmission.
-_block_size = 50
+_block_size = 1000
```
MarkRiddoch: We need to be careful with the block size we choose. Too small and the performance will be poor, too large and we will build up a big in memory structure. I would suggest we make this a configuration options to allow for easy tuning in the short term. Longer term this may be variable on a per stream basis.

sstefano69: done

PR #110, opened 2017-08-07
Fogl 336 - add, update, delete schedules

amarendra-dianomic:


MarkRiddoch:
Personally I would rather see us split the API into a number of functional areas rather than put all the bits in a single api.py file. It's already several 100 lines of code, I would much rather see api_schedules.py, api_statistics.py etc.

src/python/foglamp/core/api_configuration.py
```diff
@@ -0,0 +1,107 @@
+# -*- coding: utf-8 -*-
+
+# FOGLAMP_BEGIN
+# See: http://foglamp.readthedocs.io/
+# FOGLAMP_END
+
+import asyncio
+import datetime
+import time
+import re
```
MarkRiddoch: Regular expressions don't appear top be used.

src/python/foglamp/core/api_configuration.py
```diff
@@ -0,0 +1,107 @@
...
+from aiohttp import web
+from foglamp import configuration_manager
+from foglamp.core import scheduler_db_services, statistics_db_services, audit_trail_db_services
```
MarkRiddoch: scheduler services are not used here

src/python/foglamp/core/api_scheduler.py
```diff
@@ -0,0 +1,529 @@
...
+# See: http://foglamp.readthedocs.io/
+# FOGLAMP_END
+
+import asyncio
+import datetime
+import time
+import re
+from aiohttp import web
+from foglamp import configuration_manager
+from foglamp.core import scheduler_db_services, statistics_db_services, audit_trail_db_services
```
MarkRiddoch: Unused imports

src/python/foglamp/core/api_statistics.py
```diff
@@ -0,0 +1,72 @@
+# -*- coding: utf-8 -*-
+
+# FOGLAMP_BEGIN
+# See: http://foglamp.readthedocs.io/
+# FOGLAMP_END
+
+import asyncio
+import datetime
+import time
+import re
```
MarkRiddoch: Unused imports

MarkRiddoch:
Let's do these fixes on the branch and merge again. The merge was done because it was holding up the creation of the demo branch. We can cherry pick these fixes onto the demo branch as soon as they are in develop.

PR #116, opened 2017-08-08
FOGL-269 Scheduler bug fixes for REST API

Terris-Dianomic:


amarendra-dianomic:
@Terris-Dianomic code LGTM! I have incorporated necessary modifications in api/scheduler.py and have checked CRUD methods for Schedulers and Read and Cancel methods for Tasks successfully (Cancel Task method added and tested in this PR). **All methods are working fine vis-a-vis core/scheduler.py.** Test tests/core/test_scheduler.py is also working fine.

However, core/scheduler.py -> get_tasks() still does not support **name** and **state** selection as mentioned in Rest Api doc. Also, no support exists for get_tasks_latest, also mentioned in Rest Api doc. In absence of this support, I am forced to use direct DB access to implement these methods. If you plan to provide this support in some future PR, then we are good to go for merging of this PR.

praveen-garg:
get_tasks_running() route binding was not removed. hence develop branch is broken.

PR #127, opened 2017-08-14
FOGL-196 Device ingest refactor

Terris-Dianomic:
Move code in foglamp.device.code to foglamp.device.ingest so it can be called by other input mechanisms such as mqtt.


Terris-Dianomic:
Stats writer now starts (bug fix) and stops (new feature) correctly

src/python/foglamp/device/ingest.py
```diff
@@ -0,0 +1,180 @@
...
+        # at https://docs.google.com/document/d/1rJXlOqCGomPKEKx2ReoofZTXQt9dtDiW_BHU7FYsj-k/edit#
+        # and will be moved to a .rst file
+
+        cls._num_readings += 1
+
+        # Required keys in the data
+        try:
+            asset = data['asset']
+            timestamp = data['timestamp']
+        except KeyError:
```
Terris-Dianomic: TypeError should definitely be thrown out of this method. That's the general way all Python code should handle use cases where bad data is passed in.

Terris-Dianomic:
Replaced by #128

PR #128, opened 2017-08-15
FOGL-389 Refactor device service

Terris-Dianomic:
This is a continuation of PR #127 with a more Pythonic interface for add_readings(). It also deprecates the 'sensor-values' key in the CoAP payload with the preferred key 'readings' to match the evolved nomenclature.


Terris-Dianomic:
PR #137 if approved will override this PR.

praveen-garg:
Please resolve the conflict with develop, I think its ready to be into develop. as Next PR are branched from this itself and it looks a great in.

Terris-Dianomic:
I'll deal with the conflict. Also, as in PR #137 "on conflict ignore" should be added to the insert statements. The try/catch block around executing the insert statement can be removed.

PR #131, opened 2017-08-17
Fogl 376 - omf_translator - performance improvements / error handling changed / _sensor_types removed 

sstefano69:


src/python/foglamp/translators/omf_translator.py
```diff
@@ -698,7 +608,7 @@ def send_in_memory_data_to_picromf(message_type, omf_data):
 
 
 def position_read():
-    """Retrieves the starting point for the send operation
+    """Retrieves the starting point for the sent operation
```
MarkRiddoch: This was right before you changed it :-)

src/python/foglamp/translators/omf_translator.py
```diff
@@ -45,52 +39,44 @@
 _message_list = {
 
     # Information messages
-    "i000001": "operation successfully completed",
-    "i000002": _module_name + " - Started",
+    "i000001": " ",
```
MarkRiddoch: Is this message no longer used?

sstefano69: yes, it is no more used, I avoided to pack/change all the used id for now (I will do)  but at least I have highlighted.

PR #137, opened 2017-08-19
FOGL-399 Improve device server stability and performance

Terris-Dianomic:
This is a continuation of PR #128

See also https://github.com/chrysn/aiocoap/issues/73

# Summary of Changes

1. Convert from aiopg to asyncpg

2. The device server responds to a CoAP request before a row is inserted into the readings table. Sensor readings are added to an in-memory queue and are inserted into Postgres in a delayed fashion. When the configurable maximum queue size is reached, the coap handler responds immediately with a "busy" message. This will kick in particularly when the database server is down. The maximum queue size isn't reached in the current code under maximum load.

3. Because Postgres previously validated the input data, validation logic had to be moved to Python. A new package was added to validate ISO8601 datetime strings as recommended by the Python community via stackoverflow. Please see requirements.txt.

4. A configurable number of async tasks send configurable batches of insert statements to Postgres. These tasks wait a configurable number of seconds to fill the batch with as many rows as possible. Batches are sent to Postgres via the "copy to" command which appears to be the fastest way to insert multiple rows into Postgres. However, due to the unique read_key column, rows must be copied into a temp table and then moved from the temp table into the readings table using "on conflict ignore."

5. Retry a configurable number of times when a database connection or insert fails

6. Clean up the coap handler

# Setup

coap_sender runs on the VirtualBox host. foglamp runs on Ubuntu Desktop in a VirtualBox VM with 4GB RAM.

The coap server is started via: python3 -m foglamp.device

coap_sender: delay = 10. Run for 10 minutes. Execute 7 concurrent processes.

ingest.py: 1 queue (1 db connections), insert batch size = 100, queue flush interval = 1 second, batch yield frequency = 0

# Results

Stats spreadsheet: https://docs.google.com/spreadsheets/d/1uR3lZtR-6XJAjHGbaXjAUT4VP3LJjiaPKO7F4uFTl_E/edit#gid=1237209252

The device server inserts ~600 rows per second. 

The device server maxes out at 90% CPU and 8.0% RAM (using top). Two postgres processes each max out at 1.7% CPU and 3.9% RAM.

# Error received and ignored in this codepath: [Errno 111] Connection refused

Messages are buffered by aiocoap and sent to handlers. The above message is written to the aiocoap server's console after a handler has finished and aiocoap attempts to send the response to a client that has since terminated.

# Discussion

We don't know how many concurrent aiocoap handlers can be active (either running or waiting for an asynchronous task) at any given instant. This is not the same as requests per second because an instant is sub-second. Based on experimentation the number appears to be 19. We don't know which specific factors affect this value.

In the "Changes" section above, item 2 is a strategy to prevent the device server from running out of memory or otherwise degrading due to swapping. However, it isn't foolproof because aiocoap queues requests on its own (it probably just adds tasks to the asyncio event loop). This queue is not configurable. If the device server doesn't keep up with the aiocoap queue, the device server process will run out of memory. The current maximum queue size is too low.

The strategy for keeping up with aiocoap's queue is to write the fastest possible aiocoap handler; hence the device server's own queue. However, that queue must also be processed as quickly as possible; hence, inserts are batched.

During testing, the coap handler never returned a busy message even under heavy load because a good percentage of incoming requests are not getting handled for unknown reasons.

Scaling CoAP probably requires a multi-process solution with traffic directed at different ports because CoAP is UDP and a single request could span multiple packets.

# To Do

[ ] Try uvloop - on hold until uvloop fixes https://github.com/MagicStack/uvloop/issues/98
[ ] Investigate why device server uses 3.4% of CPU and 7% of memory for a long time after 5 or more concurrent coap_sender processes finish (still an issue?)
[ ] Test performance when the statistics writer is disabled
[ ] Read configuration settings from the configuration manager
[ ] Experiment with ingest.py's configuration settings
[ ] Test with multiple machines running coap_sender instead of "single machine multiple processes"
[ ] Change timeout in coap_sender to 10 seconds so they can be detected when running coap_sender for less than 4 minutes (vaibhav)
[ ] coap_sender should wait for acks (up to 10 seconds) before terminating (vaibav)


src/python/foglamp/device/ingest.py
```diff
@@ -53,62 +55,185 @@ class Ingest(object):
...
+    """insert objects are added to this queue"""
+
+    _insert_tasks = None  # type: List[asyncio.Task]
+    """asyncio tasks for :meth:`_insert_readings`"""
+
+    _get_next_insert_tasks = None  # type: List[asyncio.Task]
+    """asyncio tasks for asyncio.Queue.get called by :meth:`_insert_readings`"""
+
+    # Configuration
+    _max_db_connections = 4
```
MarkRiddoch: Ideally we should make this configurable so that we can easily tune this when doing performance tests

src/python/foglamp/device/ingest.py
```diff
@@ -53,62 +55,185 @@ class Ingest(object):
...
+                try:
+                    insert = await waiter
+                except asyncio.CancelledError:
+                    break
+                finally:
+                    cls._get_next_insert_tasks[task_num] = None
+
+            try:
+                if cls._engine is None:
+                    cls._engine = await aiopg.sa.create_engine(_CONNECTION_STRING,
```
MarkRiddoch: We should keep statistics on how often we need to do this.

src/python/foglamp/device/ingest.py
```diff
@@ -53,62 +55,185 @@ class Ingest(object):
...
         cls._discarded_readings += 1
 
     @classmethod
-    async def _write_statistics_loop(cls):
+    async def _insert_readings(cls, task_num):
+        """Inserts rows into the readings table using _insert_queue"""
+        _LOGGER.info('Insert readings loop started')
+
+        while True:
+            insert = None
```
MarkRiddoch: Some statistics keeping here might help tune the number of db engines

src/python/foglamp/device/ingest.py
```diff
@@ -53,62 +55,185 @@ class Ingest(object):
...
+    _stop = False
     """Set to true when the server needs to stop"""
 
+    _started = False
+    """True when the server has been started"""
+
+    _engine = None  # type: aiopg.sa.Engine
+    """Database connection pool"""
+
+    _insert_queue = None  # type: asyncio.Queue
```
MarkRiddoch: Presumably we do not have to protect this queue as the gross level of locking that the Python GIL lock performs helps us here?

Terris-Dianomic: Protection isn't needed because this is a single-threaded app. That doesn't mean there can't be gnarly reentrancy though, but reentrancy is controllable since it can only happen if "await" is invoked on something that actually waits. https://stackoverflow.com/questions/45427152/does-await-guarantee-execution-order

Terris-Dianomic:
Converted to asyncpg. See PR description above.

src/python/foglamp/device/ingest.py
```diff
@@ -53,62 +50,246 @@ class Ingest(object):
...
+    """Number of seconds to wait for a queue to reach the maximum batch size"""
+
+    _batch_yield_num_items = 25
+    """While creating a batch, yield to other tasks after this taking this many
+    items from the queue"""
+
+    _max_queue_items = 3*_max_batch_items
+    """Maximum number of items in a queue"""
+
+    _max_insert_attempts = 60
```
praveen-garg: I think its too high!

src/python/foglamp/device/ingest.py
```diff
@@ -53,62 +50,246 @@ class Ingest(object):
...
             return
 
         cls._stop = True
 
-        if cls._sleep_task is not None:
-            cls._sleep_task.cancel()
-            cls._sleep_task = None
+        for _ in cls._insert_readings_wait_tasks:
+            if _ is not None:
+                _.cancel()
```
praveen-garg: fine! but usually we use _ when its of no use inside loop

src/python/foglamp/device/ingest.py
```diff
@@ -53,62 +50,246 @@ class Ingest(object):
...
+        cls._stop = False
 
     @classmethod
     def increment_discarded_readings(cls):
         """Increments the number of discarded sensor readings"""
         cls._discarded_readings += 1
 
     @classmethod
-    async def _write_statistics_loop(cls):
+    async def _close_connection(cls, connection: asyncpg.connection.Connection):
```
praveen-garg: does with context not work?

src/python/foglamp/device/ingest.py
```diff
@@ -53,62 +50,246 @@ class Ingest(object):
...
+                        await connection.execute('create temp table t_readings '
+                                                 'as select asset_code, user_ts, read_key, reading '
+                                                 'from foglamp.readings where 1=0')
+                    else:
+                        await connection.execute('truncate table t_readings')
+
+                    await connection.copy_records_to_table( table_name='t_readings',
+                                                           records=inserts)
+
+                    await connection.execute('insert into foglamp.readings '
```
praveen-garg: we should pick an item of queue and COPY it. I think this operation create temp table, insert readings, copy to main readings is too expensive.

src/python/foglamp/device/ingest.py
```diff
@@ -118,14 +299,40 @@ def increment_discarded_readings(cls):
                 cls._discarded_readings = 0
             # TODO catch real exception
             except Exception:
-                _LOGGER.exception("An error occurred while writing readings statistics")
+                _LOGGER.exception('An error occurred while writing readings statistics')
+
+        _LOGGER.info('Device statistics writer stopped')
+
+    @classmethod
+    def is_available(cls) -> bool:
```
praveen-garg: instead of doing this, we should keep a queue keep consumed if overflowing then log and halt ?!

src/python/foglamp/device/ingest.py
```diff
@@ -53,62 +50,246 @@ class Ingest(object):
...
+    """Which queue to insert into next"""
+
+    _insert_readings_tasks = None  # type: List[asyncio.Task]
+    """asyncio tasks for :meth:`_insert_readings`"""
+
+    _insert_readings_wait_tasks = None  # type: List[asyncio.Task]
+    """asyncio tasks for asyncio.Queue.get called by :meth:`_insert_readings`"""
+
+    # Configuration
+    _num_queues = 2
```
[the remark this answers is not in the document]
Terris-Dianomic: Yes, I benchmarked it against one, two, three, and four queues. Two queues are faster. While one queue is getting items the other is writing them to postgres.

src/python/foglamp/device/ingest.py
Terris-Dianomic: With context requires using a connection pool which is more overhead than is needed.

src/python/foglamp/device/ingest.py
Terris-Dianomic: Good point. I'll fix it.

src/python/foglamp/device/ingest.py
terrisgit: At some point you have to stop queueing.

This currently happens only if the database connection goes down. I can't make it happen by bombarding the server with requests - something is holding down the number of requests that are sent to the handler. There appears to be a queue (probably the event loop) of calls to the aiocoap handler. If the handler doesn't process each request in a consistent amount of time, this queue will grow without bound until the process is out of memory.

How to respond to database connection errors (or perhaps the database is out of disk space) requires more thought.

PR #134, opened 2017-08-17
FOGL-320 FOGL-279 Purge old rows in tasks table

Terris-Dianomic:
Also:
1) prepare for a merge conflict from PR #130
2) read config from the configuration manager (FOGL-279)
3) fix a bug in get_task
4) log database queries and database errors

PR #130, opened 2017-08-16
FOGL-372 Advanced task query

Terris-Dianomic:
Please see test for example usage

PR #138, opened 2017-08-20
PR's 130 and 134 combined for Scheduler

Terris-Dianomic:
This combines PR #130 and PR #134 because merging them is nontrivial. Please refer to those PR's for specific changes.


amarendra-dianomic:
@Terris-Dianomic code LGTM! Tests are passing. Please note that,
- I have modified rest api's core/api/scheduler.py -> get_tasks() get_tasks() provided. I have tested it via postman and everything seems to be working quite ok. Request you to please review core/api/scheduler.py -> get_tasks() and get_tasks_latest() and let me know if it is ok. Otherwise I will revert this commit.
- Secondly, it is the right time that we move all storage layer related code (lines 32-290, 415-445, 1231-1582)  to at least a separate module, say scheduler_storage_services.py, so that a) code becomes much cleaner and easier to understand, and b) transition to a separate storage layer, at a later date, will be smoother.
- Thirdly, we must move method populate_test_data() to tests/test_scheduler.py.

Terris-Dianomic:
scheduler.py will be refactored when all database code is moved to the storage server. How to deal with test data will need to be designed at that time. By design the API doesn't allow creation of scheduled_processes rows.

PR #139, opened 2017-08-23
FOGL-408: Implementation of unit-test for the purge process  

oshadmon:
The following is implementation of unit-test cases for the purge process. 
Until now, the only "testing" was inside the actual purge code, and was commented out.


oshadmon:
Pull request was done ahead of it's time. Will reopen when ready.

PR #36, opened 2017-06-20
FOGL_72 Encrypt JWT with RSA

amarendra-dianomic:

PR #140, opened 2017-08-24
cleanup 1.0 [and notes]

praveen-garg:
**TODO**
- [ ] FOGL-580: missing delete **TRUNCATE** from tables in foglamp_init_data 

--------------
**Other**
- [ ] what is the plan for admin_api ? src/python/foglamp/admin_api  HMAC or RSA for JWT, see https://github.com/foglamp/FogLAMP/pull/36

-  [ ] **what else**?

--------------
**After Storage layer integration**

- [ ] get rid of SqlAlchemy/ asyncpg / aiopg)

      -  hence delete from src/python/requirements.txt
       1. aiopg==0.13.0
       2. SQLAlchemy==1.1.10
       3. psycopg2==2.7.1

- [ ] remove CONN_STRING 

- [ ] **what else?** 

--------------
**Done**

- [x] remove rest api manual doc

- [x] remove unused healthcheck.py

- [x] remove httpie==0.9.9 package from requirements.txt

- [x] delete configuration  init data, it was malformed too


MarkRiddoch:
Removing SQLalchemy would be a good thing, but we really want the storage layer. The is now spec'd and we need to get a development estimate. I suspect it is not worth taking resources to remove SQLAlchemy when there are other things that can be done.

As for security, we need a story on what to use. However we are talking to OSIsoft and probably want to hold off on this until we have completed the discussions with OSIsoft.

praveen-garg:
This is LEAST and OBVIOUS cleanup for develop branch, please approve for merge!

izoratti-dianomic:
@praveen-garg - I am ok with the checklist, but I want to understand the " delete configuration init data, it was malformed too"
Can you be more specific?
1. What does "malformed" means in detail?
2. Is everything all automated now? Log Codes? Scheduled Processes? Statistics?

praveen-garg:
> What does "malformed" means in detail?

```
     VALUES ( 'STREAMING', 'Streaming', '{ "status" : "day", "window" : [ "always" ] }' );
```

**value** which will be registered by service/ plugin itself, must follow this name-value pair contract:  

```
"status": {
        "description": "streaming status",
        "type": "string",
        "default": "day"
    }
```
> Is everything all automated now? Log Codes? Scheduled Processes? Statistics?

No! 

**Log Codes? :** Audit log codes should be picked from log_codes, as of today only PURGE process is using the log code (though its not picking from log_codes but hardcoded value)

**Scheduled Processes?** instead of being registered automatically as a default schdule, we have them in init data; I believe this should be from UI and default schedule should come from plugin/ service itself (and should go away from init data)

**Statistics?**
Again keys are hard-coded, when stats are being written by each process.

Hence I did not touch these in this clean-up process; but these needs a design decision how each service will pick its log code key and stats key

PR #143, opened 2017-08-28
FOGL-417 fix to not change category_val json dictionary passed in cre…

ashwinscale:
fix in place


oshadmon:
Change seem to have fixed the issue that existed.

PR #147, opened 2017-08-30
FOGL-193 interest registration + callback + fix to only update when new value is different from old

ashwinscale:
Ready for merge pending review.


src/python/foglamp/configuration_manager.py
```diff
@@ -336,34 +365,47 @@
...
                 _logger.exception(
                     'category_value for category_name %s from storage is corrupted; using category_value without merge',
                     category_name)
+            # if validating category from storage succeeds, merge new and storage
             else:
                 category_val_prepared = await _merge_category_vals(category_val_prepared, category_val_storage, keep_original_items)
-            return await _update_category(category_name, category_val_prepared, category_description)
-
+                if json.dumps(category_val_prepared, sort_keys=True) == json.dumps(category_val_storage, sort_keys=True):
+                    # raise ValueError('category_value must be different than category_value in storage')
```
ashwinscale: When the value in storage is the same as the new value, I do not make an update to the DB. In these cases, should I raise an exception, or return with no message?

src/python/foglamp/configuration_manager.py
```diff
@@ -271,16 +283,31 @@
...
     Return Values:
     None
     """
     try:
-        return await _update_value_val(category_name, item_name, new_value_entry)
+        # get storage_value_entry and compare against new_value_value, update if different
+        storage_value_entry = await _read_value_val(category_name, item_name)
+        if storage_value_entry == new_value_entry:
+            # raise ValueError('new_value_entry must be different than value_entry in storage')
+            return
```
ashwinscale: When the value in storage is the same as the new value, I do not make an update to the DB. In these cases, should I raise an exception, or return with no message?

ashwinscale:
Currently, when new value is different from old, I return rather than throw an exception. Please advise on that. Everything else is done.

src/python/foglamp/configuration_manager.py
MarkRiddoch: I think silent return is fine

src/python/foglamp/configuration_manager.py
MarkRiddoch: If no change is made then I think it is reasonable to silently return

src/python/foglamp/configuration_manager.py
```diff
@@ -56,29 +60,37 @@
                 value_val - string (dynamic)
 """
 
+def _run_callbacks(category_name):
+    callbacks = _registered_interests.get(category_name)
+    if callbacks is not None:
+        for callback in callbacks:
+            cb = import_module(callback)
+            cb.run()
```
MarkRiddoch: It might be a good idea to pass the category in the callback, that way a single callback can be shared between multiple categories if a component had more than one category it was interested in.

src/python/foglamp/configuration_manager.py
```diff
@@ -56,29 +60,37 @@
                 value_val - string (dynamic)
 """
 
+def _run_callbacks(category_name):
+    callbacks = _registered_interests.get(category_name)
+    if callbacks is not None:
+        for callback in callbacks:
+            cb = import_module(callback)
+            cb.run(category_name)
```
praveen-garg: should we add:
```
except AttributeError:
              _logger.exception("{} must have implemented the run configuration-callback method".format(cb.__name__))
```

praveen-garg: what if this callback method wants to await another async method (may be read new configuration of the interested category-name from storage) . Basically what if we make `run` itself  async

src/python/foglamp/configuration_manager.py
```diff
@@ -336,35 +364,73 @@
...
+    except:
+        _logger.exception(
+            'Unable to run callbacks for category_name %s', category_name)
+        raise
     return None
 
 
-def register_category(category_name, callback):
-    pass
+def register_interest(category_name, callback):
```
praveen-garg: can a module register interest of multiple category names? 
`register_interest(*categories, callback_module)` and call `register_interest("CAT1", "CAT2", callback_module="foglamp.blah")`

src/python/foglamp/configuration_manager.py
```diff
@@ -35,6 +37,8 @@
 # _logger = logging.getLogger(__name__)
 _logger = logger.setup(__name__)
 
+_registered_interests = {}
```
praveen-garg: I believe, we shall need to make this `app level in memory storage`

src/python/foglamp/configuration_manager.py
ashwinscale: The callback wouldn't necessarily read back configuration. It would signal the interested component (which would read the configuration. However, signaling the interested component/microservice may be async. @MarkRiddoch what do you think?

src/python/foglamp/configuration_manager.py
ashwinscale: I could implement it this way.. but it wasn't specified in the design doc. @MarkRiddoch what do you think?

src/python/foglamp/configuration_manager.py
ashwinscale: Added exception handling, good call @praveen-garg

praveen-garg:
Approving this PR, as it works as per the specs. Closing #164 as well; Other chnages / enhancements can be taken care of, via another ticket.

PR #150, opened 2017-08-31
FOGL-427 FOGL-407 FOGL-401 FOGL-480 Various scheduler fixes

Terris-Dianomic:
Use a connection pool properly. Properly handle database errors. Shutdown-related bug fixes.


praveen-garg:
@aksinha-nerdapplabs  Please review

praveen-garg:
@Terris-Dianomic  please rebase with develop

amarendra-dianomic:
@Terris-Dianomic waiting for conflicts to be resolved on PR#150. However, I performed below steps for a manual schedule on my local on branch FOGL-429:
- I created a scheduled process testsleep2 with script ["sleep", "2"]. 
- I created a manual schedule via rest api throueh "curl -d '{"type": 4, "name": "sleep2test", "process_name": "testsleep2"}'  -X POST  http://localhost:8082/foglamp/schedule".
- I started this schedule manually via "curl -X POST  http://localhost:8082/foglamp/schedule/start/dc54c862-2624-4f7f-a38f-2fbd261ab5b1".
- This task was supposed to sleep for 2 secs and then exit. The expected status of the task was with state "COMPLETED" and with end-time suitably filled. But enquiring on the tasks via "curl -X GET  http://localhost:8082/foglamp/task" constantly returned state "RUNNING" and with end-time as None.

I will repeat the check after conclicts are resolved. Then I will proceed to check for soln for FOGL-401 and FOGL-427.

Terris-Dianomic:
@amarendra-dianomic I need to know how to start the rest service you are running against. Thanks.

amarendra-dianomic:
@Terris-Dianomic Please refer to [Jira Bug](https://scaledb.atlassian.net/browse/FOGL-480) for detailed description and steps.

amarendra-dianomic:
@Terris-Dianomic please note that,
- FOGL-427/480 are resolved in this PR. I have also tested it on local. 
- Code changes for FOGL-401 LGTM. Since tests are passing and manual schedule test is also working fine, it can be assumed that FOGL-401 also stands resolved.
- However, clarity is needed for checking FOGL-407. Please let me know as to how to reproduce the issue and how to check its solution. FYI, I tried below steps:
  - foglamp start
  - Create a schedule - curl -d '{"type": 3, "name": "sleep10", "process_name": "sleep10", "repeat": "1"}'  -X POST  http://localhost:8082/foglamp/schedule
  - Watch tasks - curl -X GET  http://localhost:8082/foglamp/task
  - delete tasks as specified in FOGL-407 in psql and watch tasks. Tasks count reduced initially but continue to grow.
  - delete scheduled_processes, as specified in FOGL-407, returned error as schedules table contained FK reference to scheduled_processes table.

So I am stuck at FOGL-407. Rest is fine.

Terris-Dianomic:
For 407, if you need to delete from tasks, schedules, and then scheduled_processes. The schedules are still in memory. Errors will go to syslog which is correct. The bug was that foglamp stop didn't work in this case. As you can imagine this is a phony use case but it was discovered because tests wipe out these tables and the scheduler was running when the unit test was started.

src/python/foglamp/core/scheduler.py
```diff
@@ -408,9 +408,9 @@ def __init__(self):
 
         # Initialize class attributes
         if not cls._logger:
+            # cls._logger = logger.setup(__name__)
             # cls._logger = logger.setup(__name__, destination=logger.CONSOLE, level=logging.DEBUG)
-            # cls._logger = logger.setup(__name__, level=logging.DEBUG)
-            cls._logger = logger.setup(__name__)
+            cls._logger = logger.setup(__name__, level=logging.DEBUG)
```
praveen-garg: we are not using foglamp logger ?

src/python/foglamp/core/scheduler.py
```diff
@@ -576,8 +587,21 @@ def max_running_tasks(self, value: int)->None:
 
         self._logger.info("Stopped")
 
+        if self._engine is not None:
+            try:
+                self._engine.close()
+            except Exception:
+                self._logger.exception('Unable to close the database connection pool')
+            self._engine = None
```
praveen-garg: should be in finally?

src/python/foglamp/core/scheduler.py
```diff
@@ -898,22 +918,20 @@ def _schedule_first_task(self, schedule, current_time):
         elif schedule.type == Schedule.Type.STARTUP:
             schedule_execution.next_start_time = current_time
 
-        self._logger.info(
-            "Scheduled task for schedule '%s' to start at %s", schedule.name,
-            datetime.datetime.fromtimestamp(schedule_execution.next_start_time))
+        if self._logger.isEnabledFor(logging.INFO):
+            self._logger.info(
```
praveen-garg: why this check is needed; it will only log info if its enable setup to that level

amarendra-dianomic:
@Terris-Dianomic w.r.t. FOGL-427, as advised, I carried out below steps in develop branch:
  1. truncate scheduled_processes, schedules, tasks tables
  2. in psql, insert into scheduled_processes (name, script) values ('sleep10', '["sleep", "10"]');
  3. foglamp start
  4. Create a schedule - curl -d '{"type": 3, "name": "sleep10", "process_name": "sleep10", "repeat": "1"}'  -X POST  http://localhost:8082/foglamp/schedule
  5. Watch schedules/tasks - curl -X GET  http://localhost:8082/foglamp/schedule and curl -X GET  http://localhost:8082/foglamp/task
  6. delete tasks, schedules and scheduled_processes and in that order. tasks and schedules deleted ok but received **ERROR:  update or delete on table "scheduled_processes" violates foreign key constraint "tasks_fk1" on table "tasks"
DETAIL:  Key (name)=(sleep10) is still referenced from table "tasks"**. Had to do a delete from tasks again to empty scheduled_processes.
  7. Again watch tasks via - curl -X GET  http://localhost:8082/foglamp/task. Empty set returned.
  8. foglamp stop did work ok.

Now when I repeated above steps in FOGL-427 branch, everything worked exactly same. So I am unable to find out what exactly went wrong and what was fixed. 

Further, **in both branches**, deleting schedules from schedules table, did not flush out the schedules from memory as curl -X GET  http://localhost:8082/foglamp/schedule continued to return the schedules even after deleting all records from schedules table.

src/python/foglamp/core/scheduler.py
Terris-Dianomic: It's to avoid an expensive call to datetime.datetime.fromtimestamp unnecessarily

src/python/foglamp/core/scheduler.py
Terris-Dianomic: I'll change it but we don't have a good handle of what should happen when non-Exceptions are raised which indicate very bad situations like out of memory.

Terris-Dianomic:
scheduler.py doesn't need to refresh its buffers from the database. Tables should not be modified outside the API.

PR #152, opened 2017-08-31
FOGL-389 FOGL-399 foglamp.device.ingest API and performance and stability enhancements

Terris-Dianomic:
Same changes as in PR #137 (which is a continuation of #128). PR 137 was accidentally merged into the FOGL-389 branch


src/python/foglamp/device/ingest.py
```diff
@@ -0,0 +1,497 @@
...
+            raise
+
+        # Comment out to test IntegrityError
+        # key = '123e4567-e89b-12d3-a456-426655440000'
+
+        cls.is_available()  # Locate a queue that isn't maxed out
+        queue_index = cls._current_readings_queue_index
+        queue = cls._readings_queues[queue_index]
+
+        if not cls._queue_readings_as_dict:
```
praveen-garg: if not ? and is this check actually required as we are checking and using json.dumps when fetching item for queue as an object to insert

src/python/foglamp/device/ingest.py
```diff
@@ -0,0 +1,497 @@
...
+    """Maximum number of items in a queue"""
+
+    _readings_batch_yield_items = 50
+    """While creating a batch, yield to other tasks after this taking this many
+    items from the queue"""
+
+    _readings_batch_wait_seconds = 1
+    """Number of seconds to wait for a queue to reach the minimum batch size"""
+
+    _max_insert_readings_batch_attempts = 60
```
praveen-garg: Though it shall be configurable; but I think its too high ... 60 attempts!

praveen-garg:
Need attention on 2 lines of code, other LGTM!

src/python/foglamp/device/ingest.py
Terris-Dianomic: This point of this variable is to control whether json.dumps is called in add_readings or in _insert_readings. Benchmarks show that it should be done in _insert_readings. add_readings should be as fast as possible.

src/python/foglamp/device/ingest.py
Terris-Dianomic: In the case of a connection failure, there is a 1 second sleep in between attempts. So 60 means to try to connect for 1 minute. I can change it to 30.

Terris-Dianomic:
Please review

PR #158, opened 2017-09-04
FOGL-399 foglamp.device.ingest.Ingest improvements

Terris-Dianomic:
Performance enhancements. Documentation and code fixes. Do not use asyncio.queue, enabling the readings buffer to be sent directly to postgres's 'copy' command without copying the buffer to another list. Use as many database connections as necessary up to a specified maximum per Praveen's suggestion.


praveen-garg:
@vaibhav-ScaleDB  could you please review and record the performance.

vaibhav-ScaleDB:
@praveen-garg , @Terris-Dianomic The performance results are available at https://docs.google.com/spreadsheets/d/1uR3lZtR-6XJAjHGbaXjAUT4VP3LJjiaPKO7F4uFTl_E/edit#gid=1550433153

Yes, there is improvement, went upto 7 senders. Looking at the trend, we can go upto 8 senders where we can expect around 900 m/sec operating at 100% CPU usage

Terris-Dianomic:
Thanks for all the time spent

PR #163, opened 2017-09-08
FOGL-461 - Fixed get_latest_tasks

amarendra-dianomic:


amarendra-dianomic:
@vaibhav-ScaleDB this failure will continue to occur intermittently as still we do not have any mechanism to provide a clean (i.e. empty task table) to each test. I have already proposed this, as a pure test requirement, vide [FOGL-410](https://scaledb.atlassian.net/browse/FOGL-410) but I am not sure if it would be provided or if at all it is provided, where exactly its location will be.

praveen-garg:
does not original test data (so called master test data for this module) help here?  if Yes, can we just use that itself?

amarendra-dianomic:
@praveen-garg init data fills scheduled_processes and schedules tables only and the schedules are a kind of long running processes, hence we need a separate test data. Further, tasks table gets filled by scheduler itself and we need a clear tasks table support from scheduler for each test to work reliably.

amarendra-dianomic:
@praveen-garg @vaibhav-ScaleDB please check.

vaibhav-ScaleDB:
Tested, LGTM!! Please resolve conflicts and then we are good to merge.

amarendra-dianomic:
@vaibhav-ScaleDB done. Pl check.

PR #164, opened 2017-09-08
decorator approach to allow async / sync registration; 

praveen-garg:
[JUST FOR DISCUSSION PURPOSE]
added some todo; examples where modules can create config and register interest too

### DIFF
--- a/src/python/foglamp/configuration_manager.py
+++ b/src/python/foglamp/configuration_manager.py
@@ -14,6 +14,7 @@ from sqlalchemy.sql import text
 from importlib import import_module
 import copy
 import json
+import functools
 
 from foglamp import logger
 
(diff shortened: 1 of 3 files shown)


src/python/foglamp/configuration_manager.py
```diff
@@ -60,12 +62,51 @@
...
-        for callback in callbacks:
-            cb = import_module(callback)
-            cb.run(category_name)
+    try:
+        if callbacks is not None:
+            for callback in callbacks:
+                cb = import_module(callback)
+                cb.run(category_name)
+    except AttributeError:
+        print("{} has not implemented the run method".format(cb.__name__))
```
praveen-garg: yes, add import error too! but let's have only 1 try block have 2 following except:

PR #171, opened 2017-09-13
FOGL-506 db connection user change, socket dir use from snap

ashwinscale:
explicitly state foglamp username on connection string, use snap's socket dir if available

everyone please review if your component is affected


ashwinscale:
I've asked everyone to review because I changed EVERY db connection

praveen-garg:
This is the PAIN, when you have to change a config thing at 10+ places.. I wish we could have a single file for such items or may be move to configuration-manager.py itself. 👎

PR #173, opened 2017-09-13
{poc and benchmarks tests only} added uvloop for performance comparision (CoAP ingress)

praveen-garg:


praveen-garg:
@MarkRiddoch There are some open issues for aiocoap +uvloop; I would suggest to keep this PR open as a pointer to evaluate! When all issues are resolved, we can rebase with develop and try again.

PR #176, opened 2017-09-13
FOGL-467 Device plugins

Terris-Dianomic:
The old device server can be started via "foglamp -m foglamp.legacy_device"

The new device server can be started via "foglamp -m foglamp.device --name=plugin_name"

For compatibility with today's processes so that, for example, "foglamp start" continues to work, --name= defaults to CoAP and there is a hack to load the module foglamp.device.coap_device since I don't know how the CoAP plugin configuration will be created (init_data.sql)?


Terris-Dianomic:
It's not quite working yet.

ingest.py
```diff
@@ -0,0 +1,368 @@
...
+
+        # Read config
+        """read config
+        _max_db_connections
+        _min_inserts_per_transaction
+        _max_inserts_per_transaction
+         *** validate > 0 ***
+        """
+
+        cls._max_insert_queue_size = 200
```
MarkRiddoch: I'd prefer t have this as a constant at the top somewhere, or even something we can override with configuration or in the constructor - but that is icing on the cake.

On moving further down I saw _readings_buffer_size, should this value be that?

PR #170, opened 2017-09-12
FOGL-484: Statistics Unit-Testing 

oshadmon:
The following is unit-testing for statistics.py and statistics_history.py

PR #185, opened 2017-09-19
FOGL-491: Unit-Test for configuration_manager (Priority 1)

oshadmon:
The following are unit-tests for the configuration_manager. I'd greatly appreciate it if whomever looks at the code, can thoroughly double check stuff related to register_interest since it was the one method I wasn't 100% sure on.

### DIFF
--- a/src/python/tests/test_configuration_manager.py
+++ b/src/python/tests/test_configuration_manager.py
@@ -1,0 +1,530 @@
+"""
+The following tests the configuration manager component For the most part,
+the code uses the boolean type for testing due to simplicity; but contains
+tests to verify which data_types are supported and which are not.
+"""
+
+import asyncio
+import pytest
+import sqlalchemy as sa
+import aiopg.sa
+from foglamp.configuration_manager import (create_category, set_category_item_value_entry,
+                                           register_interest, get_all_category_names,
+                                           get_category_all_items, get_category_item,
+                                           get_category_item_value_entry, _registered_interests,
+                                           _configuration_tbl)
+__author__ = "Ori Shadmon"
+__copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
+__license__ = "Apache 2.0"
+__version__ = "${VERSION}"
+
+pytestmark = pytest.mark.asyncio
+
+_CONNECTION_STRING = "dbname='foglamp'"
+_KEYS = ('boolean', 'integer', 'string', 'IPv4', 'IPv6', 'X509 cer', 'password', 'JSON')
+
+
+async def delete_from_configuration():
+    """ Remove initial data from configuration table """
+
+    sql = sa.text("DELETE FROM configuration WHERE key IN {}".format(_KEYS))
+    async with aiopg.sa.create_engine(_CONNECTION_STRING) as engine:
+        async with engine.acquire() as conn:
+            await conn.execute(sql)
+ 
+
+@pytest.allure.feature("unit")
+@pytest.allure.story("configuration manager")
+class TestConfigurationManager:
+    """ configuration_manager tests
+    
+    The following tests need to be fixed/implemented:
+
+        - Anything that is currently under @pytest.mark.xfail; currently doesn't work due to an expected behavior change
+        - FOGL-572: Verification of data type value in configuration manager (new test needs to be added)
+        - FOGL-585: Invalid Error when setting invalid category_name in create_category (new test needs to be added)
+        - FOGL-577
+        - 1 not yet implemented test
+    """
+
+    def setup_method(self):
+        """ reset configuration table data for specific category_name/s,
+        and clear data (if exists) in _registered_interests object"""
+
+        asyncio.get_event_loop().run_until_complete(delete_from_configuration())
+        _registered_interests.clear()
+
+    def teardown_method(self):
+        """reset foglamp data in database, and clear data (if exists)
+        in _registered_interests object"""
+        asyncio.get_event_loop().run_until_complete(delete_from_configuration())
+        _registered_interests.clear()
+
+    async def test_accepted_data_types(self):
+        """ Test that the accepted data types get inserted
+
+            - create_category
+            - get_all_category_names (category_name and category_description)
+            - get_category_all_items (category_value by category_name)
+
+        :assert:
+            1. Assert that the number of values returned by get_all_category_names
+                equals len(data)
+            2. category_description returned with get_all_category_names correlates to the
+                correct ke
+            3. get_category_all_items returns valid category_values for a given key
+        """
+
+        data = {
+            'boolean': {'category_description': 'boolean type',
+                        'category_value': {
+                            'info': {
+                                'description': 'boolean type with default False',
+                                'type': 'boolean',
+                                'default': 'False'}}},
+            'integer': {'category_description': 'integer type',
+                        'category_value': {
+                            'info': {
+                                'description': 'integer type with default 1',
+                                'type': 'integer',
+                                'default': '1'}}},
+            'string': {'category_description': 'string type',
+                       'category_value': {
+                           'info': {
+                               'description': "string type with default 'ABCabc'",
+                               'type': 'string',
+                               'default': 'ABCabc'}}},
+            'JSON': {'category_description': 'JSON type',
+                     'category_value': {
+                         'info': {
+                             'description': "JSON type with default {}",
+                             'type': 'JSON',
+                             'default': '{}'}}},
+            'IPv4': {'category_description': 'IPv4 type',
+                     'category_value': {
+                         'info': {
+                             'description': "IPv4 type with default '127.0.0.1'",
+                             'type': 'IPv4',
+                             'default': '127.0.0.1'}}},
+            'IPv6': {'category_description': 'IPv6 type',
+                     'category_value': {
+                         'info': {
+                             'description': "IPv6 type with default '2001:db8::'",
+                             'type': 'IPv6',
+                             'default': '2001:db8::'}}},
+            'X509 cer': {'category_description': 'X509 Certification',
+                     'category_value': {
+                         'info': {
+                             'description': "X509 Certification",
+                             'type': 'X509 certificate',
+                             'default': 'x509_certificate.cer'}}},
+            'password': {'category_description': 'Password Type',
+                         'category_value': {
+                             'info': {
+                                 'description': "Password Type with default ''",
+                                 'type': 'password',
+                                 'default': ''}}}
+        }
+
+        existing_records = 0
+
+        select_count_stmt = sa.select([sa.func.count()]).select_from(_configuration_tbl)
+        async with aiopg.sa.create_engine(_CONNECTION_STRING) as engine:
+            async with engine.acquire() as conn:
+                result = conn.execute(select_count_stmt)
+                async for r in result:
+                    existing_records = int(r[0])
+
+        for category_name in data:
+            await create_category(category_name=category_name,
+                                  category_description=data[category_name]['category_description'],
+                                  category_value=data[category_name]['category_value'],
+                                  keep_original_items=True)
+
+        sql = sa.text("SELECT * FROM configuration WHERE key IN {}".format(_KEYS))
+        async with aiopg.sa.create_engine(_CONNECTION_STRING) as engine:
+            async with engine.acquire() as conn:
+                result = await conn.execute(sql)
+                assert 8 == result.rowcount
+
+        categories = await get_all_category_names()
+        assert len(categories) == existing_records + 8
+
+        # only filter and test above 8 records
+
+        for key in data:
+            # print(key)
+            assert key in [cat[0].strip() for cat in categories]
+            assert data[key]['category_description'] in [cat[1] for cat in categories]
+
+            category_info = await get_category_all_items(category_name=key)
+            assert data[key]['category_value']['info']['description'] == (
+                category_info['info']['description'])
+            assert data[key]['category_value']['info']['type'] == (
+                category_info['info']['type'])
+            assert data[key]['category_value']['info']['default'] == (
+                category_info['info']['default'])
+
+    async def test_create_category_keep_original_items_true(self):
+        """ Test the behavior of create_category when keep_original_items == True
+
+        :assert:
+            1. `values` dictionary has both categories
+            2. values in 'data' category are as expected
+            3. values in 'info' category did not change
+        """
+        await create_category(category_name='boolean', category_description='boolean type',
+                              category_value={
+                                  'info': {
+                                      'description': 'boolean type with default False',
+                                      'type': 'boolean',
+                                      'default': 'False'}},
+                              keep_original_items=False)
+
+        await create_category(category_name='boolean',
+                              category_description='boolean type',
+                              category_value={'data': {
+                                  'description': 'int type with default 0',
+                                  'type': 'integer',
+                                  'default': '0'}},
+                              keep_original_items=True)
+
+        category_info = await get_category_all_items(category_name='boolean')
+        # Both category_values exist
+        assert sorted(list(category_info.keys())) == ['data', 'info']
+        # Verify 'info' category_value
+        assert category_info['info']['description'] == 'boolean type with default False'
+        assert category_info['info']['type'] == 'boolean'
+        assert category_info['info']['default'] == 'False'
+        # Verify 'data' category_value
+        assert category_info['data']['description'] == 'int type with default 0'
+        assert category_info['data']['type'] == 'integer'
+        assert category_info['data']['default'] == '0'
+
+    async def test_create_category_keep_original_items_false(self):
+        """ Test the behavior of create_category when keep_original_items == False
+
+        :assert:
+            1. initial `info` data has been added
+            2. `values` dictionary only has 'data' category
+            3. values in 'data' category are as expected
+        """
+        await create_category(category_name='boolean', category_description='boolean type',
+                              category_value={'info': {
+                                  'description': 'boolean type with default False',
+                                  'type': 'boolean',
+                                  'default': 'False'}})
+
+        await create_category(category_name='boolean',
+                              category_description='boolean type',
+                              category_value={'data': {
+                                  'description': 'int type with default 0',
+                                  'type': 'integer',
+                                  'default': '0'}},
+                              keep_original_items=False)
+
+        category_info = await get_category_all_items(category_name='boolean')
+        # only 'data category_values exist
+        assert sorted(list(category_info.keys())) == ['data']
+        # Verify 'data' category_value
+        assert category_info['data']['description'] == 'int type with default 0'
+        assert category_info['data']['type'] == 'integer'
+        assert category_info['data']['default'] == '0'
+
+    async def test_set_category_item_value_entry(self):
+        """ Test updating of configuration.value for a specific key using
+
+            - create_category to create the category
+            - get_category_item_value_entry to check category_value
+            - set_category_item_value_entry to update category_value
+
+        :assert:
+            1. `default` and `value` in configuration.value are the same
+            2. `value` in configuration.value gets updated, while `default` does not
+        """
+        select_value_stmt = sa.select([_configuration_tbl.c.value]).select_from(
+            _configuration_tbl).where(_configuration_tbl.c.key == 'boolean')
+        await create_category(category_name='boolean', category_description='boolean type',
+                              category_value={
+                                  'info': {
+                                      'description': 'boolean type with default False',
+                                      'type': 'boolean',
+                                      'default': 'False'}})
+        result = await get_category_item_value_entry(category_name='boolean', item_name='info')
+        assert result == 'False'
+
+        await set_category_item_value_entry(category_name='boolean',
+                                            item_name='info', new_value_entry='True')
+        result = await get_category_item_value_entry(category_name='boolean', item_name='info')
+        assert result == 'True'
+
+    async def test_get_category_item(self):
+        """ Test that get_category_item returns all the data in configuration.
+
+        :assert:
+            Information in configuration.value match the category_values declared
+
+        """
+        await create_category(category_name='boolean', category_description='boolean type',
+                              category_value={
+                                  'info': {
+                                      'description': 'boolean type with default False',
+                                      'type': 'boolean',
+                                      'default': 'False'}
+                              })
+        result = await get_category_item(category_name='boolean', item_name='info')
+        assert result['description'] == 'boolean type with default False'
+        assert result['type'] == 'boolean'
+        assert result['default'] == 'False'
+        assert result['value'] == 'False'
+
+    async def test_create_category_invalid_dict(self):
+        """ Test that create_category returns the expected error when category_value is a 'string' rather than a JSON
+
+        :assert:
+            Assert that TypeError gets returned when type is not dict
+        """
+        with pytest.raises(TypeError) as error_exec:
+            await create_category(category_name='integer', category_description='integer type',
+                                  category_value='1')
+        assert "TypeError: category_val must be a dictionary" in str(error_exec)
+
+    async def test_create_category_invalid_type(self):
+        """ Test that create_category returns the expected error when type is invalid
+
+        :assert:
+            Assert that TypeError gets returned when type is not allowed e.g. float
+        """
+        with pytest.raises(ValueError) as error_exec:
+            await create_category(category_name='float', category_description='float type',
+                                  category_value={
+                                      'info': {
+                                          'description': 'float type with default 1.1',
+                                          'type': 'float',
+                                          'default': '1.1'}})
+        assert ('ValueError: Invalid entry_val for entry_name "type" for item_name info. valid: ' +
+                "['boolean', 'integer', 'string', 'IPv4', " +
+                "'IPv6', 'X509 certificate', 'password', 'JSON']") in str(error_exec)
+
+    async def test_create_category_case_sensitive_type(self):
+        """ Test that create_category returns the expected error when type is upper case
+
+        :assert:
+            Assert that TypeError gets returned when type is uppercase e.g. INTEGER
+        """
+        # TODO: should be case insensitive? EVEN for this SCREAMING_SNAKE_CASE makes more sense!
+        # e.g. X509_CERTIFICATE, IPV4 etc.
+
+        with pytest.raises(ValueError) as error_exec:
+            await create_category(category_name='INTEGER', category_description='INTEGER type',
+                                  category_value={
+                                      'info': {
+                                          'description': 'INTEGER type with default 1',
+                                          'type': 'INTEGER',
+                                          'default': '1'}})
+        assert ('ValueError: Invalid entry_val for entry_name "type" for item_name info. valid: ' +
+                "['boolean', 'integer', 'string', 'IPv4', " +
+                "'IPv6', 'X509 certificate', 'password', 'JSON']") in str(error_exec)
+
+    async def test_create_category_invalid_entry_value_for_type(self):
+        """ Test the case where value is set to the actual "value" rather than the string of the value
+
+        :assert:
+            Assert TypeError when type is set to bool rather than 'boolean'
+        """
+        with pytest.raises(TypeError) as error_exec:
+            await create_category(category_name='boolean', category_description='boolean type',
+                                  category_value={'info': {
+                                      'description': 'boolean type with default False',
+                                      'type': bool,
+                                      'default': 'False'
+                                  }})
+        assert ("TypeError: entry_val must be a string for item_name " +
+                "info and entry_name type") in str(error_exec)
+
+    async def test_create_category_invalid_entry_value_for_default(self):
+        """ Test the case where value is set to the actual value as per type instead of string of the value
+
+        :assert:
+            Assert TypeError when default is set to False rather than 'False'
+        """
+        with pytest.raises(TypeError) as error_exec:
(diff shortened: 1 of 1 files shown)


praveen-garg:
@oshadmon  Please use class TestConfigurationManager and add allure annotation there only, use deleteConfiguration as setupMethod/ tearDownMethod. Can use the same for `_registered_interest.clear() `
better to go  with `from foglamp.configuration_manager import * ` as we really need almost everything  instead of `from foglamp.configuration_manager import a, ..z`

remove print statements

praveen-garg:
also you can put `pytestmark = pytest.mark.asyncio` after `_CONNECTION_STRING = "dbname='foglamp'"` and then you shall not need `@pytest.mark.asyncio`

praveen-garg:
same applies to https://github.com/foglamp/FogLAMP/pull/170

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,710 @@
...
+        try:
+            async with aiopg.sa.create_engine(_CONNECTION_STRING) as engine:
+                async with engine.acquire() as conn:
+                    async for result in conn.execute(stmt):
+                        assert sorted(list(result[0].keys())) == ['data', 'info']
+                        assert result[0]['data']['description'] == (
+                            'boolean type with default True')
+                        assert result[0]['data']['type'] == 'boolean'
+                        assert result[0]['data']['default'] == 'True'
+                        assert result[0]['info']['description'] == (
```
praveen-garg: above to this line, add a line break  or add comment `# verify info item`

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,710 @@
...
+                                  'description': 'boolean type with default True',
+                                  'type': 'boolean',
+                                  'default': 'True'}},
+                              keep_original_items=True)
+        try:
+            async with aiopg.sa.create_engine(_CONNECTION_STRING) as engine:
+                async with engine.acquire() as conn:
+                    async for result in conn.execute(stmt):
+                        assert sorted(list(result[0].keys())) == ['data', 'info']
+                        assert result[0]['data']['description'] == (
```
praveen-garg: above to this line, add a line break  or add comment `# verify data item`

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,710 @@
...
+        """
+        Test the behavior of create_category when keep_original_items == False
+        :assert:
+            1. initial `info` data has been added
+            2. `values` dictionary only has 'data' category
+            3. values in 'data' category are as expected
+        """
+        stmt = sa.select([_configuration_tbl.c.value]).select_from(_configuration_tbl).where(
+            _configuration_tbl.c.key == 'boolean')
+        await self.setup_method()
```
praveen-garg: as above!

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,710 @@
...
+                    async for result in conn.execute(stmt):
+                        assert sorted(list(result[0].keys())) == ['data']
+                        assert result[0]['data']['description'] == (
+                            'boolean type with default True')
+                        assert result[0]['data']['type'] == 'boolean'
+                        assert result[0]['data']['default'] == 'True'
+        except Exception:
+            raise
+
+        await self.teardown_method()
```
praveen-garg: as above!

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,710 @@
...
+                              category_description='boolean type',
+                              category_value={'data': {
+                                  'description': 'boolean type with default True',
+                                  'type': 'boolean',
+                                  'default': 'True'}},
+                              keep_original_items=False)
+        try:
+            async with aiopg.sa.create_engine(_CONNECTION_STRING) as engine:
+                async with engine.acquire() as conn:
+                    async for result in conn.execute(stmt):
```
praveen-garg: also assert nothing like 'info' exists in result in anyway

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,710 @@
...
+                                          'type': 'float',
+                                          'default': '1.1'}})
+        assert ('ValueError: Invalid entry_val for entry_name "type" for item_name info. valid: ' +
+                "['boolean', 'integer', 'string', 'IPv4', " +
+                "'IPv6', 'X509 certificate', 'password', 'JSON']") in str(error_exec)
+        await self.teardown_method()
+
+        await self.setup_method()
+        with pytest.raises(ValueError) as error_exec:
+            await create_category(category_name='INTEGER', category_description='INTEGER type',
```
praveen-garg: testing case sensitivity of type? we may want to support this

praveen-garg: move to another test;

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,710 @@
...
+            await create_category(category_name='boolean',
+                                  category_description='boolean type',
+                                  category_value={'info': {
+                                      'description': 'boolean type with default False',
+                                      'type': 'boolean',
+                                      'default': False
+                                  }})
+        assert ("TypeError: entry_val must be a string for item_name " +
+                "info and entry_name default") in str(error_exec)
+        await self.teardown_method()
```
praveen-garg: break into 2 tests `test_create_category_invalid_entry_value_type` and `test_create_category_invalid_entry_value`

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,710 @@
...
+
+        await self.setup_method()
+        with pytest.raises(ValueError) as error_exec:
+            await create_category(category_name='boolean', category_description='boolean type',
+                                  category_value={
+                                      'info': {
+                                          'type': 'integer',
+                                          'description': 'integer type with value False'}})
+        assert "ValueError: Missing entry_name default for item_name info" in str(error_exec)
+        await self.teardown_method()
```
praveen-garg: breakdown into 3? and 4th can be empty `'info': {}`?

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,710 @@
...
+
+    async def test_create_category_invalid_entry_none(self):
```
praveen-garg: what if empty string?

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,710 @@
...
+            raise
+
+        await set_category_item_value_entry(category_name='boolean',
+                                            item_name='info', new_value_entry='True')
+        try:
+            async with aiopg.sa.create_engine(_CONNECTION_STRING) as engine:
+                async with engine.acquire() as conn:
+                    async for result in conn.execute(stmt):
+                        assert result[0]['info']['value'] == 'True'
+                        assert result[0]['info']['default'] == 'False'
```
praveen-garg: nice test!

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,710 @@
...
+        except Exception:
+            raise
+        await self.teardown_method()
+
+    async def test_set_category_item_value_error(self):
+        """
+        Test updating of configuration.value when configuration is not set
+        :assert:
+            As of now the expected behavior is nothing gets returned; however,
+            there is an open issue expecting an error be returned when an update
```
praveen-garg: open issue? better to skip or xFail (expected failure) this test with JIRA id.
`@pytest.mark.skip(reason="FOGL-xxx")
` async def test_set_category_item_value_error(self):):`

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,710 @@
...
+                                  'type': 'boolean',
+                                  'default': 'False'}})
+        result = await get_category_item_value_entry(category_name='boolean', item_name='info')
+        assert result == 'False'
+
+        await set_category_item_value_entry(category_name='boolean',
+                                            item_name='info', new_value_entry='True')
+        result = await get_category_item_value_entry(category_name='boolean', item_name='info')
+        assert result == 'True'
+        await self.teardown_method()
```
praveen-garg: this has already been tested? or you want to `test_set_get_category_item_value_entry`

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,710 @@
...
+
+    async def test_get_category_item_value_entry_empty(self):
```
praveen-garg: I really doubt this should be here. we should not check non-existent things there can be 100x many things for each particle!

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,710 @@
...
+                                      'default': 'False'}
+                              })
+        result = await get_category_item_value_entry(category_name='boolean', item_name='data')
+        assert result is None
+
+        result = await get_category_item_value_entry(category_name='integer', item_name='info')
+        assert result is None
+        await self.teardown_method()
+
+    async def test_get_category_item(self):
```
praveen-garg: seems its already tested; we can not keep mixing scenarios

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,710 @@
...
+
+        await set_category_item_value_entry(category_name='boolean',
+                                            item_name='info', new_value_entry='True')
+        result = await get_category_item(category_name='boolean', item_name='info')
+        assert result['default'] == 'False'
+        assert result['value'] == 'True'
+
+        await self.teardown_method()
+
+    async def test_get_category_item_empty(self):
```
praveen-garg: no! checking which never tried or existed?!

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,710 @@
...
+                                      'default': 'False'}
+                              })
+        result = await get_category_item(category_name='integer', item_name='info')
+        assert result is None
+
+        result = await get_category_item(category_name='boolean', item_name='data')
+        assert result is None
+        await self.teardown_method()
+
+    async def test_get_category_all_items(self):
```
praveen-garg: this should be very first test and its already done!

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,710 @@
...
+
+        await  set_category_item_value_entry(category_name='boolean', item_name='info',
+                                             new_value_entry='True')
+        result = await get_category_all_items(category_name='boolean')
+        assert result['info']['default'] == 'False'
+        assert result['info']['value'] == 'True'
+
+        await self.teardown_method()
+
+    async def test_get_category_all_items_empty(self):
```
praveen-garg: how many times? and don't try non-existent things!

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,710 @@
...
+                                      'description': 'boolean type with default False',
+                                      'type': 'boolean',
+                                      'default': 'False'}
+                              })
+
+        result = await get_category_all_items(category_name='integer')
+        assert result is None
+        await self.teardown_method()
+
+    async def test_get_all_category_names(self):
```
praveen-garg: duplicate test? sounds like its a test for " " should be trimmed and empty string should not be allowed (a bug?)? though `  `  spaces are valid password but only for password?

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,710 @@
...
+            Assert error messages when both are None
+        """
+        await self.setup_method()
+        with pytest.raises(ValueError) as error_exec:
+            register_interest(category_name=None, callback='foglamp.callback')
+        assert "ValueError: Failed to register interest. category_name cannot be None" in (
+            str(error_exec))
+
+        with pytest.raises(ValueError) as error_exec:
+            register_interest(category_name='integer', callback=None)
```
praveen-garg: where is third test?

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,710 @@
+"""
+The following tests the configuration manager component For the most part,
+the code uses the boolean type for testing due to simplicity; but contains
```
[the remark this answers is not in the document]
oshadmon: @praveen-garg - what do you mean by verification of type?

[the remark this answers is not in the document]
praveen-garg: if type is JSON, make sure non son value is not allowed; and same for all allowed types. **test coverage**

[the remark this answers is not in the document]
oshadmon: @praveen-garg - According to @ashwinscale, the verification of the default value is done on the component, rather than when creating or updating a category. 

Example: For `_DEFAULT_PURGE_CONFIG`  it is up to the PURGE script to verify that the default/value given are a string of type int, and not something else. 
`_DEFAULT_PURGE_CONFIG = { 'iage': { 'description'='age of data', 'type'='integer', 'default'='72'}}` 


As of now, I'm not aware of process that actually check's its own configuration, in which case either everyone has to implement it, or there is (another) bug in the configuration_manager.

src/python/tests/test_configuration_manager.py
oshadmon: this tests `get_category_item_value_entry` method. Notice that while some methods repeat, such as  `create_category`, and  `set_category_item_value_entry`, the test name corresponds to a method call that has not been done before. In this case it is the `get_category_item_value_entry`.

src/python/tests/test_configuration_manager.py
oshadmon: you are absolutely right; however, we should have some level of negative testing. As such, these can either be as part of the existing class (what we currently have), or as an independent class.

src/python/tests/test_configuration_manager.py
oshadmon: meant to test get_all_category_names. this has been fixed, and simplified to only test a few categories.

src/python/tests/test_configuration_manager.py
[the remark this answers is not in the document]
praveen-garg: I believe this must be responsibility of conf manager to make sure, data value is as per allowed types specs!
Open a TICKET please, and let's discuss there as an improvement to implement or won't fix

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,653 @@
...
+                                 'default': ''}}}
+        }
+        for category_name in data:
+            await create_category(category_name=category_name,
+                                  category_description=data[category_name]['category_description'],
+                                  category_value=data[category_name]['category_value'])
+        try:
+            async with aiopg.sa.create_engine(_CONNECTION_STRING) as engine:
+                async with engine.acquire() as conn:
+                    async for count in conn.execute(select_count_stmt):
```
praveen-garg: should not we execute only ONCE? fetch info and assert

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,653 @@
...
+        all_category_name = await get_all_category_names()
+
+        for key in data:
+            select_all_stmt = sa.select([_configuration_tbl.c.key, _configuration_tbl.c.description,
+                                         _configuration_tbl.c.value]).select_from(
+                                         _configuration_tbl).where(_configuration_tbl.c.key == key)
+            try:
+                async with aiopg.sa.create_engine(_CONNECTION_STRING) as engine:
+                    async with engine.acquire() as conn:
+                        async for result in conn.execute(select_all_stmt):
```
praveen-garg: again fetchRows ? and loop through that. execute statement once

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,653 @@
...
+                            assert result[0].replace(" ", "") == key
```
praveen-garg: just fyi, there is `.strip()` method too for TRIM

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,653 @@
...
+                              category_description='boolean type',
+                              category_value={'data': {
+                                  'description': 'boolean type with default True',
+                                  'type': 'boolean',
+                                  'default': 'True'}},
+                              keep_original_items=True)
+        try:
+            async with aiopg.sa.create_engine(_CONNECTION_STRING) as engine:
+                async with engine.acquire() as conn:
+                    async for result in conn.execute(select_value_stmt):
```
praveen-garg: execute once?

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,653 @@
...
+                        assert sorted(list(result[0].keys())) == ['info']
+                        assert result[0]['info']['description'] == (
+                            'boolean type with default False')
+                        assert result[0]['info']['type'] == 'boolean'
+                        assert result[0]['info']['default'] == 'False'
+        except Exception:
+            raise
```
praveen-garg: this try block looks redundant check to me, 

`assert sorted(list(result[0].keys())) == ['data']` and following will ensure it has been replaced!

praveen-garg: No harm though!

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,653 @@
...
+        """
+        Test case where value is set to the actual "value" rather than the string of the value
+        :Assert:
+            Assert TypeError when description is set to None rather than ''
+        """
+        with pytest.raises(TypeError) as error_exec:
+            await create_category(category_name='boolean',
+                                  category_description='boolean type',
+                                  category_value={'info': {
+                                      'description': None,
```
praveen-garg: is empty string "" allowed?

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,653 @@
...
+        except Exception:
+            raise
+
+    async def test_set_category_item_value_error(self):
+        """
+        Test updating of configuration.value when category_name does not exist
+        :assert:
+            Nothing happens / returned
+        :expect:
+            FOGL-522: When updating category item value, when the category_name
```
praveen-garg: add xfail reason=FOGL-522 ?

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,653 @@
...
+                                            item_name='info', new_value_entry='True')
+        try:
+            async with aiopg.sa.create_engine(_CONNECTION_STRING) as engine:
+                async with engine.acquire() as conn:
+                    async for result in conn.execute(select_value_stmt):
+                        assert result is None
+        except Exception:
+            raise
+
+    async def test_get_category_item_value_entry(self):
```
praveen-garg: name should be `test_get_set_category_item_value_entry` as it verifies set too ( update )

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,653 @@
...
+        assert result['description'] == 'boolean type with default False'
+        assert result['type'] == 'boolean'
+        assert result['default'] == 'False'
+        assert result['value'] == 'False'
+
+        await set_category_item_value_entry(category_name='boolean',
+                                            item_name='info', new_value_entry='True')
+        result = await get_category_item(category_name='boolean', item_name='info')
+        assert result['default'] == 'False'
+        assert result['value'] == 'True'
```
praveen-garg: can not this be verified in `async def test_get_category_item_value_entry(self):` itself?

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,653 @@
...
+
+        result = await get_category_item(category_name='boolean', item_name='data')
+        assert result is None
+
+    async def test_get_category_all_items(self):
+        """
+        Test get_category_all_items method returns full "dictionary" of category_value
+        :assert:
+            1.  Values in dictionary are as expected
+            2. default doesn't get updated when value does
```
praveen-garg: again? why?

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,653 @@
...
+
+        result = await get_category_all_items(category_name='boolean')
+        assert result['info']['description'] == 'boolean type with default False'
+        assert result['info']['type'] == 'boolean'
+        assert result['info']['default'] == 'False'
+        assert result['info']['value'] == 'False'
+
+        await  set_category_item_value_entry(category_name='boolean', item_name='info',
+                                             new_value_entry='True')
+        result = await get_category_all_items(category_name='boolean')
```
praveen-garg: again the same test?

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,653 @@
...
+        assert result['info']['default'] == 'False'
+        assert result['info']['value'] == 'True'
+
+    async def test_get_category_all_items_dne(self):
```
praveen-garg: does not look a valid test to me! there can be x many things.

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,653 @@
...
+                               'default': 'ABCabc'}}}
+        }
+
+        for category_name in data:
+            await create_category(category_name=category_name,
+                                  category_description=data[category_name]['category_description'],
+                                  category_value=data[category_name]['category_value'])
+
+        all_category_name = await get_all_category_names()
+        for set in all_category_name:
```
praveen-garg: set is  builtin in python avoid this keyword

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,653 @@
...
+            assert data[set[0].replace(" ", "")]['category_description'] == set[1]
```
praveen-garg: just FYI, you can use `.strip()` to TRIM.

praveen-garg:
Passing to @vaibhav-ScaleDB / @ashwinscale  My feedback is done over https://github.com/foglamp/FogLAMP/pull/185#pullrequestreview-65334319

praveen-garg:
Also, I think `try except` block are redundant there in tests

src/python/tests/test_configuration_manager.py
oshadmon: As long as the information for `category_value` is quoted (""), the code will accept it. As such, I didn't see a reason to add such a test; though it's not a problem to add it.

vaibhav-ScaleDB:
Unit testing should be covering maximum code. This unit test covers only the positive scenarios and no exceptions and negative scenarios are covered. Include them too.
File generated after running:
`~/Development/FogLAMP/src/python/foglamp $ pytest ../tests/test_configuration_manager.py --cov . --cov-report=html`
is attached.
[Coverage for configuration_manager.pdf](https://github.com/foglamp/FogLAMP/files/1336695/Coverage.for.configuration_manager.pdf)

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,653 @@
...
+        await create_category(category_name='boolean', category_description='boolean type',
+                              category_value={'info': {
+                                  'description': 'boolean type with default False',
+                                  'type': 'boolean',
+                                  'default': 'False'}})
+        try:
+            async with aiopg.sa.create_engine(_CONNECTION_STRING) as engine:
+                async with engine.acquire() as conn:
+                    async for result in conn.execute(select_value_stmt):
+                        assert sorted(list(result[0].keys())) == ['info']
```
vaibhav-ScaleDB: Same as above

src/python/tests/test_configuration_manager.py
vaibhav-ScaleDB: @sstefano69 Also, is FOGL-522 the correct jira?

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,653 @@
...
+        await create_category(category_name='boolean', category_description='boolean type',
+                              category_value={'info': {
+                                  'description': 'boolean type with default False',
+                                  'type': 'boolean',
+                                  'default': 'False'}})
+        result = await get_category_item_value_entry(category_name='boolean', item_name='info')
+        assert result == 'False'
+
+        await set_category_item_value_entry(category_name='boolean',
+                                            item_name='info', new_value_entry='True')
```
vaibhav-ScaleDB: You dont need this update as you are checking for get_category_item_value_entry and not for update. As @praveen-garg pointed out, update is already been tested in test_get_set_category_item_value_entry

src/python/tests/test_configuration_manager.py
oshadmon: Since the behavior would change when FOGL-522 is fixed, I don't see the point of adding an `xfail` comment since the test (as it stands) currently passes. Once the test will fail, then putting `xfail`, and a signing someone to fix it  makes sense.

praveen-garg: 1) FOGL-522 is not the correct ticket id.
2) instead assert as per expectation, let it fail xFail reason = bug id ..... and once fixed this test (removal of xfail) pass will ensure bug fixed. TDD

src/python/tests/test_configuration_manager.py
oshadmon: `test_get_category_item_value_entry` (renamed to `test_get_set_category_item_value_entry`) tests  the `get_category_item_value_entry` which returns the value in `statistics.value` json object. While `test_get_category_item` tests `get_category_item`, which the json object for a given `item_name`.

src/python/tests/test_configuration_manager.py
oshadmon: @MarkRiddoch - There is a question of when doing `get_category_all_items` on a non-existent `category_name`, right now returns `None`; however, an error (to me) makes more sense. 

@praveen-garg - Until Mark reply's I'm commenting this test out.

src/python/tests/test_configuration_manager.py
oshadmon: ?

praveen-garg: ignore this, as either of value missing results in error. so OR type check is good enough

oshadmon:
1. @praveen-garg - Fixed changes. 
2. Based on Praveen's comments, I have opened 2 issues: 
- [FOGL-577](https://scaledb.atlassian.net/browse/FOGL-552) which asks for an Error Message when retrieving a `category_name` that doesn't exist 
-   [FOGL-552](https://scaledb.atlassian.net/browse/FOGL-552) which asks for Error Message when setting a `new_value_entry` for a `category_name` that doesn't exist. 
Tests related to either of these issues are currently skipped.

oshadmon:
@vaibhav-ScaleDB 
1. I re-ran the coverage test, and found that the unit-test covers 76%. While I agree that for the most part the 24% missing are the negative tests, there are already ones that do exist. 

2. I ran a performance test (`pytest tests/test_configuration_manager.py --durations=0`) and found that the reason for the "slow" performance is not because of the actual tests, but rather the fact that we need to setup/teardown the database for each test. I'm currently using psql < src/sql/foglamp_ddl.sql` to do that, unless you have better suggestions

3. I removed many of the duplicate tests by using the different `get` commands as part of the `create` and `set` testing.

vaibhav-ScaleDB:
Some of my review comments are not yet addressed, please incorporate these too.
Re:
1. I am not sure what you mean. Is the code coverage report wrong?
2. Yes, better suggestion - do not drop db and create with every test, that will be done when pull #198 is in. In your test setup and cleanup, just create the data you need for the particular test and delete that. That will be much faster. Refer tests/core/api/test_configuration.py for reference. Its deleting only the test data created at class level, you can do the same at method level.
3. Ok

izoratti-dianomic:
@ashwin and @ori - Didn't we agree to use the *real* initialization? 
_foglamp_ init will stay also in the future, if you use the sql file or what @vaibhav-ScaleDB suggested, you will have to modify all your tests in the next sprint. 

I strongly recommend to stick to the way FogLAMP will be released in the next sprint. if it is not possible to do so now, put this branch on hold until the end of the next sprint.

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,560 @@
...
+                                           _configuration_tbl)
+__author__ = "Ori Shadmon"
+__copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
+__license__ = "Apache 2.0"
+__version__ = "${VERSION}"
+
+_CONNECTION_STRING = "dbname='foglamp'"
+
+_CMD = "psql < %s > /dev/null 2>&1" % str(os.popen(
+    "locate foglamp_ddl.sql | grep 'FogLAMP/src/sql'").read()).strip()
```
praveen-garg: I believe,  if we go this way, it should be init not ddl to reset actual data!

src/python/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,560 @@
...
+
+_CONNECTION_STRING = "dbname='foglamp'"
+
+pytestmark = pytest.mark.asyncio
+
+async def delete_from_configuration():
+    """Remove initial data from configuration table"""
+    async with aiopg.sa.create_engine(_CONNECTION_STRING) as engine:
+         async with engine.acquire() as conn:
+             await conn.execute(_configuration_tbl.delete())
```
praveen-garg: @oshadmon  I think this PR can be merged; **if you delete the stuff which you created only.** instead of DELETE all; as we have no provision yet to reset db and actually use test_foglamp db for test. I think for you can have a list of key in data (make it module var) and delete them all.

praveen-garg: also please avoid db execute in loop; seems may of them got fixed. But I still see few references

PR #186, opened 2017-09-19
storage layer client [common table operations]

praveen-garg:


src/python/foglamp/storage/__main__.py
```diff
@@ -0,0 +1,45 @@
...
+
+from foglamp.core.service_registry.service_registry import Service
+
+from foglamp.storage.storage import Storage, Readings
+from foglamp.storage.exceptions import *
+
+# register the service to test the code
+Service.Instances.register(name="store", s_type="Storage", address="blah", port=81)
+
+# discover the Storage type the service: how do we know the instance name?
```
MarkRiddoch: We probably just have a well known name for now. The other option is to get the service registration record by type.

praveen-garg: sure, I will fix that find by name in next PR i.e. for Readings

PR #190, opened 2017-09-22
Fogl 517 - Sequence startup with storage layer [Don't merge]

amarendra-dianomic:
@MarkRiddoch, I have attempted to incorporate all features as discussed regarding FOGL-517. A brief explanation of the code changes is as below:

1. As discussed, now Management API and Admin API are available at two different addresses. T have separated routes for Management API in routes_core.py.

2. Now the startup sequence is as follows:
	- Management API will be loaded as the very first process. In case of errors/exceptions, management process will exit with error code. It will be currently available at http://localhost:8083/foglamp/service.
	
	- Upon successful loading of Management API, Storage services will be loaded next. A separate package foglamp->core->storage_server has been created for this. It contains module storage_server.py which contains a run() method, which will actually run the external storage server. Upon successful start, storage service registers itself with Management API via Rest API vide register_storage() method. However, this external module must have a ping() method which will also send connection info when called and which will finally be used to register itself with Management API. Currently, for demo purpose, register_storage() uses dummy data. For demo purpose, another method test_storage() has been created to simulate a long running external process. This module also contains other methods such start(), stop(), restart(), get_pid(), status(). In case of errors/exceptions, storage process will exit with error code. 
	
	- Upon successful loading of Storage service, foglamp server will be loaded next. In case of errors/exceptions, foglamp process will exit with error code. The Rest server will be available at http://localhost:8082/foglamp as before.

3. Please note that each of the above three processes will create respective pid files in ~/var/run folder. These pid files get deleted upon stopping of the foglamp server.

4. I have checked all the services vide both start methods - "python3 -m foglamp.core" and "foglamp start". Both methods seem to work fine. I have provided console print messages for debugging and verfification which may be removed in final commit.

5. This code essentially demands manual testing only. To test this PR, please proceed as below:
	- In Terminal, open two tabs. In one tab, cd to FogLAMP/src/python, run ". build.sh -c" followed by ". build.sh -i". Run "python3 -m foglamp.core". Wait for few seconds. Then you will see console messages appearing on the screen. Wait till "Testing Storage" message appears.

	- In other tab, you may run "ps -ef". Please note the last lines of the output. You will find lines as:

		asinha   21912  2980  0 15:06 pts/7    00:00:00 foglamp
		asinha   21914 21912  0 15:06 pts/7    00:00:00 management
		asinha   21918 21912  0 15:06 pts/7    00:00:00 [storage] <defunct>
		asinha   21920  2184  0 15:06 pts/7    00:00:00 python3 -m foglamp.core.storage_server

	- Run "ls ~/var/run". You may see as:

		management.pid  storage.pid

	  foglamp.pid appears only with "foglamp start"
	
	- You may want to confirm pid from both the outputs.

	- In first tab, now click CTRL-C. In second tab, all four processes are gone. However, ~/var/run still contains both the pid files.

	- You may want to repeat the steps with "foglamp start". You will notice exactly the same output, different pids of course. Upon "foglamp stop", again output is exactly the same and ALL pids in ~/var/run are also gone.

6. To verify the Management API and Admin API, please proceed as in step 5. In second tab, run curl command as below:
	- $ curl -X GET http://localhost:8083/foglamp/service/ping
	Output is: {"uptime": 554.8963160514832} 

	- $ curl -X GET http://localhost:8082/foglamp/ping
	Output is: {"uptime": 639.5163638591766}

	- $ curl -X GET http://localhost:8083/foglamp/service
	Output is: {"services": [{"port": 8084, "name": "Storage Services 1", "type": "Storage", "protocol": "http", "address": "127.0.0.1", "id": "4c0c87e6-513e-4bd9-ab8b-235a27ac4d70"}]}


src/python/foglamp/core/server_daemon.py
```diff
@@ -132,6 +132,12 @@ def stop(cls, pid=None):
         if not stopped:
             raise TimeoutError("Unable to stop FogLAMP")
 
+        from foglamp.core.server import Server
+        Server.stop_management()
+
+        from foglamp.core.storage_server.storage import Storage
+        Storage.stop()
```
praveen-garg: are these 2 imports (inside def) temporary? and before merge will be fixed?

src/python/foglamp/core/storage_server/__main__.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env python3
```
praveen-garg: @amarendra-dianomic  can you please remove this now? and simply call ./storage  from the build directory itself; add TODO for manaul build and cp process (or may be write a shell script and execute that to do that inside build/run/<>

src/python/foglamp/core/storage_server/storage.py
```diff
@@ -0,0 +1,180 @@
+#!/usr/bin/env python3
```
praveen-garg: remove this now; Please refer comment above.

src/python/foglamp/core/server.py
```diff
@@ -63,6 +75,114 @@ def start(cls):
...
+
+        :rtype: web.Application
+        """
+        core = web.Application(middlewares=[middleware.error_middleware])
+        routes_core.setup(core)
+        return core
+
+    @classmethod
+    def _run_management_api(cls):
+        web.run_app(cls._make_core(), host='0.0.0.0', port=8083)
```
praveen-garg: I believe you should use 8081 (very first) for this

src/python/foglamp/core/server.py
```diff
@@ -63,6 +75,114 @@ def start(cls):
...
+        try:
+            try:
+                from foglamp.core.server_daemon import Daemon
+                Daemon._safe_make_dirs(os.path.dirname(_MANAGEMENT_PID_PATH))
+
+                setproctitle.setproctitle('management')
+                m = Process(target=cls._run_management_api, name='management')
+                m.start()
+
+                # Create storage pid in ~/var/run/storage.pid
```
praveen-garg: fix comment!

src/python/foglamp/core/server.py
```diff
@@ -63,6 +75,114 @@ def start(cls):
...
+                    pid_file.write(str(m.pid))
+
+                # Allow Management core api to start
+                time.sleep(3)
+            except OSError as e:
+                raise Exception("[{}] {} {} {}".format(e.errno, e.strerror, e.filename, e.filename2))
+
+            try:
+                setproctitle.setproctitle('storage')
+                from foglamp.core.storage_server.storage import Storage
```
praveen-garg: this import should be on top!

src/python/foglamp/core/server.py
```diff
@@ -63,6 +75,114 @@ def start(cls):
...
+                raise Exception("[{}] {} {} {}".format(e.errno, e.strerror, e.filename, e.filename2))
+
+            try:
+                setproctitle.setproctitle('foglamp')
+                cls._start()
+            except OSError as e:
+                raise Exception("[{}] {} {} {}".format(e.errno, e.strerror, e.filename, e.filename2))
+
+        except Exception as e:
+            sys.stderr.write(format(str(e)) + "\n");
```
praveen-garg: this must be logged too? or logger is not ready yet?

src/python/foglamp/core/server.py
```diff
@@ -85,3 +205,9 @@ def start(cls):
             task.cancel()
 
         loop.stop()
+
+        cls.stop_management()
+
+        from foglamp.core.storage_server.storage import Storage
```
praveen-garg: import?

src/python/requirements.txt
```diff
@@ -24,3 +24,4 @@ PyJWT==1.5.0
 
 # service
 python-daemon==2.1.2
+setproctitle==1.1.10
```
praveen-garg: needs approval! @izoratti

src/python/tests/core/api/test_configuration.py
```diff
@@ -48,12 +48,13 @@ def setup_class(cls):
         asyncio.get_event_loop().run_until_complete(add_master_data())
         from subprocess import call
         call(["foglamp", "start"])
-        time.sleep(4)
+        time.sleep(10)
```
praveen-garg: needed in this PR?

src/python/tests/core/api/test_scheduler.py
```diff
@@ -55,13 +55,13 @@ def setup_class(cls):
         asyncio.get_event_loop().run_until_complete(add_master_data())
         from subprocess import call
         call(["foglamp", "start"])
-        time.sleep(4)
+        time.sleep(10)
```
praveen-garg: 10 or 5

src/python/tests/core/api/test_task.py
```diff
@@ -63,12 +63,13 @@ def setup_class(cls):
         asyncio.get_event_loop().run_until_complete(add_master_data())
         from subprocess import call
         call(["foglamp", "start"])
-        time.sleep(2)
+        time.sleep(10)
```
praveen-garg: 10 or 5

src/python/tests/core/service_registry/test_services_registry_api.py
```diff
@@ -19,7 +19,7 @@
 
 # Module attributes
 __DB_NAME = "foglamp"
-BASE_URL = 'http://localhost:8082/foglamp'
+BASE_URL = 'http://localhost:8083/foglamp'
```
praveen-garg: better to use 8081?

src/python/foglamp/core/storage_server/storage.py
```diff
@@ -0,0 +1,180 @@
...
+__author__ = "Amarendra K Sinha"
+__copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
+__license__ = "Apache 2.0"
+__version__ = "${VERSION}"
+
+_PID_PATH = os.path.expanduser('~/var/run/storage.pid')
+_WORKING_DIR = os.path.expanduser('~/var/log')
+
+_WAIT_STOP_SECONDS = 5
+"""How many seconds to wait for the core server process to stop"""
```
ashwinscale: fix comment please

src/python/foglamp/core/storage_server/storage.py
```diff
@@ -0,0 +1,180 @@
...
+        logger.setup()
+        cls.logging_configured = True
+
+    @classmethod
+    def run(cls):
+        cls._configure_logging()
+
+        # TODO: Think of some method, other than this, to run the external command within this process, otherwise,
+        # this creates a defunct parent.
+        from subprocess import Popen
```
ashwinscale: It looks like there are three forks to start the storage service:
server.py multiprocess
start forks once
run creates a subprocess

I do see the defunct process in the process list. It seems like we would only need to spawn once. Am I mistaken?

src/python/foglamp/core/storage_server/__main__.py
MarkRiddoch: Please do not do anything with the build.sh scripts now. Ashwin will look at the build process, let's not propagate two ways of doing things. @ashwinscale can you work with Amarendra here please?

src/python/foglamp/core/server.py
```diff
@@ -5,22 +5,34 @@
...
 
-__author__ = "Praveen Garg, Terris Linenbach"
+__author__ = "Praveen Garg, Terris Linenbach, Amarendra K Sinha"
 __copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
 __license__ = "Apache 2.0"
 __version__ = "${VERSION}"
 
 _LOGGER = logger.setup(__name__)  # logging.Logger
+_MANAGEMENT_PID_PATH = os.path.expanduser('~/var/run/management.pid')
+_WORKING_DIR = os.path.expanduser('~/var/log')
```
MarkRiddoch: Why is this the working directory?

src/python/foglamp/core/storage_server/storage.py
```diff
@@ -0,0 +1,180 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+# FOGLAMP_BEGIN
+# See: http://foglamp.readthedocs.io/
+# FOGLAMP_END
+
+"""Starts the FogLAMP Storage service"""
```
MarkRiddoch: I'm confused as to why we have this python code here. I was expecting the storage service to be started simply by the scheduler as at startup. What is thew benefit of these extra layers of process?

amarendra-dianomic:
@MarkRiddoch, @ashwinscale, @praveen-garg, I have incorporated all the changes per feedback except for below:
- Per Ashwin's observation that we should some kind polling or ping mechanism to ascertain readiness of Management API - this will happen after Management API provides a ping method combined with a timeout mechanism. That is why it has been kept as a TODO item. But it may be noted that this is not a blocking item per se.
- To test the Storage service, I have provided a sample python script temp_test_storage.py which tries to emulate a long running process. However, to use real Storage service, we need to specify as to 1) where from we will pick Storage executable, 2) How its dependencies will be resolved. Once this info is provided, I will replace the dummy executable with the real one. Currently this has been marked as a TODO item. 

It may also be noted that,
- Service Registry is now available at port 8081.
- Now, per Mark's obserrvation, the storage_server layer has been removed and all necessary methods have been moved to core/server.py.
- I have rearranged methods in core/server.py for easy reference.
- I have tested ping method successfully with foglamp server by both ways, with and without daemon i.e. both with "python3 -m foglamp.core" and "foglamp start". I have also checked registering a service via curl, inquring the services via curl with both ways and all are working fine.
- Ideally, tests will not be starting/stopping foglamp server as now it becomes a rather complicated process. Hence, the sleep times will be removed once FOGL-521 PR is merged.

src/python/foglamp/core/server.py
```diff
@@ -30,6 +43,197 @@ class Server:
...
+
+        cls._safe_make_dirs(os.path.dirname(_STORAGE_PID_PATH))
+
+        pid = cls.get_storage_pid()
+
+        if pid:
+            print("Storage is already running in PID {}".format(pid))
+        else:
+            from subprocess import Popen, run
+            # TODO: Change below to use real Storage Service
```
ashwinscale: Can we make this an env. variable? @MarkRiddoch @izoratti okay with you?

src/python/foglamp/core/server.py
```diff
@@ -30,6 +43,197 @@ class Server:
...
+            p = Popen('foglamp/core/temp_test_storage.py')
+            # p.poll()
+
+            # Create storage pid in ~/var/run/storage.pid
```
ashwinscale: Please discuss with @MarkRiddoch regarding the use and purpose of pid files and the new directories ~/var/log and ~/var/run dirs. Verify that this is an acceptable approach to process management.

amarendra-dianomic:
@MarkRiddoch @ashwinscale @praveen-garg I have tried to implement startup sequence per new guidelines but have hit below roadblocks:
- I a) created a folder in FogLAMP/services/storage and copy storage, libpostgres.so to this folder, b) created an env var FOGLAMP_ROOT in .bashrc and set it to FogLAMP path. But storage executable does not start from the path i.e. if both the files are not in the current directory (which happens to be src/python).
- Management port has got to be fixed e.g. 8081 so that it is accessible by Storage Services. If we select this port via port discovery, how each subsequent service will come to know about it? 
- In the Storage invocation, what is meant by --port=port and --address=address params? Bcoz with latest build, we still need 8080 for service port and 1081 as management port for Storage, even if we provide some free port info through these command line params. And if we are going to ask Kernel for both service port and management port for Storage, how and where these info can be fed into storage executable?
- Storage is not registering with services_registry.
- There are a few TODO items in server.py and server_daemon.py related to above and some more.
- Tests will be fixed vide a separate PR after merger of this PR.
- However, all three services viz, management, storage and foglamp start well (vide both "foglamp start" and "python3 -m foglamp.core") and respond well to ping methods when both storage executable and libpostgres.so are copied in the current folder. There is a TODO item to fix storage stop method.

src/python/foglamp/core/server.py
```diff
@@ -30,6 +51,300 @@ class Server:
...
+
+    # TODO: Fix a TODO in _start() below to accept below via discovery
+    _RESTAPI_PORT = 8082
+
+    @staticmethod
+    def request_available_port():
+        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        s.bind(('localhost', 0))
+        addr, port = s.getsockname()
+        s.close()
```
ashwinscale: A service (web, storage, etc.) must directly bind to and use a port. With this method, we request an open port, close it, and return the port number. Then we pass this as a config to a service to bind to and use. However, by the time the service uses the port, the port may have been claimed by another service or process.

src/python/foglamp/core/server.py
```diff
@@ -5,23 +5,44 @@
...
 __copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
 __license__ = "Apache 2.0"
 __version__ = "${VERSION}"
 
 _LOGGER = logger.setup(__name__)  # logging.Logger
 
+_FOGLAMP_ROOT = os.getenv('FOGLAMP_ROOT', '/home/asinha/Development/FogLAMP')
+_STORAGE_PATH = '.' #os.path.expanduser(_FOGLAMP_ROOT+'/services/storage')
+
+_MANAGEMENT_PID_PATH = os.getenv('MANAGEMENT_PID_PATH', os.path.expanduser('~/var/run/management.pid'))
```
ashwinscale: @amarendra-dianomic I see that PID files are being used for process management. @MarkRiddoch please clarify how much/little process management should be done here.

src/python/foglamp/core/server.py
```diff
@@ -30,6 +51,300 @@ class Server:
...
+        print("Starting Management API")
+        try:
+            cls._safe_make_dirs(os.path.dirname(_MANAGEMENT_PID_PATH))
+            setproctitle.setproctitle('management')
+            # Process used instead of subprocess as it allows a python method to run in a separate process.
+
+            # TODO: Investigate if below line is required and remove it if _MANAGEMENT_API_PORT is going to be fixed
+            # cls._MANAGEMENT_API_PORT = cls.request_available_port()
+
+            m = Process(target=cls._run_management_api, name='core', args=(cls._MANAGEMENT_API_PORT,))
```
ashwinscale: @MarkRiddoch please verify if management API is supposed to run as a separate process.

src/python/foglamp/core/server.py
```diff
@@ -5,23 +5,44 @@
...
 
-__author__ = "Praveen Garg, Terris Linenbach"
+__author__ = "Praveen Garg, Terris Linenbach, Amarendra K Sinha"
 __copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
 __license__ = "Apache 2.0"
 __version__ = "${VERSION}"
 
 _LOGGER = logger.setup(__name__)  # logging.Logger
 
+_FOGLAMP_ROOT = os.getenv('FOGLAMP_ROOT', '/home/asinha/Development/FogLAMP')
```
MarkRiddoch: We should push this with somebodies home directory included in the code

src/python/foglamp/core/server.py
MarkRiddoch: We can get rid of all the pid files and the like from within the code as snap has its own way of dealing with this. In non-san environments we will wrap it up inside init.d scripts

src/python/foglamp/core/server.py
```diff
@@ -30,6 +51,300 @@ class Server:
     scheduler = None
     """ foglamp.core.Scheduler """
 
+    logging_configured = False
+    """Set to true when it's safe to use logging"""
+
+    # TODO: Discuss if this needs to be pre-decided and fixed
+    _MANAGEMENT_API_PORT = 8081
```
MarkRiddoch: This does not need to be fixed, we should be using port 0

src/python/foglamp/core/server.py
```diff
@@ -30,6 +51,300 @@ class Server:
...
+
+    # TODO: Set below storage ports to None after Storage layer is fixed to accept these values via command line arguments
```
MarkRiddoch: This should be in place.

src/python/foglamp/core/server.py
MarkRiddoch: No, management API is not a separate process

src/python/foglamp/core/server.py
```diff
@@ -30,6 +51,300 @@ class Server:
...
+
+        return pid
+
+    @classmethod
+    def _start_management_service(cls):
+        # Start Management API
+        print("Starting Management API")
+        try:
+            cls._safe_make_dirs(os.path.dirname(_MANAGEMENT_PID_PATH))
+            setproctitle.setproctitle('management')
```
MarkRiddoch: What is this for?

src/python/foglamp/core/server.py
```diff
@@ -30,6 +51,300 @@ class Server:
...
+        if not stopped:
+            raise TimeoutError("Unable to stop Management API")
+
+        print("Management API stopped")
+
+
+
+    """ Storage Services """
+    @staticmethod
+    def _get_storage_pid():
```
MarkRiddoch: Why are we looking at the pid?

src/python/foglamp/core/server.py
```diff
@@ -30,6 +51,300 @@ class Server:
...
+    def _start_storage(cls):
+        """Starts Storage"""
+        cls._safe_make_dirs(os.path.dirname(_STORAGE_PID_PATH))
+        pid = cls._get_storage_pid()
+        if pid:
+            print("Storage is already running in PID {}".format(pid))
+        else:
+            # TODO: Uncomment below after Storage layer is fixed to accept below via command line params
+            # cls._STORAGE_PORT = cls.request_available_port()
+            p = subprocess.Popen([_STORAGE_PATH+'/storage', '--port={}'.format(cls.request_available_port()), '--address=localhost'])
```
MarkRiddoch: The port here should be the port of the core's management API

src/python/foglamp/core/server.py
```diff
@@ -30,6 +51,300 @@ class Server:
...
+            # Create storage pid in ~/var/run/storage.pid
+            with open(_STORAGE_PID_PATH, 'w') as pid_file:
+                pid_file.write(str(p.pid))
+
+    @classmethod
+    def _start_storage_service(cls):
+        # Start Storage Service
+        print("Starting Storage Services")
+        try:
+            setproctitle.setproctitle('storage')
```
MarkRiddoch: Why are we doing this?

src/python/foglamp/core/server_daemon.py
```diff
@@ -17,18 +17,18 @@
 import sys
 import time
 import daemon
+import setproctitle
```
MarkRiddoch: I would suggest not using this

src/python/foglamp/core/server_daemon.py
```diff
@@ -17,18 +17,18 @@
...
 
 __author__ = "Amarendra K Sinha, Terris Linenbach"
 __copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
 __license__ = "Apache 2.0"
 __version__ = "${VERSION}"
 
-_PID_PATH = os.path.expanduser('~/var/run/foglamp.pid')
-_WORKING_DIR = os.path.expanduser('~/var/log')
+_PID_PATH =  os.getenv('FOGLAMP_PID_PATH', os.path.expanduser('~/var/run/foglamp.pid'))
+_WORKING_DIR = os.getenv('WORKING_DIR', os.path.expanduser('~/var/log'))
```
MarkRiddoch: We should not create this, it doesn't fit within the file structure as defined by Ivan

src/python/foglamp/core/server_daemon.py
```diff
@@ -86,16 +86,24 @@ def start(cls):
...
-            # os.getpid() reports the wrong pid so it's not easy.
-            print("Starting FogLAMP")
-
-            with daemon.DaemonContext(
-                working_directory=_WORKING_DIR,
-                umask=0o002,
-                pidfile=pidfile.TimeoutPIDLockFile(_PID_PATH)
-            ):
-                cls._start_server()
+            try:
```
MarkRiddoch: I don't think we need any of the daemon stuff related to pid files etc as it will be taken care of by the snap daemon.

src/python/foglamp/core/server.py
ashwinscale: @MarkRiddoch to rename a process... for example: it is also being used to rename core process to foglamp

src/python/requirements.txt
izoratti-dianomic: I don't this we should use this. Please note that we will remove the daemon run of Python: this has been a wrong implementation that we have carried on since we started the project and we must fix asap. As I explained during the scrum yesterday, FogLAMP will be governed by the foglamp bash script (in snap or in a standard installation), so the process will be spawned by the script when needed. I looked at the code, I don't think the changes applied and the use of the library is necessary (like here: https://github.com/foglamp/FogLAMP/blob/FOGL-517/src/python/foglamp/core/server_daemon.py#L95).

We should remove this.

amarendra-dianomic:
@MarkRiddoch, @ashwinscale, @praveen-garg as presented by Mark and also explained by Ashwin, mu understanding of the correct sequence startup is as follows:
- Start management API on a dynamic management api port (by binding to 0 and getting a free port from Kernel).
- Next Start storage passing management api port as a command line argument. Storage registers itself with Management API with a dynamic service port and a dynamic management port.
- Next Start scheduler (pass in the management port as arg). This will help starting each subsequent microservice with the given management port (we will need to modify each task code for it).

Similarly, the correct shutdown procedure is as follows:
- Block Management API from accepting new requests. Then fetch the list of running microservices from Management API. Shudown each microservice, except Storage, by calling its shutdown method via Rest call.
- Shutdown Scheduler.
- Shutdown Storage by calling its shutdown method via Rest call.

I have tried to implement each of the above step in the revised commits. A summary of the new features added is as below:
- Now there are only two processes running - one process running Management API at a dynamic port + Rest Server at a fixed port 8082 + Scheduler, and another process running Storage Services.
- I have removed deamon module from server_daemon.py. Now the Foglamp server starts in daemon mode without use of Daemon module as it was interferring with our core Server code.
- Now code has been simplified a lot. Also, I have done a preliminary testing of the startup and shutdown sequence and all seem to be working fine. Storage also self registers. I have confirmed via curl. However, we may need to do some detailed testing to reconfirm the same.
- As reported by @ashwinscale, now registering a new service does not stop foglamp server when started by "foglamp start".  
- There are TODO items and also need to check/confirm SYSLOGs. We also need to modify Scheduler code to associate management api port to each STARTUP task.

amarendra-dianomic:
@MarkRiddoch @ashwinscale I have also noted that Praveen has also submitted a [PR#224](https://github.com/foglamp/FogLAMP/pull/224) for the Squence startup and has approached the solution a bit differently.

praveen-garg:
@amarendra-dianomic 

1) I have removed deamon module from server_daemon.py. Now the Foglamp server starts in daemon mode without use of Daemon module as it was interferring with our core Server code.

2) service shutdown / ping 

3) kill signal binding and shutdown/ clean up of aiohttp

Let's handle these 3 things in 3 sepreate tickets. I believe #2 shall be out next priority if #229 gets merged. and surely this PR helped and will help in these 3 items too. As there is some code messup, cleaning and simplifying the code with smaller def units is also becoming the right priority.

izoratti:
@praveen-garg and @amarendra-dianomic (+ @MarkRiddoch and @ashwinscale)

Great to see that we are making progress, it looks to me we are almost there!

There is an important point that we have discussed many times but we have never grasped: Python must not have any daemon-specific code in it, we never planned to have it. The simple reason is that FogLAMP is not a Python-only app, we must to have a common way to handle all the modules. This way, as it happens in server-side systems, is through management scripts, generally written in shell. 

The plan is to remove the legacy daemon code that conflicts with the bash scripts in the next sprint, so please do not add any new one - on the contrary, if you can remove it while you are working on the current issue, it will simplify the next sprint.

To be more specific this is what is going to happen:
1. FogLAMP is handled by a bash script. The script manages the launch of all modules (C binaries, Python or others in the future) using the typical interface used by Linux services. The interface includes the options of start/stop/status, but also some extra tasks, when required.
2. Python code will be called as regular application with stdin/stdout control, unless the script starts it in background.
3. The installation via snap makes FogLAMP to work as a service at startup. In the same way, it handles the graceful stopping of the services at shutdown.
4. When administrators use the source code to install FogLAMP as a service, they will adopt the usual approach via systemctl/systemd.

I hope this clarifies the issue.

izoratti-dianomic:
The previous comment came from my personal GH account... sorry.

praveen-garg:
Hi @izoratti  Agree!

Please see https://github.com/foglamp/FogLAMP/pull/229 which just take care of start up sequence. And does not add anything like process id, daemon adventure etc. 

PR 229 needs 2 fixes, before final merge:
1) FOGLAMP ROOT env var
2) add `services/storage` with `.gitkeep` so that storage build process can put the static files in it

regarding fixing/ removal of daemon, let's handle it as a separate ticket; As I mentioned in the previous comment, https://github.com/foglamp/FogLAMP/pull/190#issuecomment-336679199

amarendra-dianomic:
Closed in favor of PR#229

PR #199, opened 2017-09-29
Fogl 520 readings

praveen-garg:


praveen-garg:
@MarkRiddoch  please review. 
All common table and Readings operations are in; with basic checks in `__main__.py` and payload_builder is also testing the same.

PR #201, opened 2017-09-29
Fogl 521 - Port scheduled tasks to use storage service

amarendra-dianomic:


praveen-garg:
let's rename this branch to payload builder ticket FOGL-581

src/python/foglamp/storage/payload_builder.py
```diff
@@ -0,0 +1,217 @@
...
+
+    def __init__(self, payload=OrderedDict()):
+        self.payload = payload
+
+    @staticmethod
+    def verify_condition(arg):
+        retval = False
+        if isinstance(arg, list):
+            if len(arg) == 3:
+                if arg[1] in ['<', '>', '=', '>=', '<=', 'LIKE', 'IN', '!=']:
```
MarkRiddoch: LIKE and IN are not currently supported by the storage service, these can be added if we need them

PR #203, opened 2017-09-29
Fogl 589 - Create Payload Builder (SubTask 1 of FOGL-581)

amarendra-dianomic:


amarendra-dianomic:
Linked to [PR#201](https://github.com/foglamp/FogLAMP/pull/201)

src/python/foglamp/storage/payload_builder.py
```diff
@@ -0,0 +1,217 @@
...
+
+    def __init__(self, payload=OrderedDict()):
+        self.payload = payload
+
+    @staticmethod
+    def verify_condition(arg):
+        retval = False
+        if isinstance(arg, list):
+            if len(arg) == 3:
+                if arg[1] in ['<', '>', '=', '>=', '<=', 'LIKE', 'IN', '!=']:
```
praveen-garg: **Pasting Mark's message:**
 MarkRiddoch :
 LIKE and IN are not currently supported by the storage service, these can be added if we need them

src/python/foglamp/storage/payload_builder.py
```diff
@@ -0,0 +1,217 @@
...
+    def HAVING(self):
+        # TODO: To be implemented
+        return self
+
+    def LIMIT(self, arg):
+        if isinstance(arg, int):
+            self.payload.update({"limit": arg})
+        return self
+
+    def ORDERBY(self, *args):
```
praveen-garg: ORDER_BY

src/python/foglamp/storage/payload_builder.py
```diff
@@ -0,0 +1,217 @@
...
+                        self.payload['sort'].append(sort)
+                    else:
+                        self.payload['sort'] = list(self.payload.get('sort'))
+                        self.payload['sort'].append(sort)
+                else:
+                    self.payload.update({"sort": sort})
+        return self
+
+    def execute(self):
+        return dict(self.payload)
```
praveen-garg: rename to payload, see above (constructor message)
it should return json.dumps

src/python/foglamp/storage/payload_builder.py
```diff
@@ -0,0 +1,217 @@
...
+
+    def execute_query_params(self):
```
praveen-garg: rename to query_params (sans execute)

src/python/foglamp/storage/payload_builder.py
```diff
@@ -0,0 +1,217 @@
...
+        where = self.payload['where']
+        query_params = {where['column']: where['value']}
+        for key, value in where.items():
+            if key == 'and':
+                query_params.update({value['column']: value['value']})
+        return urllib.parse.urlencode(query_params)
+
+if __name__ == "__main__":
+    pb = PayloadBuilder(payload=OrderedDict())
```
praveen-garg: 1) better if don;t need to make it an object (should optional)
2) pb = PayloadBuilder() will work too, yes? as payload=OrderedDict() is default arg

src/python/foglamp/storage/payload_builder.py
```diff
@@ -0,0 +1,217 @@
...
+        FROM('schedules').\
+        WHERE(['id', '=', 'test']).\
+        WHERE_AND(['process_name', '=', 'test']). \
+        WHERE_OR(['process_name', '=', 'sleep']).\
+        LIMIT(3).\
+        GROUPBY('process_name', 'id').\
+        ORDERBY(['process_name', 'desc']).\
+        AGGREGATION(['count', 'process_name']).\
+        execute()
+    print(str(sql).replace("'", '"'))
```
praveen-garg: do we need this for json.dumps too? if Yes, can we just move to execute() [to be renamed to paylaod()] itself?

src/python/foglamp/storage/payload_builder.py
```diff
@@ -0,0 +1,217 @@
...
+    pb = PayloadBuilder(payload=OrderedDict())
+    # Insert
+    sql = pb.\
+        INSERTINTO('schedules').\
+        INSERT(id='test', process_name='sleep', type=3, repeat=45677).\
+        execute()
+    print(str(sql).replace("'", '"'))
+
+    pb = PayloadBuilder(payload=OrderedDict())
+    # Insert
```
praveen-garg: update

amarendra-dianomic:
@praveen-garg all changes done as advised. All changes tested successfully with examples in the code, However, pl note that:
- I doubt we can chain methods within an instance, which is very convenient feature, if we change methods to @classmethod. Because in this case, methods would return the class and NOT the instance.
- Some suggestions could not be done due to some errors for which I have added @FIXME.
- I have added a kind of very basic schema validator. We can expand on this later.
- Only task left is adding tests. Please advise @vaibhav-ScaleDB accordingly as you have suggested earlier otherwise I will take up this next week.

PR #206, opened 2017-10-04
FOGL-491 improvement

praveen-garg:
@oshadmon  please see this; now it does not matter init data is  there or not, or even if we play with dev DB; As we are deleting our created data only!


praveen-garg:
Merge this and pull/ fix in your branch itself. then squash as there tons of commits now in FOGL-491 branch.

PR #211, opened 2017-10-04
Fogl 521 - Port scheduled tasks to use storage service

amarendra-dianomic:


amarendra-dianomic:
Rewok needed after merger of FOGL-517 PR.

PR #215, opened 2017-10-05
[#no-ticket] adding service management port with TESTS

praveen-garg:


src/python/foglamp/core/service_registry/service_registry.py
```diff
@@ -133,18 +146,23 @@
             "type": service._type,
             "address": service._address,
             "port": service._port,
+            "management_port": service._management_port,
```
MarkRiddoch: This does not conform to the conventions we use. It would be managementPort

src/python/foglamp/core/service_registry/service_registry.py
```diff
@@ -133,18 +146,23 @@
             "type": service._type,
             "address": service._address,
             "port": service._port,
```
MarkRiddoch: Changing this to servicePort is changing the meaning of port for existing code and also means it now is not in line with the specification. If we are going to break the existing code we should make this servicePort and not port and then update the uses of this and the documentation.

src/python/foglamp/core/service_registry/service_registry.py
praveen-garg: I believe, for input keys we are **snake_case**

praveen-garg: and this also aligns with python naming convention.

src/python/foglamp/core/service_registry/service_registry.py
praveen-garg: **.port** should be clear itself in terms of service's port. but if we want to change with some diff name, please open the ticket and we can change to that word

PR #217, opened 2017-10-06
Fogl 602 poll mode - JIRA FOGL-588 Add poll mechanism to the device service

sstefano69:


src/python/foglamp/device/__main__.py
```diff
@@ -5,36 +5,88 @@
...
+    # Information messages
+    "i000000": "",
+
+    # Warning / Error messages
+    "e000000": "generic error.",
+    "e000001": "cannot proceed the execution, invalid device service name - service name |{0}|  error details |{1}|",
+    "e000002": "cannot proceed the execution, missed device service name",
+    "e000003": "cannot complete the operation - error details |{0}|",
+}
+""" Messages used for Information, Warning and Error notice """
```
praveen-garg: @sstefano69 It will be great to have exceptions.py as src/python/foglamp/storage/exceptions.py with code and message and log while catching / raising these exceptions; eventually better will be to allow each package has a util(s).py and exception(s).py including root python package

src/python/foglamp/device/exceptions.py
```diff
@@ -0,0 +1,27 @@
...
+
+""" Device service exceptions module """
+
+__author__ = "Stefano Simonelli"
+__copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
+__license__ = "Apache 2.0"
+__version__ = "${VERSION}"
+
+
+class InvalidServiceNameError(Exception):
```
praveen-garg: let's make a  def DeviceServiceError or DeviceServiceException with init having code and message? and subclass each next with that?

praveen-garg:
@sstefano69  could you please rebase with develop, @ashwinscale  please review w.r.t. https://github.com/foglamp/FogLAMP/pull/250

PR #220, opened 2017-10-09
FOGL-552 - Missing expected error when setting a value to a non-existent category

ashish-ScaleDB:
[FOGL-552](https://scaledb.atlassian.net/browse/FOGL-552)

**Fixed items:**
* validation of non-existence category_name and item_name added in _set_category_item_value_entry_
* **xfail annotation** removed from _test_set_category_item_value_error_ and updated test accordingly.
* FYI, **'value**' key added as test_data in  _test_configuration.py_ as needed here we are inserting direct DB insertion. **Note** : **create_category** method automatically created *value* key  and its value with exactly default value if it does not pass in json as test_data. See below code snippet:

```
test_data = {"item2": { "description": "desc", "type": "string", "default": "def"} }
await create_category('TESTAPI2', test_data, 'RESTAPI Test Config')
json = await get_category_all_items('TESTAPI2')
print(json)
```

**O/P:** - "item2": {"description": "desc", "type": "string", "default": "def", **"value":"def"** }

_No Regression takes place with the change._

**See TESTS report below**
1) unit

```
collected 8 items 
tests/core/api/test_configuration.py ......s.
===================================================== 7 passed, 1 skipped in 6.34 seconds
```
2) Rest API

```
collected 24 items 
tests/test_configuration_manager.py ................X.XXX...
==================================================== 20 passed, 4 xpassed in 1.39 seconds
```


ashish-ScaleDB:
Essentially we will get rid of these hard-coded sleep time in tests soon, once JIRA [486](https://scaledb.atlassian.net/browse/FOGL-486) and [510](https://scaledb.atlassian.net/browse/FOGL-510) fixed

PR #221, opened 2017-10-09
port name and service.service_port is optional now

praveen-garg:


praveen-garg:
service's service port is optional, please see https://github.com/foglamp/FogLAMP/pull/221/files#diff-22b177034a5d24e6daab6996668bd34bR60

praveen-garg:
should we make  src/C/foglamp/common/service_record.cpp key changes in this PR itself?

PR #224, opened 2017-10-11
start up sequence fixed! [Do not merge]

praveen-garg:
see https://github.com/foglamp/FogLAMP/pull/229 for polished version; As this PR has many prints and commented code to understand the issues encountered and to show help text while running / testing


src/python/foglamp/core/server.py
```diff
@@ -40,15 +43,70 @@ def _make_app():
...
+        # https://aiohttp.readthedocs.io/en/stable/_modules/aiohttp/web.html#run_app
+        # web.run_app(cls._make_core_app(), host='0.0.0.0', port=8082)
+
+        # TODO: make management port dynamic?!
+        host_address = "localhost"  # fix?
+        core_mgt_port = 8082
+
+        ma = MultiApp()
+        ma.configure_app(cls._make_core_app(), host=host_address, port=core_mgt_port)
+        service_port = cls.request_available_port()
```
ashwinscale: request_available_port should not be used (because a port is only temporarily claimed in this way). instead, let the web server bind to port 0, it will claim a free port. then we ask which port the web server claimed.

praveen-garg:
Please put your comments what needs to be chnaged as skeleton  / stubs to be implemneted or if TODO comments are incorrect or need something else too

src/python/foglamp/core/server.py
```diff
@@ -59,12 +129,36 @@ def start(cls):
                 signal_name,
                 lambda: asyncio.ensure_future(cls._stop(loop)))
 
+        # TODO: make management port dynamic?!
+        host_address = "localhost"  # fix?!
+        core_mgt_port = 8082
+
+        # start scheduler
```
ashwinscale: Aren't we supposed to start the management webserver first? Otherwise, how do we start the scheduler?

src/python/foglamp/core/server.py
```diff
@@ -59,12 +129,36 @@ def start(cls):
                 signal_name,
                 lambda: asyncio.ensure_future(cls._stop(loop)))
 
+        # TODO: make management port dynamic?!
+        host_address = "localhost"  # fix?!
+        core_mgt_port = 8082
```
ashwinscale: This port MUST be dynamic. Therefore, the management port MUST be started before starting the storage.

src/python/foglamp/core/server.py
```diff
@@ -59,12 +129,36 @@ def start(cls):
...
+        # start scheduler
         # The scheduler must start first because the REST API interacts with it
         loop.run_until_complete(asyncio.ensure_future(cls._start_scheduler()))
 
-        # https://aiohttp.readthedocs.io/en/stable/_modules/aiohttp/web.html#run_app
-        web.run_app(cls._make_app(), host='0.0.0.0', port=8082, handle_signals=False)
+        loop.create_task(cls.start_storage(loop, host_address, core_mgt_port))
+        # loop.call_soon(asyncio.ensure_future(cls._start_storage(host_address, core_mgt_port)))
+
+        cls._start_core(host=host_address, management_port=core_mgt_port)
```
ashwinscale: This has to happen first.

praveen-garg: Please check via `python -m foglamp.core` or see syslog. Its being started first. 
I did not fix the scheduler line as it had comments as "MUST start before"; I'm not sure if it was designed per that architecture. we can simply move that to after storage (which I did not intentionally put yet) as `loop.create_task(cls._start_scheduler())` just next task after the `start storage` to be started in event loop ; loop  starts with start_core hence first thing.

src/python/foglamp/core/server.py
praveen-garg: and service port?

src/python/foglamp/core/http_server.py
```diff
@@ -0,0 +1,92 @@
...
+
+https://stackoverflow.com/questions/44850701/multiple-aiohttp-applications-running-in-the-same-process
+Added host to constructor
+"""
+
+import asyncio
+from aiohttp import web
+
+
+class AppWrapper:
```
MarkRiddoch: I would ideally like to see more documentation in here as this was clearly an issue that caused problems. From a maintenance perspective it is crucial that this is well understood

src/python/foglamp/core/server.py
```diff
@@ -40,15 +48,77 @@ def _make_app():
...
+        # https://aiohttp.readthedocs.io/en/stable/_modules/aiohttp/web.html#run_app
+        # web.run_app(cls._make_core_app(), host='0.0.0.0', port=8082)
+
+        ma = MultiApp()
+        ma.configure_app(cls._make_core_app(), host=host, port=management_port)
+
+        # TODO: fetch from app info
+        # to register to core; as there is no update mechanism so just register once
+        # with service port
+        service_port = cls.request_available_port()
```
MarkRiddoch: This request available port is a little dangerous as the port may be taken in the interim, why can't we simply bind to port 0?

src/python/foglamp/core/server.py
MarkRiddoch: I suspect the way this works is that the create_task does not actually run something immediately, but rather some time later its the event loop. Of course this is dangerous if we change some in the start_core that ends up executing the event loop.

MarkRiddoch:
Please note there is no "ping required to wait" - the rest of the start up process proceeds when the registration arrives from the storage, please see micro services management document.

src/python/foglamp/core/server.py
```diff
@@ -40,15 +52,73 @@ def _make_app():
...
+        ma.run_all()
+
+    @classmethod
+    def _register_core(cls, host, mgt_port, service_port):
+        core_service_id = Service.Instances.register(name="FogLAMP Core", s_type="Core", address=host,
+                                                     port=service_port, management_port=mgt_port)
+
+        return core_service_id
+
+    # TODO: remove me | NOT NEEDED (hopefully, we shall be able to get info back from aiohttp)
```
MarkRiddoch: Yes, the port can be determined after the fact. See previous code posted in slack that does this.

praveen-garg: loop.run_forever()

praveen-garg:
@MarkRiddoch @ashwinscale @amarendra-dianomic removed multiapp hack!  Used the code Amarendra had and Mark suggested. Please see each code comment https://github.com/foglamp/FogLAMP/pull/224 ; 

**: pending**

removal of print, commented code, more polishing

src/python/foglamp/core/http_server.py
```diff
@@ -0,0 +1,103 @@
+# -*- coding: utf-8 -*-
```
praveen-garg: this file is no more needed!

src/python/foglamp/core/scheduler.py
```diff
@@ -1618,13 +1624,49 @@ def _check_purge_tasks(self):
...
+        # make sure that it go forward only when storage service is ready
+        storage_service = None
+
+        while storage_service is None:  # TODO: wait for x minutes?
+            try:
+                found_services = Service.Instances.get(name="FogLAMP Storage")
+                storage_service = found_services[0]
+            except Exception:
+                await asyncio.sleep(5)
+        self._logger.info("Starting scheduler; mgt port received is %d", self._core_management_port)
```
praveen-garg: to make it a PURE client sdk type thing, let's check via api

src/python/foglamp/core/server.py
```diff
@@ -59,12 +193,32 @@ def start(cls):
...
                 lambda: asyncio.ensure_future(cls._stop(loop)))
 
-        # The scheduler must start first because the REST API interacts with it
-        loop.run_until_complete(asyncio.ensure_future(cls._start_scheduler()))
+        host_address = cls._HOST  # fix?!
 
-        # https://aiohttp.readthedocs.io/en/stable/_modules/aiohttp/web.html#run_app
-        web.run_app(cls._make_app(), host='0.0.0.0', port=8082, handle_signals=False)
+        cls._start_core_and_user(loop=loop, host=host_address,
+                                 service_port=cls._API_SERVICE_PORT, management_port=cls._MANAGEMENT_PORT)
```
praveen-garg: cls._MANAGEMENT_PORT is 0 when started

PR #226, opened 2017-10-12
FOGL-582 Integration tests for storage client

vaibhav-ScaleDB:
14 passed, 8 skipped in 9.87 seconds


src/python/tests/storage/test_storage_api.py
```diff
@@ -0,0 +1,265 @@
+# -*- coding: utf-8 -*-
+
+# FOGLAMP_BEGIN
+# See: http://foglamp.readthedocs.io/
+# FOGLAMP_END
+import pytest
+import os
+import py
+from foglamp.storage.payload_builder import PayloadBuilder
+
```
ashish-ScaleDB: Move all imports under __version__  OR L#16 move up

src/python/tests/storage/foglamp_test_storage_init.sql
```diff
@@ -0,0 +1,4 @@
+DELETE FROM foglamp.statistics;
+INSERT INTO foglamp.statistics ( key, description, value, previous_value )
+    VALUES ( 'TEST_1',   'Testing the storage service data 1', 10, 2 ),
+           ( 'TEST_2',   'Testing the storage service data 2', 15, 2 );
```
ashish-ScaleDB: Add comments here as well - "once FOGL-510 is done", we should delete this file

src/python/tests/storage/test_storage_api.py
```diff
@@ -0,0 +1,265 @@
...
+    _dir = os.path.dirname(os.path.realpath(__file__))
+    file_path = py.path.local(_dir).join('/foglamp_test_storage_init.sql')
+    os.system("psql < {} > /dev/null 2>&1".format(file_path))
+    yield
+    os.system("psql < `locate foglamp_ddl.sql | grep 'FogLAMP/src/sql'` > /dev/null 2>&1")
+    os.system("psql < `locate foglamp_init_data.sql | grep 'FogLAMP/src/sql'` > /dev/null 2>&1")
+
+
+@pytest.allure.feature("api")
+@pytest.allure.story("storage")
```
ashish-ScaleDB: @praveen-garg any thoughts on this naming?

src/python/tests/storage/test_storage_api.py
```diff
@@ -0,0 +1,265 @@
...
+class TestStorageRead:
```
ashish-ScaleDB: Add brief description.

src/python/tests/storage/test_storage_api.py
```diff
@@ -0,0 +1,265 @@
...
+    def test_select(self):
+        res = Storage().query_tbl_with_payload("statistics", PayloadBuilder().SELECT().payload())
```
ashish-ScaleDB: res variable => **result**

src/python/tests/storage/test_storage_api.py
```diff
@@ -0,0 +1,265 @@
...
+        assert res["rows"][0]["description"] == "Testing the storage service data 1"
+        assert res["rows"][0]["value"] == 10
+        assert res["rows"][0]["previous_value"] == 2
+
+    @pytest.mark.skip(reason="FOGL-615")
+    def test_where_invalid_key(self):
+        res = Storage().query_tbl_with_payload("statistics", PayloadBuilder().WHERE(["bla", "=", "invalid"]).payload())
+        assert "ERROR" in res["message"]
+
+    @pytest.mark.skip(reason="Payload builder does not parse more than 1 AND_WHERE correctly")
```
ashish-ScaleDB: Better to reason with ticket id FOGL-607#7

src/python/tests/storage/test_storage_api.py
```diff
@@ -0,0 +1,265 @@
...
+                                               AND_WHERE(["key", "=", "TEST_1"]).
+                                               AND_WHERE(["previous_value", "<", 10]).payload())
+        assert len(res["rows"]) == 1
+        assert res["count"] == 1
+        assert res["rows"][0]["key"] == "TEST_1"
+        assert res["rows"][0]["description"] == "Testing the storage service data 1"
+        assert res["rows"][0]["value"] == 10
+        assert res["rows"][0]["previous_value"] == 2
+
+    @pytest.mark.skip(reason="Payload builder does not parse more than 1 OR_WHERE correctly")
```
ashish-ScaleDB: Better to reason with ticket id FOGL-607#7

src/python/tests/storage/test_storage_api.py
```diff
@@ -0,0 +1,265 @@
...
+    def test_limit(self):
+        res = Storage().query_tbl_with_payload("statistics", PayloadBuilder().LIMIT(1).payload())
+        assert len(res["rows"]) == 1
+        assert res["count"] == 1
+        assert res["rows"][0]["key"] == "TEST_1"
+        assert res["rows"][0]["description"] == "Testing the storage service data 1"
+        assert res["rows"][0]["value"] == 10
+        assert res["rows"][0]["previous_value"] == 2
+
+    @pytest.mark.skip(reason="Payload builder does not support offset without limit")
```
ashish-ScaleDB: Better to reason with ticket id FOGL-607#4

src/python/tests/storage/test_storage_api.py
```diff
@@ -0,0 +1,265 @@
...
+
+    def test_limit_offset(self):
+        res = Storage().query_tbl_with_payload("statistics", PayloadBuilder().LIMIT(2).OFFSET(1).payload())
+        assert len(res["rows"]) == 1
+        assert res["count"] == 1
+        assert res["rows"][0]["key"] == "TEST_2"
+        assert res["rows"][0]["description"] == "Testing the storage service data 2"
+        assert res["rows"][0]["value"] == 15
+        assert res["rows"][0]["previous_value"] == 2
+
```
ashish-ScaleDB: Should we add test_skip and test_limit_skip?

src/python/tests/storage/test_storage_api.py
```diff
@@ -0,0 +1,265 @@
...
+    def test_order(self):
+        res = Storage().query_tbl_with_payload("statistics", PayloadBuilder().
+                                               ORDER_BY(["key", "desc"]).LIMIT(1).payload())
```
ashish-ScaleDB: Please remove .LIMIT(1) and check only first index of result

src/python/tests/storage/test_storage_api.py
```diff
@@ -0,0 +1,265 @@
...
+        assert res["rows"][0]["previous_value"] == 2
+
+    def test_aggregate(self):
+        res = Storage().query_tbl_with_payload("statistics", PayloadBuilder().
+                                               AGGREGATE(["max", "value"]).payload())
+        assert len(res["rows"]) == 1
+        assert res["count"] == 1
+        assert res["rows"][0]["max_value"] == 15
+
+    @pytest.mark.skip(reason="Storage does not support GROUP_BY without aggregate")
```
ashish-ScaleDB: Please file a bug and add ticket here

src/python/foglamp/storage/payload_builder.py
```diff
@@ -210,96 +210,3 @@ def query_params(cls):
...
-        print(str(ex))
-
-    limit = PayloadBuilder().LIMIT(2).payload()
-    print(limit)
-
-    limit_and_offset = PayloadBuilder().LIMIT(2).OFFSET(10).payload()
-    print(limit_and_offset)
-
-    limit_and_skip = PayloadBuilder().LIMIT(2).SKIP(12).payload()
-    print(limit_and_skip)
```
ashish-ScaleDB: Remove 1 line more in end of file

src/python/tests/storage/test_storage_api.py
vaibhav-ScaleDB: Both skip and offset generates the same JSON payload. Purpose of these tests was to see the integration with storage layer. Since we have already tested that skip and offset generates the same json in test_payload_builder.py there was no point to test the same payload with storage layer again. That is why skipped

src/python/tests/storage/foglamp_test_storage_init.sql
vaibhav-ScaleDB: No, We wont delete this. We cant insert test data to test storage layer via storage layer. FOGL-510 will migrate fixtures at a common place only.

src/python/tests/storage/test_storage_api.py
```diff
@@ -0,0 +1,283 @@
...
+        Before the tests starts, it creates the init data
+        After all the tests, it clears database and sets the init data
+    Fixture called by default (autouse=True)
+    """
+    _dir = os.path.dirname(os.path.realpath(__file__))
+    file_path = py.path.local(_dir).join('/foglamp_test_storage_init.sql')
+    os.system("psql < {} > /dev/null 2>&1".format(file_path))
+    yield
+    os.system("psql < `locate foglamp_ddl.sql | grep 'FogLAMP/src/sql'` > /dev/null 2>&1")
+    os.system("psql < `locate foglamp_init_data.sql | grep 'FogLAMP/src/sql'` > /dev/null 2>&1")
```
praveen-garg: okay with init, but why ddl is needed?

Keeping the test data in mind; seems its not needed; just DELETE is enough with where condition.

src/python/tests/storage/test_storage_api.py
praveen-garg: @pytest.allure.story("storage client")

src/python/tests/storage/test_storage_api.py
```diff
@@ -0,0 +1,283 @@
...
+
+
+@pytest.allure.feature("api")
+@pytest.allure.story("storage")
+class TestStorageRead:
+    """
+    This class tests SELECT (Read) queries of Storage layer using payload builder
+    """
+    def test_select(self):
+        result = Storage().query_tbl_with_payload("statistics", PayloadBuilder().SELECT().payload())
```
praveen-garg: kindly hold payload in a var; it will help to assert/ debug payload in case of future enhancements, bug fixes! `paylaod = PayloadBuilder().SELECT().payload()`

src/python/tests/storage/test_storage_api.py
```diff
@@ -0,0 +1,283 @@
...
+        result = Storage().query_tbl_with_payload("statistics", PayloadBuilder().
+                                                  WHERE(["bla", "=", "invalid"]).payload())
+        assert "ERROR" in result["message"]
+
+    @pytest.mark.skip(reason="FOGL-607 #7")
+    def test_multiple_and_where(self):
+        result = Storage().query_tbl_with_payload("statistics", PayloadBuilder().
+                                                  WHERE(["value", "!=", 0]).
+                                                  AND_WHERE(["key", "=", "TEST_1"]).
+                                                  AND_WHERE(["previous_value", "<", 10]).payload())
```
praveen-garg: do we have UNIT test for the same payload verification?

src/python/tests/storage/test_storage_api.py
```diff
@@ -0,0 +1,283 @@
...
+        assert result["rows"][0]["value"] == 10
+        assert result["rows"][0]["previous_value"] == 2
+
+    @pytest.mark.skip(reason="FOGL-607 #7")
+    def test_multiple_or_where(self):
+        result = Storage().query_tbl_with_payload("statistics", PayloadBuilder().
+                                                  WHERE(["value", "=", 0]).
+                                                  OR_WHERE(["key", "=", "TEST_1"]).
+                                                  OR_WHERE(["previous_value", ">", 10]).payload())
+        assert len(result["rows"]) == 1
```
praveen-garg: should not the test data be like that AND has 1 result and OR gives 2 results, while having atleast 3 records in table

src/python/tests/storage/foglamp_test_storage_init.sql
praveen-garg: should not we use configuration table or some other which shall allow to have JSON col too, To keep current stuff as is; let's keep statistics too;

src/python/tests/storage/test_storage_api.py
```diff
@@ -0,0 +1,283 @@
...
+    def test_order(self):
+        result = Storage().query_tbl_with_payload("statistics", PayloadBuilder().
+                                                  ORDER_BY(["key", "desc"]).payload())
+        assert len(result["rows"]) == 2
+        assert result["count"] == 2
+        assert result["rows"][0]["key"] == "TEST_2"
+        assert result["rows"][0]["description"] == "Testing the storage service data 2"
+        assert result["rows"][0]["value"] == 15
+        assert result["rows"][0]["previous_value"] == 2
+
```
praveen-garg: test_default_sort
add blank stub with `pass` or `skip`, multi_sort

src/python/tests/storage/test_storage_api.py
```diff
@@ -0,0 +1,283 @@
...
+    def test_aggregate(self):
+        result = Storage().query_tbl_with_payload("statistics", PayloadBuilder().
+                                                  AGGREGATE(["max", "value"]).payload())
+        assert len(result["rows"]) == 1
+        assert result["count"] == 1
+        assert result["rows"][0]["max_value"] == 15
+
```
praveen-garg: add  test stub multi_aggregate

src/python/tests/storage/test_storage_api.py
```diff
@@ -0,0 +1,283 @@
...
+    def test_aggregate_group(self):
+        result = Storage().query_tbl_with_payload("statistics", PayloadBuilder().
+                                                  AGGREGATE(["min", "previous_value"]).
+                                                  GROUP_BY("previous_value").payload())
+        assert len(result["rows"]) == 1
+        assert result["count"] == 1
+        assert result["rows"][0]["min_previous_value"] == 2
+        assert result["rows"][0]["previous_value"] == 2
+
```
praveen-garg: test group_by with having; as having woks with and without

src/python/tests/storage/test_storage_api.py
```diff
@@ -0,0 +1,283 @@
...
+
+@pytest.allure.feature("api")
+@pytest.allure.story("storage")
+class TestStorageInsert:
+    """
+    This class tests INSERT queries of Storage layer using payload builder
+    """
```
praveen-garg: docstring guideline 

```
""" blah text  
"""
```

src/python/tests/storage/test_storage_api.py
```diff
@@ -0,0 +1,283 @@
...
+    def test_valid_insert(self):
```
praveen-garg: valid? :) test_insert itself should mean valid itself

praveen-garg: can not we insert more than 1 record at a time?

src/python/tests/storage/test_storage_api.py
```diff
@@ -0,0 +1,283 @@
...
+
+    @pytest.mark.skip(reason="FOGL-616")
+    def test_invalid_key_update(self):
+        result = Storage().update_tbl("statistics", PayloadBuilder().
+                                      SET(value=23, description="Updated test value 2").
+                                      WHERE(["key", "=", "bla"]).payload())
+        assert "ERROR" in result["message"]
+
+        # Assert that values are not updated (Check any single record)
+        result = Storage().query_tbl_with_payload("statistics", PayloadBuilder().
```
praveen-garg: May be `assert "Updated test value 2" not in rows `?

src/python/tests/storage/test_storage_api.py
```diff
@@ -0,0 +1,283 @@
...
+                                                  WHERE(["key", "=", "TEST_2"]).payload())
+        assert result["rows"][0]["key"] == "TEST_2"
+        assert result["rows"][0]["description"] == "Testing the storage service data 2"
+        assert result["rows"][0]["value"] == 15
+        assert result["rows"][0]["previous_value"] == 2
+
+    @pytest.mark.skip(reason="FOGL-615")
+    def test_invalid_type_update(self):
+        result = Storage().update_tbl("statistics", PayloadBuilder().
```
praveen-garg: can you please add a comment here that col type is int , trying with string

src/python/tests/storage/test_storage_api.py
```diff
@@ -0,0 +1,283 @@
...
+
+    def test_update_without_key(self):
+        result = Storage().update_tbl("statistics", PayloadBuilder().
+                                      SET(value=1, description="Updated test value 4").payload())
+        assert result == {'response': 'updated'}
+        result = Storage().query_tbl_with_payload("statistics", PayloadBuilder().SELECT_ALL().payload())
+        for _i in range(len(result["rows"])):
+            assert result["rows"][_i]["value"] == 1
+            assert result["rows"][_i]["description"] == "Updated test value 4"
+
```
praveen-garg:
```
for r in result["rows"]:
       assert 1 == r["value"]
       assert "Updated test value 4" ==  r["description"]
```

src/python/tests/storage/test_storage_api.py
```diff
@@ -0,0 +1,283 @@
...
+        result = Storage().query_tbl("statistics", PayloadBuilder().WHERE(["key", "=", "TEST_1"]).query_params())
+        assert len(result["rows"]) == 0
+        assert result["count"] == 0
+
+    def test_delete_with_invalid_key(self):
+        result = Storage().delete_from_tbl("statistics", PayloadBuilder().
+                                           WHERE(["key", "=", "TEST_invalid"]).payload())
+
+        # FIXME: Is deleted the correct response?
+        assert result == {'response': 'deleted'}
```
praveen-garg: open a ticket, I think it should also give  rows_affected = 0

src/python/tests/storage/test_storage_api.py
```diff
@@ -0,0 +1,283 @@
...
+
+__author__ = "Vaibhav Singhal"
+__copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
+__license__ = "Apache 2.0"
+__version__ = "${VERSION}"
+
+
+# TODO: remove once FOGL-510 is done
+@pytest.fixture(scope="module", autouse=True)
+def create_init_data(request):
```
praveen-garg: what is request arg?

src/python/tests/storage/test_storage_api.py
```diff
@@ -0,0 +1,283 @@
...
+        assert result["count"] == 2
+        assert result["rows"][0]["key"] == "TEST_1"
+        assert result["rows"][0]["description"] == "Testing the storage service data 1"
+        assert result["rows"][0]["value"] == 10
+        assert result["rows"][0]["previous_value"] == 2
+
+        assert result["rows"][1]["key"] == "TEST_2"
+        assert result["rows"][1]["description"] == "Testing the storage service data 2"
+        assert result["rows"][1]["value"] == 15
+        assert result["rows"][1]["previous_value"] == 2
```
ashish-ScaleDB: Use `expected == actual`

src/python/tests/storage/foglamp_test_storage_init.sql
vaibhav-ScaleDB: Done, added statistics data. This will help to write tests for JSON column data too for which stubs are added

src/python/tests/storage/test_storage_api.py
vaibhav-ScaleDB: DDL is needed because init is not perfect at the moment. It does inserts without deleting the previous values for some tables. We may have duplicate rows if db is not cleaned. This all will go away with test environment setup and cleanup (FOGL-510)

src/python/tests/storage/test_storage_api.py
vaibhav-ScaleDB: done

src/python/tests/storage/test_storage_api.py
vaibhav-ScaleDB: done

izoratti-dianomic:
@vaibhav-ScaleDB - great addition to the testing suite, but it must be in sync with what we have as microservice architecture and plugin.
To be more specific, this test is not for the storage, but for the PostgreSQL plugin.
Please follow the following:
- Coordinate with @MarkRiddoch regarding the tests that make sense with the PG plugin
- Refer to @ashwinscale to review where the storage files are placed and how they can be orginized in the repo. Also, review the Env variables used (psql as is, for example, will not work).
- Review with @ashwinscale how the scripts to control FogLAMP must be shaped to be in sync with the installed versions.

src/python/tests/storage/foglamp_test_storage_init.sql
MarkRiddoch: We have an issue tat we can not currently create test table structures and data inside the storage system using the storage layer and thus we have to bypass the storage layer and go direct. This means we are tied to testing a particular plugin at the moment.

Ultimately we should remove this and have the facilities we need via the storage layer and thus allow the tests to be agnostic of the plugin that is used.

src/python/tests/storage/test_storage_api.py
MarkRiddoch: This is an area we will need to improve over time, it will only really become an issue when we have our second storage plugin and wish to support systems without Postgres. Ideally all DDL will be handled by the plugin and not by use of scripts. We will get rid of these scripts and replace them with a database agnostic mechanism.

ashwinscale:
please pull develop with recent changes FOGL-517

PR #228, opened 2017-10-13
Fogl 612 - Convert Sending process to use Storage Service

sstefano69:


src/python/foglamp/sending_process.py
```diff
@@ -658,12 +653,30 @@ def _last_object_id_update(self, new_last_object_id, stream_id):
...
             _logger.debug("Last position, sent |{0}| ".format(str(new_last_object_id)))
 
-            sql_cmd = "UPDATE foglamp.streams SET last_object={0}, ts=now()  WHERE id={1}" \
-                .format(new_last_object_id, stream_id)
+            # TODO : to be changed using PayloadBuilder after FOGL-616
+            # TODO : verify the support for .SET(rs='now()')
+            # payload = payload_builder.PayloadBuilder() \
+            #     .WHERE(['column', '=', stream_id]) \
+            #     .SET(last_object=new_last_object_id) \
+            #     .SET(rs='now()') \
```
vaibhav-ScaleDB: rs? You mean ts? As per my understanding ts column is db specific, not to be used in code. ts should be updated by db itself on insert/update.

src/python/foglamp/sending_process.py
```diff
@@ -683,8 +696,7 @@ def _send_data_block(self, stream_id):
 
         data_sent = False
 
-        # noinspection PyProtectedMember
-        _logger.debug("{0} - ".format(sys._getframe().f_code.co_name))
+        _logger.debug("{0} - ".format(inspect.currentframe().f_code.co_name))
```
vaibhav-ScaleDB: same as above

src/python/foglamp/sending_process.py
```diff
@@ -718,7 +729,7 @@ def send_data(self, stream_id):
         Todo:
         """
 
-        _logger.debug("{0} - ".format(sys._getframe().f_code.co_name))
+        _logger.debug("{0} - ".format(inspect.currentframe().f_code.co_name))
```
vaibhav-ScaleDB: same as above

src/python/foglamp/sending_process.py
```diff
@@ -736,11 +747,13 @@ def send_data(self, stream_id):
                     _logger.error(_message)
 
                 if not data_sent:
-                    _logger.debug("{0} - sleeping".format(sys._getframe().f_code.co_name))
+                    _logger.debug("{0} - sleeping".format(inspect.currentframe().f_code.co_name))
```
vaibhav-ScaleDB: same as above

src/python/foglamp/sending_process.py
```diff
@@ -736,11 +747,13 @@ def send_data(self, stream_id):
...
                     time.sleep(self._config['sleepInterval'])
 
                 elapsed_seconds = time.time() - start_time
-                _logger.debug("{0} - elapsed_seconds {1}".format(sys._getframe().f_code.co_name, elapsed_seconds))
+                _logger.debug("{0} - elapsed_seconds {1}".format(
+                                                                inspect.currentframe().f_code.co_name,
+                                                                elapsed_seconds))
```
vaibhav-ScaleDB: same as above

src/python/foglamp/translators/omf_translator.py
```diff
@@ -544,17 +526,13 @@ def _flag_created_omf_type(configuration_key, type_id, asset_code):
...
-              "VALUES ('{0}', '{1}', {2})".format(configuration_key,
-                                                  asset_code,
-                                                  type_id)
+    data['configuration_key'] = configuration_key
+    data['asset_code'] = asset_code
+    data['type_id'] = type_id
 
-    _pg_cur.execute(sql_cmd)
-    _pg_conn.commit()
+    _storage.insert_into_tbl("omf_created_objects", json.dumps(data))
```
vaibhav-ScaleDB: Use payload builder and not json.dumps.

src/python/foglamp/sending_process.py
```diff
@@ -160,7 +157,7 @@ class SendingProcess:
         "duration": {
             "description": "How long the sending process should run before stopping.",
             "type": "integer",
-            "default": "3"
+            "default": "60"
```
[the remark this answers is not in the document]
sstefano69: The previous value seemed to small, so I extended.

src/python/foglamp/sending_process.py
```diff
@@ -231,7 +223,7 @@ def _retrieve_configuration(self, stream_id):
         .. todo::
         """
 
-        _logger.debug("{0} - ".format(sys._getframe().f_code.co_name))
+        _logger.debug("{0} - ".format(inspect.currentframe().f_code.co_name))
```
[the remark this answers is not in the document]
sstefano69: I changed from sys._getframe() to inspect.currentframe() as the first one was also working but searching, it seemed that the proper way was to use inspect.currentframe()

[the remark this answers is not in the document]
sstefano69: I prefer to use inspect.currentframe() as it is a standard way/code that I could use without the need to change a string manually.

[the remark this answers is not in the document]
sstefano69: the sending process is going to be in execution for a while, currently 60 seconds, and that it will end and then restarted, do you think that memory consumption could be a problem ?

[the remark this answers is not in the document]
sstefano69: if we have to change, should we do something like : 
frame = inspect.currentframe()
_logger.debug("{0} - ".format(frame.f_code.co_name))
del frame

src/python/foglamp/sending_process.py
```diff
@@ -292,7 +283,7 @@ def start(self, stream_id):
 
         exec_sending_process = False
 
-        _logger.debug("{0} - ".format(sys._getframe().f_code.co_name))
+        _logger.debug("{0} - ".format(inspect.currentframe().f_code.co_name))
```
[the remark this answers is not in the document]
sstefano69: I used that piece of code a lot around.

src/python/foglamp/sending_process.py
```diff
@@ -302,8 +293,11 @@ def start(self, stream_id):
             _logger.info(_MESSAGES_LIST["i000001"])
 
             try:
-                self._pg_conn = psycopg2.connect(self._DB_CONNECTION_STRING)
-                self._pg_cur = self._pg_conn.cursor()
+                status = self._storage.check_service_availibility()
```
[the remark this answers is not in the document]
sstefano69: ok, removed.

src/python/foglamp/sending_process.py
```diff
@@ -513,18 +504,23 @@ def _transform_in_memory_data_statistics(raw_data):
...
+        # Extracts only the asset_code column
+        # and renames the columns to id, asset_code, user_ts, reading
+
         try:
-            for item in raw_data:
+            for row in raw_data:
 
                 # Removes spaces
-                asset_code = item[1].replace(" ", "")
+                asset_code = row['key'].replace(" ", "")
```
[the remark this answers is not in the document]
sstefano69: I used replace to remove spaces in the middle also,  strip removes only leading and trailing characters

src/python/foglamp/sending_process.py
```diff
@@ -574,10 +572,9 @@ def _last_object_id_read(self, stream_id):
         """
 
         try:
-            sql_cmd = "SELECT last_object FROM foglamp.streams WHERE id={0}".format(stream_id)
-
-            self._pg_cur.execute(sql_cmd)
-            rows = self._pg_cur.fetchall()
+            where = 'id={0}'.format(stream_id)
+            streams = self._storage.query_tbl('streams', where)
+            rows = streams['rows']
```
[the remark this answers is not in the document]
sstefano69: I think it should remain as it is : _is_stream_id_valid is executed once at the start of the program - _last_object_id_read is executed within a loop (send_data()/_send_data_block) and the value will change.

src/python/foglamp/sending_process.py
sstefano69: you are right, it is 'ts' - changed. but I don't think the DB is updating that field autonomously, am I wrong ?

src/python/foglamp/sending_process.py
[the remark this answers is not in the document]
sstefano69: I will change all the occurrences  when we will consolidate the direction.

src/python/foglamp/translators/omf_translator.py
sstefano69: ok, done

src/python/foglamp/sending_process.py
sstefano69: verified : 
- the 'ts' field is not updated automatically
- the current code that doesn't use the PayloadBuilder, is :
            values = dict()
            values['last_object'] = new_last_object_id
            values['ts'] = 'now()'

and it works fine.

src/python/foglamp/sending_process.py
```diff
@@ -658,15 +648,35 @@ def _last_object_id_update(self, new_last_object_id, stream_id):
...
-                .format(new_last_object_id, stream_id)
+            # TODO : to be changed using PayloadBuilder after FOGL-616
+            #
+            # timestamp = datetime.datetime.now(tz=timezone.utc)
+            # payload = payload_builder.PayloadBuilder() \
+            #     .WHERE(['column', '=', stream_id]) \
+            #     .SET(last_object=new_last_object_id) \
+            #     .SET(ts=str(timestamp)) \
+            #     .payload()
+            #
```
vaibhav-ScaleDB: 1. Payload builder supports update. You are not using it correctly (No multiple .SET, it should be a single .SET with multiple key value), Btw, as discussed today, you need not update ts from code. It should be done by DB
2. Reference of FOGL-616 is wrong here. FOGL-616 is bug where storage is updating all value irrespective of where.

sstefano69: 1. I'm not using the Payload as for the bug FOGL-616, I retried the code  :
            timestamp = datetime.datetime.now(tz=timezone.utc)
            payload = payload_builder.PayloadBuilder() \
                .SET(last_object=new_last_object_id, ts=str(timestamp)) \
                .WHERE(['id', '=', stream_id]) \
                .payload()

            self._storage.update_tbl("streams", payload)

that seems fine to me (I changed with the single SET), but it updates all the rows,  so I'm using the other way for the moment. 
the reference to the FOGL-616 was for this reason, I will switch to use the commented code instead of current one when FOGL-616 will be solved.

2. about the update of ts, I sent the email, we have the Jira task but we need that for the moment to have the functionality working properly.

PR #232, opened 2017-10-16
python daemon vs allow to start via init script

praveen-garg:
- use `make install`  NO virtualenv ;
- removed python daemon; 
- `foglamp-service.sh` to register as init script; see info in script file itself for howto

**- Make sure to use `python3 -m foglamp.core` when not via /etc/init.d**

**Logs when run as init.d:**

start

```
/etc/init.d$ sudo /etc/init.d/foglamp-service.sh start 
[ ok ] Starting foglamp-service.sh (via systemctl): foglamp-service.service.
```

stop

```
:/etc/init.d$ sudo /etc/init.d/foglamp-service.sh status● foglamp-service.service - LSB: foglamp script
   Loaded: loaded (/etc/init.d/foglamp-service.sh; bad; vendor preset: enabled)
   Active: active (running) since Tue 2017-10-17 02:13:48 IST; 10s ago
     Docs: man:systemd-sysv-generator(8)
  Process: 436 ExecStop=/etc/init.d/foglamp-service.sh stop (code=exited, status=0/SUCCESS)
  Process: 492 ExecStart=/etc/init.d/foglamp-service.sh start (code=exited, status=0/SUCCESS)
   CGroup: /system.slice/foglamp-service.service
           ├─501 /usr/bin/python3 /home/foglamp/.local/bin/foglamp start
           └─535 ./storage --address=localhost --port=39781
 port[535]: Load storage plugin postgres.
 port[535]: Loaded storage plugin postgres.
 port[535]: Starting management api on port 0.
 port[535]: Starting service...
INFO :: instance: foglamp.core.service_registry.instance: Registered service instance id=e5c38b37-52c6-4239-a9f2-149908d749c2: <FogLAMP Storage, type=Storage, protocol=http, address=localhost, service port=36638, management port=35549>
t[535]: Registered service e5c38b37-52c6-4239-a9f2-149908d749c2.
Oct 17 02:14:04 nerd047-ThinkPad [FOGLAMP] 10-17-2017 02:14:03 - INFO :: scheduler: foglamp.core.scheduler: Starting scheduler; mgt port received is 39781
```

status

```
/etc/init.d$ sudo /etc/init.d/foglamp-service.sh stop
[ ok ] Stopping foglamp-service.sh (via systemctl): foglamp-service.service.
```

reference: https://wiki.debian.org/LSBInitScripts


src/python/Makefile
```diff
@@ -92,22 +91,22 @@ doc-build-test: install-tox
 	tox -e docs
 
 install-dep: check-root
-	pip install -r requirements.txt
+	pip3 install -r requirements.txt --user
 
 install: install-dep
-	pip install -e .
+	pip3 install -e . --user
```
praveen-garg: can use `python3 setup.py install --user` 

```Installing foglamp script to /home/foglamp/.local/bin

Installed /home/foglamp/.local/lib/python3.5/site-packages/FogLAMP-0.1-py3.5.egg
Processing dependencies for FogLAMP==0.1
Finished processing dependencies for FogLAMP==0.1
```
may be we also want to define v 0.0.1 per http://semver.org/

src/foglamp-service.sh
```diff
@@ -0,0 +1,83 @@
...
+
+
+# At this point you should be able to start foglamp using the command:
+
+# sudo /etc/init.d/foglamp-service.sh start,
+# check its status with the /etc/init.d/foglamp-service.sh status argument
+# and stop it with sudo /etc/init.d/foglamp-service.sh stop.
+### END INIT INFO
+
+DIR=/home/foglamp/.local/bin
```
MarkRiddoch: Where has this directory come from?

src/python/Makefile
```diff
@@ -16,7 +16,6 @@ CODEDIR := ../src/python/foglamp
 check-root:
 ifeq ($(USER), root)
 	@echo "WARNING: Installing as root should be avoided. Use a virtual environment."
```
MarkRiddoch: We are moving away from virtualenv

src/foglamp-service.sh
praveen-garg: Installing Python packages to a system directory (as `/usr/local/lib/python3.5` and entry point to /usr/local/bin). This requires root access - (sudo) make install

`--user` makes pip / `python3 setup.py install` packages in user's home directory instead, which doesn't require any special privileges.


 https://github.com/foglamp/FogLAMP/pull/232/commits/5d9def6e2344ee5ff43b7189870bd7aa33a0b7d0, shows the difference

src/python/Makefile
praveen-garg: Fixed the text, however kept the check (removed exit already, if root)

src/foglamp-service.sh
MarkRiddoch: Yes, it is well understood that make install needs root access to match the required structure.

izoratti-dianomic:
I would say, let's merge and close, we will review this in the next sprint / review of the init script. 

So for me it is good.

izoratti-dianomic:
Sorry - wrong button!

praveen-garg:
@MarkRiddoch We can avoid foglamp daemon test cycle and focus this way while working on other stuff. if this is the direction to go.

PR #234, opened 2017-10-23
Payload Builder enhancements

amarendra-dianomic:
Enhancements per FOGL-607:
```
6. ORDER_BY and AGGREGATE should return correct payload in case of 
multiple key value pairs passed as dictionaries
7. Multiple AND/OR support
```
Also, support provided for chaining of payload builder json, should there be a need to build the json incrementally.


amarendra-dianomic:
Closed in favour of PR #234 as branch name was changed.

PR #239, opened 2017-10-26
Port file to allow storage client  access across  processes

praveen-garg:
storage client needs service registry access via a) internal api or b) http

#a is not possible for tests and other services as they are different process
#b is the solution, as of today we are accessing it via .port.txt 
Another approach can be,  all other consumer services / plugins have access to core mgt port and make `storage = Storage(core_management_port=x)  object once` to allow storage client to discover the service to connect


MarkRiddoch:
What is the motivation for doing this?

The specification clearly says we determine the port for the storage service by getting the registration record for the storage service from the core.

praveen-garg:
Yes, here we are making core_mgt port available to storage client to connect to storage service.

praveen-garg:
Instead of writing port info to file, services/ task must get core host and port info via arguments. For Tests we either may go with static port for core mgt or will figure out better approach later

PR #242, opened 2017-10-27
Fogl 634 argument parser, scheduler to start processes with address/port of core mgt, device service accepts cmd line args

ashwinscale:
Common argument parser Class foglamp/parser.py
Scheduler starts scheduled tasks with address, port, and name 
Device uses parser.py to parse command line arguments

Note that with this method, arguments can be indirectly via argparse without all arguments known beforehand.


.gitignore
```diff
@@ -14,3 +14,7 @@ docs/.cache/
 # Allure
 allure/
 allure-report/
+
+# storage specific
+build/
+services/storage/*
```
izoratti-dianomic: not sure about line 20, why should it be ignored?

src/python/foglamp/device/__main__.py
```diff
@@ -16,25 +18,19 @@
...
-    plugin = 'CoAP'
+plugin = Parser.get('--name')
+core_mgt_port = Parser.get('--port')
+core_mgt_address = Parser.get('--address')
 
 if plugin is None:
-    # TODO: Output to STDERR
-    print("Required argument 'name' is missing")
-    # TODO: exit 1
+    print("Required argument '--name' is missing")
```
izoratti-dianomic: maybe it does (so please ignore the msg), this should go to syslog, and so the other print(s)

src/python/foglamp/core/scheduler.py
```diff
@@ -627,15 +632,24 @@ def max_running_tasks(self, value: int)->None:
         # the start of the awaited coroutine.
         args = self._process_scripts[schedule.process_name]
 
+        # add core management host and port to process script args
+        args_to_exec = args.copy()
+        args_to_exec.append("--port")
+        args_to_exec.append(str(self._core_management_port))
+        args_to_exec.append("--address")
+        args_to_exec.append("127.0.0.1")
```
izoratti-dianomic: @MarkRiddoch - just want to make sure this goes through socket file... does it?

src/python/foglamp/device/__main__.py
```diff
@@ -16,25 +18,19 @@
...
-        plugin = None
-    else:
-        plugin = plugin
-except IndexError:
-    plugin = None
 
-if plugin is None:
-    # TODO: delete this
-    plugin = 'CoAP'
+plugin = Parser.get('--name')
```
MarkRiddoch: The plugin is not passed in the name parameter the plugin comes form the configuration. The name is used to work out the name of the configuration section to look at https://docs.google.com/document/d/1c1dmMxH36IQNA_gFwxTPttyCeEStsrSxWYUqKeBbfNk/edit#heading=h.o2dnspmkfx4p

src/python/foglamp/core/scheduler.py
MarkRiddoch: Yes, this uses the loopback interface and does not go down to the network hardware - I am guessing this is the question you are asking.

src/python/foglamp/parser.py
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env python3
```
praveen-garg: remove this too

src/python/foglamp/parser.py
```diff
@@ -0,0 +1,23 @@
...
+# FOGLAMP_END
+
+import argparse
+__author__ = "Ashwin Gopalakrishnan"
+__copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
+__license__ = "Apache 2.0"
+__version__ = "${VERSION}"
+
+
+
```
praveen-garg: add docstring and keep 2 lines only :)

src/python/foglamp/parser.py
```diff
@@ -0,0 +1,23 @@
...
+class Parser(object):
+    @staticmethod
+    def get(key):
+      parser = argparse.ArgumentParser()
+      parser.add_argument(key)
+      parser.parse_known_args()
+      return list(vars(parser.parse_known_args()[0]).values())[0]
```
praveen-garg: it may have any known exception?

src/python/foglamp/parser.py
```diff
@@ -0,0 +1,23 @@
...
+__version__ = "${VERSION}"
+
+
+
+class Parser(object):
+    @staticmethod
+    def get(key):
+      parser = argparse.ArgumentParser()
+      parser.add_argument(key)
+      parser.parse_known_args()
```
praveen-garg: is it being used as var in next line? returns a tuple or ?

.gitignore
praveen-garg: I believe, we don't want to push static files to git (built via storage cmake/make)

src/python/foglamp/device/__main__.py
```diff
@@ -16,25 +19,22 @@
...
-    print("Required argument 'name' is missing")
-    # TODO: exit 1
+    _logger.warning("Required argument '--name' is missing")
+elif core_mgt_port is None:
+    _logger.warning("Required argument '--port' is missing")
+elif core_mgt_address is None:
+    _logger.warning("Required argument '--address' is missing")
 else:
-    Server.start(plugin)
+    Server.start(plugin, core_mgt_port, core_mgt_address)
```
praveen-garg: will be fixed in next PR; order of args

PR #246, opened 2017-10-27
FOGL-652 storage client fixes for storage service discovery with core mgt host and port;

praveen-garg:
…  also fixed the tests; provide mgt port at test execution time until we get this fixed
```

tests/storage/test_storage_api.py ................ssss..........

===================== 26 passed, 4 skipped in 7.28 seconds =====================
```

### DIFF
--- a/src/python/foglamp/core/server.py
+++ b/src/python/foglamp/core/server.py
@@ -13,11 +13,13 @@ import asyncio
 from aiohttp import web
 import subprocess
 
+import http
+import json
+
 from foglamp import logger
 from foglamp.core import routes
 from foglamp.core import middleware
 from foglamp.core.service_registry.instance import Service
-from foglamp.storage.storage import Storage
 from foglamp.core.scheduler import Scheduler
 
 __author__ = "Amarendra K. Sinha, Praveen Garg, Terris Linenbach"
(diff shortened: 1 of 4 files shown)


src/python/foglamp/core/server.py
```diff
@@ -184,11 +186,35 @@ def start(cls):
 
         # see http://host:<core_mgt_port>/foglamp/service for registered services
 
-    @staticmethod
-    def stop_storage():
-        # TODO: make client call to service mgt API to ask to shutdown storage?
+    @classmethod
```
ashwinscale: This work is not in the scope of this ticket. Please remove it.

src/python/foglamp/device/__main__.py
```diff
@@ -36,5 +36,5 @@
 elif core_mgt_address is None:
     _logger.warning("Required argument '--address' is missing")
 else:
-    Server.start(plugin, core_mgt_port, core_mgt_address)
```
ashwinscale: Good catch.

src/python/foglamp/storage/storage.py
```diff
@@ -104,35 +104,13 @@ def check_service_availibility(self):
...
+    def _get_storage_service(self, host, port):
         """ get Storage service """
 
-        # TODO: URL to service registry api?
-        conn = http.client.HTTPConnection("{0}:{1}".format("", ""))
+        conn = http.client.HTTPConnection("{0}:{1}".format(host, port))
         # TODO: need to set http / https based on service protocol
 
-        conn.request('GET', url='/foglamp/service')
+        conn.request('GET', url='/foglamp/service?name=FogLAMP%20Storage')
```
ashwinscale: Is this correct?

src/python/foglamp/core/server.py
praveen-garg: Please refer TODO.  Even ping (_check service ability is also just for quick there only) else ping will also be done by microservice mgt API.

src/python/foglamp/storage/storage.py
praveen-garg: Yes, for service discovery by name. for %20 we can do explicity encoding. but I believe its not needed. as name is fixed.

src/python/foglamp/core/server.py
ashwinscale: I see because it has been removed from storage. Personally I'm fine with breaking shutdown on storage as we originally had a temporary fix. We can finish off the client code here. Create a separate ticket for shutdown so that there is separation of concerns on this review. Then we can address shutdown properly when the time comes.

praveen-garg: refer FOGL-653 and a PR (on top of this PR itself, which allows we are not breaking anything and allows to fix FOGL-653 post storage engine migration)

praveen-garg: Also I believe 
1. this is a matter of clean-up and  not to break existing code as its tightly coupled in storage client and we used that to stop storage vs SoC (i.e. Fix is itself in its  concern area)
2. a step towards FOGL-653 hence TODO

PR #248, opened 2017-10-30
Fogl 521 - Port scheduler to use storage service

amarendra-dianomic:


src/python/foglamp/core/scheduler/scheduler.py
```diff
@@ -500,126 +164,113 @@ def __init__(self, core_management_host=None, core_management_port=None):
...
+            # Update the task's status
+            update_payload = PayloadBuilder() \
+                .SET(exit_code=exit_code,
+                     state=int(state),
+                     end_time=str(datetime.datetime.now())) \
+                .WHERE(['id', '=', str(task_process.task_id)]) \
+                .payload()
             try:
-                self._engine.close()
+                with Storage() as conn:
```
praveen-garg: please use _storage = Storage(...) object; create once for Scheduler class

src/python/foglamp/core/scheduler/scheduler.py
```diff
@@ -665,152 +316,45 @@ def max_running_tasks(self, value: int)->None:
...
+                        pid=(self._schedule_executions[schedule.id].
+                             task_processes[task_id].process.pid),
+                        process_name=schedule.process_name,
+                        state=int(Task.State.RUNNING),
+                        start_time=str(datetime.datetime.now())) \
+                .payload()
             try:
-                async with (await self._get_connection_pool()).acquire() as conn:
-                    await conn.execute(insert)
+                with Storage() as conn:
```
praveen-garg: _storage object

src/python/foglamp/core/scheduler/scheduler.py
```diff
@@ -948,208 +581,324 @@ def _schedule_first_task(self, schedule, current_time):
                 "Scheduled task for schedule '%s' to start at %s", schedule.name,
                 datetime.datetime.fromtimestamp(schedule_execution.next_start_time))
 
-    def _schedule_next_task(self, schedule)->None:
-        """Computes the next time to start a task for a schedule.
+    async def _get_process_scripts(self):
+        query_payload = PayloadBuilder().LIMIT(10000).payload()
```
praveen-garg: not sure `.LIMIT(10000)` ?

src/python/foglamp/core/scheduler/scheduler.py
```diff
@@ -948,208 +581,324 @@ def _schedule_first_task(self, schedule, current_time):
...
+        except Exception:
+            self._logger.exception('Query failed: %s', query_payload)
+            raise
 
-        For nonexclusive schedules, this method is called after starting
-        a task automatically (it is not called when a task is started
-        manually).
+    async def _get_schedules(self):
+        # TODO: Get processes first, then add to Schedule
+        query_payload = PayloadBuilder().LIMIT(10000).payload()
```
praveen-garg: not sure .LIMIT(10000) ?

src/python/foglamp/core/scheduler/scheduler.py
```diff
@@ -948,208 +581,324 @@ def _schedule_first_task(self, schedule, current_time):
...
+        try:
+            with Storage() as conn:
+                self._logger.debug('Database command: %s', query_payload)
+                res = conn.query_tbl_with_payload("schedules", query_payload)
+                for row in res['rows']:
```
praveen-garg: what if `res has 'error'`

src/python/foglamp/core/scheduler/scheduler.py
```diff
@@ -948,208 +581,324 @@ def _schedule_first_task(self, schedule, current_time):
...
+    async def _mark_tasks_interrupted(self):
+        """The state for any task with a NULL end_time is set to interrupted"""
+        # Update the task's status
+        update_payload = PayloadBuilder() \
+            .SET(state=int(Task.State.INTERRUPTED),
+                 end_time=str(datetime.datetime.now())) \
+            .WHERE(['end_time', '=', "NULL"]) \
+            .payload()
+        try:
+            with Storage() as conn:
```
praveen-garg: _storage object

src/python/foglamp/core/scheduler/scheduler.py
```diff
@@ -948,208 +581,324 @@ def _schedule_first_task(self, schedule, current_time):
...
-                now < schedule_execution.next_start_time):
-            # The task was started manually
-            # Or the schedule was modified after the task started (AVOID_ALTER_NEXT_START)
+        self._logger.info("Starting")
+
+        self._start_time = self.current_time if self.current_time else time.time()
+
+        # make sure that it go forward only when storage service is ready
+        storage_service = None
+        attempts_left = 3
```
praveen-garg: can we make it 
``` 
max_wait = 30
attempts_left = max_wait / 5
asyncio.sleep(5)

``` 
```
signal.setitimer(signal.ITIMER_REAL, max_wait)

```
to save 5 seconds if received in any of the attempt; cost will be minimal as its in-memory object check

src/python/foglamp/core/scheduler/scheduler.py
```diff
@@ -1169,9 +918,9 @@ def _resume_check_schedules(self):
             raise ValueError("name can not be empty")
 
         if schedule.repeat is not None and not isinstance(schedule.repeat, datetime.timedelta):
-            raise ValueError('repeat must be of type datetime.time')
+            raise ValueError('repeat must be of type datetime.timedelta')
 
-        if schedule.exclusive is None:
+        if schedule.exclusive is None or not (schedule.exclusive == True or schedule.exclusive == False):
```
praveen-garg: could you please elaborate why? you mean it can be non-boolean as well? then add type check ?

src/python/foglamp/core/scheduler/scheduler.py
```diff
@@ -1202,62 +951,59 @@ def _resume_check_schedules(self):
...
+                     exclusive='t' if schedule.exclusive else 'f',
+                     process_name=schedule.process_name) \
+                .WHERE(['id', '=', str(schedule.schedule_id)]) \
+                .payload()
             try:
-                async with (await self._get_connection_pool()).acquire() as conn:
-                    result = await conn.execute(update)
-
-                    if result.rowcount == 0:
+                with Storage() as conn:
```
praveen-garg: _storage object

src/python/foglamp/core/scheduler/scheduler.py
```diff
@@ -1202,62 +951,59 @@ def _resume_check_schedules(self):
...
+                    self._logger.debug('Database command: %s', update_payload)
+                    res = conn.update_tbl("schedules", update_payload)
+                    if res.get('count') == 0:
```
praveen-garg: what if `res['error']`

src/python/foglamp/core/scheduler/scheduler.py
```diff
@@ -1202,62 +951,59 @@ def _resume_check_schedules(self):
...
+                        schedule_interval=str(schedule.repeat),
+                        schedule_day=day if day else 0,
+                        schedule_time=str(schedule_time) if schedule_time else '00:00:00',
+                        exclusive='t' if schedule.exclusive else 'f',
+                        process_name=schedule.process_name) \
+                .payload()
             try:
-                async with (await self._get_connection_pool()).acquire() as conn:
-                    await conn.execute(insert)
+                with Storage() as conn:
```
praveen-garg: _storage object

src/python/foglamp/core/scheduler/scheduler.py
```diff
@@ -1268,120 +1014,71 @@ def _resume_check_schedules(self):
...
         except KeyError:
             raise ScheduleNotFoundError(schedule_id)
 
-        return self._schedule_row_to_schedule(schedule_id, schedule_row)
+        # TODO: Inspect race conditions with _set_first
+        delete_payload = PayloadBuilder() \
+            .WHERE(['id', '=', str(schedule_id)]) \
+            .payload()
+        try:
+            with Storage() as conn:
```
praveen-garg: _storage object

src/python/foglamp/core/scheduler/scheduler.py
```diff
@@ -1454,61 +1141,48 @@ def _schedule_row_to_schedule(cls,
...
-
-        if offset:
-            query = query.offset(offset)
-
-        if limit:
-            query = query.limit(limit)
+            chain_payload = PayloadBuilder(chain_payload).ORDER_BY(sort).chain_payload()
 
+        query_payload = PayloadBuilder(chain_payload).payload()
+        print(query_payload)
```
praveen-garg: remove print

src/python/foglamp/core/scheduler/scheduler.py
```diff
@@ -1454,61 +1141,48 @@ def _schedule_row_to_schedule(cls,
...
         tasks = []
 
-        self._logger.debug('Database command: %s', query)
-
         try:
-            async with (await self._get_connection_pool()).acquire() as conn:
-                async for row in conn.execute(query):
+            with Storage() as conn:
```
praveen-garg: _storage object

src/python/foglamp/core/scheduler/scheduler.py
```diff
@@ -1454,61 +1141,48 @@ def _schedule_row_to_schedule(cls,
...
+                self._logger.debug('Database command: %s', query_payload)
+                res = conn.query_tbl_with_payload("tasks", query_payload)
+                for row in res['rows']:
```
praveen-garg: what if res['error]

src/python/foglamp/core/scheduler/scheduler.py
```diff
@@ -1565,100 +1239,24 @@ def _check_purge_tasks(self):
...
-        self._logger.debug('Database command: %s', delete)
-
+        delete_payload = PayloadBuilder() \
+            .WHERE(["state", "!=", int(Task.State.RUNNING)]) \
+            .AND_WHERE(["start_time", "<", str(datetime.datetime.now() - self._max_completed_task_age)]) \
+            .LIMIT(self._DELETE_TASKS_LIMIT) \
+            .payload()
         try:
-            async with (await self._get_connection_pool()).acquire() as conn:
+            with Storage() as conn:
```
praveen-garg: _storage object

src/python/foglamp/core/server.py
```diff
@@ -6,19 +6,19 @@
...
-from foglamp.core import routes
 from foglamp.core import middleware
+from foglamp.core import routes
+from foglamp.core.scheduler.scheduler import Scheduler
 from foglamp.core.service_registry.instance import Service
 from foglamp.storage.storage import Storage
-from foglamp.core.scheduler import Scheduler
 
 __author__ = "Amarendra K. Sinha, Praveen Garg, Terris Linenbach"
 __copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
```
praveen-garg: seems we can avoid this file changes

src/python/foglamp/storage/payload_builder.py
```diff
@@ -18,7 +18,7 @@
 import numbers
 
 from foglamp import logger
-
+from foglamp.storage.utils import Utils
```
praveen-garg: its okay, though can be via FOGL-607's sibiling (other payload builder tickets)

src/python/tests/core/test_scheduler.py
```diff
@@ -9,25 +9,119 @@
...
+        cmd_with_args = ['./storage', '--address={}'.format(host),
+                         '--port={}'.format(m_port)]
+        import subprocess
+        subprocess.call(cmd_with_args, cwd=_STORAGE_DIR)
+    except Exception as ex:
+        pass
+
+def stop_storage():
+    try:
+        Storage().shutdown()
```
praveen-garg: make a http client call. if you want to keep this test infra. Need to review this test structure

src/python/foglamp/core/scheduler/scheduler.py
MarkRiddoch: The original plan I had was to make things that depended on the storage service being there to be triggered off the registration coming form the storage service rather than have some form of polling or busy wait going on. Hence the words int he spec about callbacks that are triggered by the registration. However we didn't implement that and now os no the time to do that. It is highly questionable about having limits on the number of sleeps we do waiting for the storage service, if it doesn't come up then there is little we can do to continue. Rather than change this at this stage I think we should continue using the mechanism you have implemented for now and allow us to complete this sprint.

src/python/tests/core/test_scheduler.py
```diff
@@ -9,25 +9,119 @@
...
+    loop.run_until_complete(_server.wait_closed())
+    loop.run_until_complete(_app.shutdown())
+
 
 @pytest.allure.feature("unit")
 @pytest.allure.story("scheduler")
 class TestScheduler:
+    _engine = None  # type: aiopg.sa.Engine
+
+    async def _get_connection_pool(self) -> aiopg.sa.Engine:
```
MarkRiddoch: This test will not work if our storage engine is not Postgres. OK for today but long term we need to approach this differently.

We could simply use the storage layer to insert the test data. Maybe add this as a TODO in the test.

src/python/tests/core/test_scheduler.py
```diff
@@ -9,25 +9,119 @@
 import os
 import time
 import uuid
-
+import aiopg
```
MarkRiddoch: See comment below, we should aim to remove this in time.

src/python/tests/core/test_scheduler.py
```diff
@@ -9,25 +9,119 @@
...
+        return self._engine
+
+    async def populate_test_data(self):
+        """Delete all schedule-related tables and insert processes for testing"""
+        async with (await self._get_connection_pool()).acquire() as conn:
+            await conn.execute('delete from foglamp.tasks')
+            await conn.execute('delete from foglamp.schedules')
+            await conn.execute('delete from foglamp.scheduled_processes')
+            await conn.execute(
+                '''insert into foglamp.scheduled_processes(name, script)
```
MarkRiddoch: We can't call sleep anymore, as noted in the scrum call. I suggest we wrap the sleep in a shell script, that way we can also test the scheduler is passing the three required arguments --name=, --address= and --port=

src/python/tests/core/test_scheduler_get_tasks.py
```diff
@@ -25,29 +25,31 @@
...
 import os
 import random
-import sqlalchemy
-import sqlalchemy.dialects.postgresql
+import signal
+import sys
 import time
 import uuid
-
+import aiopg
```
MarkRiddoch: We should aim to remove this from tests so that the tests will work with any database and use the storage service

src/python/tests/core/test_scheduler_get_tasks.py
```diff
@@ -25,29 +25,31 @@
...
+from foglamp.storage.storage import Storage
+
+__author__ = "Terris Linenbach, Amarendra K Sinha"
 __copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
 __license__ = "Apache 2.0"
 __version__ = "${VERSION}"
 
 _CONNECTION_STRING = "dbname='foglamp'"
-
 _TASKS_TABLE = sqlalchemy.Table('tasks', sqlalchemy.MetaData(),
```
MarkRiddoch: As with the other comments on the test I think we must move away form this. For now lets mark this with a TODO and tackle this when we address testing in a future sprint.

PR #249, opened 2017-10-30
Fogl 635 - Sending process: add the handling of 3 new command line parameters

sstefano69:
Requires FOGL-652 PR 246


src/python/foglamp/translators/omf_translator.py
```diff
@@ -1075,6 +1075,6 @@ def _transform_in_memory_row(data_to_send, row, target_stream_id):
 if __name__ == "__main__":
 
     # Used to assign the proper objects type without actually executing them
-    _storage = Storage()
+    _storage = Storage("127.0.0.1", "0")
```
MarkRiddoch: What is the used for? Clearly this will not work if the storage client is used.

ashwinscale:
Requires FOGL-652 PR 246

src/python/foglamp/translators/omf_translator.py
sstefano69: that piece of code is not going to be executed as it is under the if -  if __name__ == "__main__"

it is used to properly define the type of the _storage object without executing/instantiating  the Storage class.

the sending process will instantiate the Storage class (passing received address and port ) and it will pass the reference to the omf_translator.py setting omf_translator's _storage object.

PR #250, opened 2017-10-30
Fogl 650 startup and configuration of device server fix

ashwinscale:
1. change foglamp_init_data.sql - COAP configuration, task name for device is COAP also
2. change device server code to retrieve plugin to load (from configuration) based on command line argument passed in

### DIFF
--- a/src/python/foglamp/device/server.py
+++ b/src/python/foglamp/device/server.py
@@ -79,10 +79,11 @@ class Server:
             try:
                 plugin_module = config['plugin']['value']
             except KeyError:
-                sys.exit(1)
+                _LOGGER.warning("Unable to obtain configuration of module for plugin {}".format(plugin))
+                raise
 
             try:
-                cls._plugin = __import__(plugin_module, fromlist=[''])
+                cls._plugin = __import__("foglamp.device.{}_device".format(plugin_module), fromlist=[''])
             except Exception:
                 error = 'Unable to load module {} for device plugin {}'.format(plugin_module,
                                                                                plugin)
--- a/src/python/foglamp/device/server.py
+++ b/src/python/foglamp/device/server.py
@@ -69,20 +69,17 @@ class Server:
         cls._core_management_port = core_mgt_port
 
         try:
-            # TODO: Category name column is allows only 10 characters.
-            # This needs to be increased
             category = plugin
-
             config = {}
-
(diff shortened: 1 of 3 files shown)


src/python/foglamp/device/server.py
```diff
@@ -69,20 +69,17 @@ class Server:
...
 
+            config = await configuration_manager.get_category_all_items(category)
+
             try:
-                plugin_module = config['plugin']
+                plugin_module = config['plugin']['value']
             except KeyError:
-                # TODO: Delete me
-                plugin_module = 'foglamp.device.coap_device'
+                sys.exit(1)
```
ashwinscale: Should we throw an exception instead?

src/sql/foglamp_init_data.sql
```diff
@@ -73,7 +73,10 @@ INSERT INTO foglamp.configuration ( key, description, value )
 INSERT INTO foglamp.configuration ( key, description, value )
      VALUES ( 'SYPURGE', 'System Purge', to_jsonb( '{ "retention" : 259200, "last purge" : "' || now() || '" }' ) );
 
-
+-- COAP:  CoAP device server
+--        plugin: python module to load dynamically
+INSERT INTO foglamp.configuration ( key, description, value )
+     VALUES ( 'COAP', 'CoAP Plugin Configuration', ' { "plugin" : { "type" : "string", "value" : "foglamp.device.coap_device", "default" : "foglamp.device.coap_device", "description" : "Python module name of the plugin to load" } } ');
```
ashwinscale: Note than this is the fully specified module to import.

src/python/foglamp/device/server.py
MarkRiddoch: Let's at the very least log something. The worst thing is a service that mysteriously exists when this is at a customer site - the support guy will hate you for this :-)

src/sql/foglamp_init_data.sql
MarkRiddoch: I don't get why we have the fog lamp.device.coap_device in the config. T~his is meant to be dealt with programmatically when we do the import. This is a python and code organisation issue, not something to. burden the end user with, they just want to know they are using the coap plugin, where the python code lives is our problem not theirs.

ashwinscale: fixed

PR #219, opened 2017-10-06
FOGL-519 : append readings using Storage layer

praveen-garg:
- [x] remove storage registration as it will be self registration by storage layer **Blocked by** https://github.com/foglamp/FogLAMP/pull/223

PR #253, opened 2017-10-31
FOGL-660 Error handing for Device service plugin ingest (FOGL-519)

praveen-garg:
* attempt only once, if unknown error/ exception
* if storage engine says, retry-able then try again 
* if no, just get put of loop


praveen-garg:
https://github.com/foglamp/FogLAMP/pull/219

PR #254, opened 2017-10-31
[FOGL-638] DISTINCT payload builder support added with test

praveen-garg:


src/python/tests/storage/data/payload_distinct.json
```diff
@@ -0,0 +1,9 @@
+{
+  "where" : {
+    "column" : "id",
+    "condition" : "<",
+    "value" : 1000
+  },
+  "modifier" : "distinct",
+  "return" : [ "description"]
+}
```
MarkRiddoch: Odd looking character in file, is this a problem?

praveen-garg: its not a problem. this just show that there is no new line at EOF (end of file).  Though fixed in new commit and also remove space before ` "description"]` in return key value

PR #255, opened 2017-10-31
FOGL-599_Send_Proc_log - implement the usage of the log table through the Storage Layer

sstefano69:
Requires FOGL-652 PR 246


src/python/foglamp/sending_process.py
```diff
@@ -189,7 +241,17 @@ class SendingProcess:
 
     }
 
-    def __init__(self):
+    def __init__(self, _mgt_name, _mgt_port, _mgt_address):
+        """
+
+        Args:
+            _mgt_name: Unique name that represents the microservice - Used by the Storage layer
```
MarkRiddoch: Strictly speaking this is not used by the storage layer, this is the name of your micro service or task

PR #256, opened 2017-10-31
FOGL-626 [Port the purge process to use the storage service]

vaibhav-ScaleDB:
Purge process migrated,
Tests done with hard coded core_management_port, to run, change and execute:
9 passed in 2.41 seconds

Previous unit test need to be removed after review

purge_test_scenarios


src/python/tests/data_purge/test_data_purge.py
```diff
@@ -0,0 +1,391 @@
...
+from foglamp.data_purge.purge import start, _CONFIG_CATEGORY_NAME
+from foglamp.storage.payload_builder import PayloadBuilder
+from foglamp.storage.storage import Storage, Readings
+
+__author__ = "Ori Shadmon"
+__copyright__ = "Copyright (c) 2017 OSI Soft, LLC"
+__license__ = "Apache 2.0"
+__version__ = "${VERSION}"
+
+_core_management_port = 36979
```
MarkRiddoch: This will not always work as the port will change. A better approach for tests would be to tun the storage service with a fixed port, which can be achieved via configuration options, and use that.

Please add a TODO that this needs to be addressed.

praveen-garg:
1) remove use of `global`; if it can not be avoided (rare) add comment (reason)
2) tests/data_purge/test_purge.py shold be deleted in favor of tests/data_purge/test_data_purge.py ?
3) can we simplify that requirements paragraph explanation and keep it simple to stat what it does (bullet points are good to have)?
4) move this to write_statistics()?
  ```
    def write_statistics(total_purged, unsent_purged)
        loop = asyncio.get_event_loop()
        loop.run_until_complete(statistics.update_statistics_value('PURGED', total_purged))
        loop.run_until_complete(statistics.update_statistics_value('UNSNPURGED', unsent_purged))
```

vaibhav-ScaleDB:
@MarkRiddoch Added the TODO comment 
@praveen-garg 
1. Removed globals, Added class and variables are private to the class
2. test_purge deleted
3. Simplified comments (As per the design doc)
4. Done

@oshadmon Yes test_purge will not run. The point of this pull is to migrate to storage layer. Earlier tests were using sqlalchemy. Also I realised these were some scenarios that were not required to test like test_config_change, also there were helper methods that were not required and scenarios that were missing. Looking all those changes, I created a separate file (test_data_purge) that runs the scenarios mentioned in the screenshot. Please see these scenarios and let me know if any scenario you feel is missing. I have deleted test_purge.py

Also, To run the tests, you need to edit the value of core_management_port in test_data_purge which you can retrieve from syslogs. (I assume you are aware of the new micro service architecture and how to plug the storage service).  core_management_port is a hardcoded value in test that will be removed in future when we have a test architecture ready.

oshadmon:
LGTM

praveen-garg:
FOGL-510 will fix it.

PR #261, opened 2017-11-02
FOGL-435 device registration 

ashwinscale:
I've put the code in separate methods for better organization.

Currently, routes points to service_registry.py. There is a longer discussion here (and more tasks I believe). I say this because microservicement management routes could point to a generic class (microservice_management.py). 

Service_registry.py was created when it was thought that microservice management is ONLY for core. There should be a reusable way to implement microservice management routes so that ping, shutdown, etc. can be leveraged by all microservices (including core). 

Unregister service is currently not implemented either. 

I've kept this PR to a minimum so to previous comments:

> shutdown
leaving Terris's old method behind which cancels all async tasks (including server). No implementation of stopping via microservice management api yet.

> routes
see above. I believe more general microservice management planning and work is needed here. 

> aiohttp client
not needed as it is run once for registration only (executed before all other tasks are run via async)


src/python/foglamp/device/server.py
```diff
@@ -126,5 +131,40 @@ def start(cls, plugin, core_mgt_host, core_mgt_port):
                 signal_name,
                 lambda: asyncio.ensure_future(cls._stop(loop)))
 
+        microservice_management_app = web.Application(middlewares=[middleware.error_middleware])
+        routes.setup(microservice_management_app)
```
praveen-garg: and routes point to ? service registry?

praveen-garg: microservice_management_app should be device_management_app

src/python/foglamp/device/server.py
```diff
@@ -126,5 +131,40 @@ def start(cls, plugin, core_mgt_host, core_mgt_port):
...
+        service_registration_payload = {
+                "name"            : plugin,
+                "type"            : "Southbound",
+                "management_port" : int(microservice_management_port),
+                "service_port"    : 0,
+                "address"         : "127.0.0.1",
+                "protocol"        : "http"
+            }
+
+        conn.request(method='POST', url='/foglamp/service', body=json.dumps(service_registration_payload))
```
praveen-garg: move to a new `def register`? may be  we also can use `async with aiohttp.ClientSession() as session: ` aiohttp client

src/python/foglamp/device/server.py
praveen-garg: put into a try catch block? 

Move start and shutdown in two def? and start_app return a tuple?
Sort of:

```
serv_generater, handler, app = loop.run_until_complete(start_app(loop))
serv = loop.run_until_complete(serv_generator)
    log.debug('start server', serv.sockets[0].getsockname())
   
try:
   # register here; run that co-routine before loop starts
    loop.run_forever()
except KeyboardInterrupt:
    log.debug(' Stop server begin')
finally:
    loop.run_until_complete(shutdown(serv, app, handler))
    loop.close()
log.debug('Stop server end')
```

AND 

```
async def shutdown(server, app, handler):
    server.close()
    await server.wait_closed()

    app.client.close()  # database connection close
    await app.shutdown()
    await handler.finish_connections(10.0)
    await app.cleanup()
```

src/python/foglamp/microservice_management/service_registry/instance.py
```diff
@@ -23,7 +23,7 @@ class Type(IntEnum):
 
         Storage = 1
         Core = 2
-        Device = 3
+        Southbound = 3
```
praveen-garg: looks fine; may be it should be *SouthBound*; but we can make search by type cAsE insensitive (sooner/ later point of time)!

src/python/foglamp/device/server.py
MarkRiddoch: I would say microservice_management_app is a better name as it is Maning the micro service elements and not managing the device elements. So I would say leave this as it is personally.

ashwinscale: > and routes point to ? service registry?
Yes. Currently the ping is also in service_registry, along with shutdown (unimplemented), etc.

src/python/foglamp/microservice_management/service_registry/instance.py
ashwinscale: leave as Southbound see https://docs.google.com/a/dianomic.com/document/d/1MTFrwyXIa7YEPAIpnlNmDG7X1qFie9SL8PW2YT-hOkU/edit?disco=AAAABamR_ys

praveen-garg: Yeah, I saw that. Hence the comments around CamelCase thing and case insensitive search

src/python/foglamp/device/server.py
```diff
@@ -61,61 +82,109 @@ class Server:
...
             _LOGGER.exception(error)
             print(error)
             asyncio.ensure_future(cls._stop(loop))
 
     @classmethod
+    def _make_microservice_management_app(cls):
+        # create web server application
+        cls._microservice_management_app = web.Application(middlewares=[middleware.error_middleware])
+        # register supported urls
+        routes.setup(cls._microservice_management_app)
```
praveen-garg: I'm not sure, why device service routes should point to service registry methods?

`ping` will be device service ping; not core service ping. @MarkRiddoch thoughts?

praveen-garg:
Okay, as @MarkRiddoch  said, please add TODO  (may be with stub with `pass`) for shutdown and unregister; but make sure routes point to these and have device service ping with uptime info as well.

src/python/foglamp/device/server.py
MarkRiddoch: Yes, the ping is to ping the management API and each micro service will have one.

praveen-garg: I doubt! `http://localhost:<device microservice mgt port>/service/ping` should give info of device microservice or core micro service managment api
Personally  I believe its worth to wait to fix this routes binding

PR #263, opened 2017-11-02
fix readings sensor-read json.dumps issue :[

praveen-garg:


praveen-garg:
thanks @sstefano69 @MarkRiddoch

sstefano69:
tried in my local dev env and it worked fine.

PR #272, opened 2017-11-06
FOGL-674 - Port Pending core/api/scheduler.py to use storage service

amarendra-dianomic:


src/python/foglamp/core/api/scheduler.py
```diff
@@ -566,8 +574,26 @@ def _extract_args(data, curr_value):
 
     try:
         name = request.query.get('name') if 'name' in request.query else None
-
-        tasks = await scheduler_db_services.read_tasks_latest(name)
+        if name:
+            payload = PayloadBuilder() \
+                    .SELECT(("id", "process_name", "state", "start_time", "end_time", "reason", "pid", "exit_code")) \
+                    .WHERE(["process_name", "=", name])\
+                    .ORDER_BY(["process_name", "asc"], ["start_time", "desc"]) \
```
MarkRiddoch: No need to sort by process name as you are selecting single process name

PR #274, opened 2017-11-07
Fogl 647 ashwin DO NOT MERGE

ashwinscale:
Note that I encountered a crash when trying to port over the device service. 
This is why I didn't complete this work. The issue is https://scaledb.atlassian.net/browse/FOGL-690

It is a new issue. Previously this particular configuration was created via the pg connector, and so the issue was not observed.


ashwinscale:
Ignore the backup restore stuff, it isn't in FOGL-518 yet but it is in develop

src/python/foglamp/core/scheduler/scheduler.py
```diff
@@ -662,7 +663,7 @@ def _schedule_first_task(self, schedule, current_time):
             },
             "max_completed_task_age_days": {
                 "description": "The maximum age, in days (based on the start time), for a rows "
-                               "in the tasks table that do not have a status of 'running'",
```
ashwinscale: @MarkRiddoch Note that having a single quoted string 'running' also caused a crash.

MarkRiddoch: Let's remove the quotes today and raise a Jira to get this fixed in the next sprint

PR #285, opened 2017-11-10
FOGL-679 and FOG-680 to port device and coap to new directory structure

ashwinscale:
Dependent on FOGL-678 PR 283


python/foglamp/services/south/server.py
```diff
@@ -102,7 +102,7 @@ class Server:
                 raise
 
             try:
-                cls._plugin = __import__("foglamp.device.{}_device".format(plugin_module_name), fromlist=[''])
+                cls._plugin = __import__("foglamp.plugins.south.{}.{}".format(plugin_module_name,plugin_module_name), fromlist=[''])
```
ashwinscale: @MarkRiddoch there isn't any new documentation on how this should be so I've kept it this way. Please advice.

PR #287, opened 2017-11-10
FOGL-688 Implementation for North HTTP translator plugin

ashish-ScaleDB:
(Initial stage) of coding and I followed OMF Translator as base

1) File http_translator.py added to `python/plugins/north/http` as per ticket
2) JIRA Questions posted on [FOGL-688](https://scaledb.atlassian.net/browse/FOGL-688)


ashish-ScaleDB:
Ready for initial review.

sstefano69:
I think the path and file name should be : 
- python\foglamp\plugins\north\http\http.py

ashish-ScaleDB:
@MarkRiddoch @sstefano69 

Second iteration:

**Fixed Items:**

1) Now plugins directory structure at right place  as per develop branch

2) file name with http.py -> Giving me ImportError: No module named 'http.client'; 'http' is not a package. So I have not changed the file naming http_translator.py

3) **Queries**:

  a) As per specs **plugin_retrieve_info** def should be rename as **plugin_info**. So we need to refactor sending process and OMF translator code accordingly

  b) **plugin_reconfigure** - simple PASS (North plugin cannot be reconfigured?)

  c) **initialize** - Which code is responsible to write/merge configuration (sending task or this plugin)

  d) cancel tasks - wait for some fixed time or this will be a configurable item?

  e) _send_payloads - In case we want to send list of blocks for diff asset_codes whereas send_payload - is for sending a single block payload

  f) _send - headers property (configurable?)
	
  g) For DEFAULT_CONFIG - we can split uri and port values, but right now it is same as OMF

  h) plugin info should be written to init.sql?

4) Pending items: 
    a) remove print statements
    b) remove main def and add tests
    c) second iteration feedback changes as per clarification on listed queries

Kindly advise.

sstefano69:
-1) looking at the task description, I understood that the dir should fit the new repository structure, so :

`python\foglamp\plugins\north\http\http.py
`
not : 

`src/python/foglamp/plugins/north/http/http_translator.py
`

-2) I think the error '...No module named..' is related at the presence or not of the file __init__.py
in the directory tree, not at the file name itself 
and btw the sending process will not load a plugin like 'http\http_translator.py' 
it requires same dir/file like 'http\http.py'

at hits point, were you able to :

2.1) ingest some data in the readings table using fogbench
2.2) having the sending process sending that data using the http plugin 
2.3) checking the data in the destination of the http plugin
2.4) checking the update of the streams table.


a) in the past we agreed to use plugin_retrieve_info (to add an action/verb) so please use this one

b) plugin_reconfigure - it is not currently implemented in the OMF.

c) the OMF is not merging configurations, please check with Mark about the http.

d) cancel tasks - please check with Mark.

e) _send_payloads - please check with Mark but I would say _send_payloads as for example the fogbench will 
    for sure generate payload for multiple asset_codes.

f) please check with Mark.

g) please check with Mark.

h)  please check with Mark, but I think we could create the job without schedule it like for example :

```
insert into foglamp.scheduled_processes (name, script) values ('http translator','["python3", "-m", "foglamp.tasks.north.sending_process", "--stream_id", "3", "--debug_level", "1"]');

INSERT INTO foglamp.streams (id,destination_id,description, last_object,ts ) VALUES (3,1,'HTTP translator', 0,now());
```

sstefano69:
comments about the code :

- plugin_retrieve_info - it should return in 'config' not the default/hard coded cfg but the one retrieved from the configuration manager
        
- plugin_init(loop, storage_client)- the sending process is not currently passing parameters

    - the loop should be instantiated locally by the plugin
    - the reference to the storage is passed as a global using the variable _storage, 'sending process' code : 
        self._plugin._storage = self._storage
        
- plugin_send - the current implementation expected by the 'sending process' is 'def plugin_send(raw_data, stream_id)' having :
    
    - raw_data  : Data to send as retrieved from the storage layer
    - stream_id : identification of the stream and referencing the foglamp.streams table

note : I do not think you were able to send data using fogbench + 'sending process' having the current http plugin

ashish-ScaleDB:
> for file naming convention

Reason: Why we cannot named file as http.py. Because it is overriding the original module http
[More info](https://stackoverflow.com/questions/28646271/when-i-am-importing-http-server-from-the-idle-it-works-but-when-i-run-a-pytho)

> About defs which we are calling via sending process
plugin_retrieve_info
Yupe, I am aware of that and it was deliberately not added, because I had some confusions when writing plugin code thats why I asked "Which code is responsible to write/merge configuration, storage reference, event loop etc... (sending task or this plugin)?" 

> plugin_init
if sending process is writing / reading  to config; will remove that storage pass
also will remove loop arg too and decomment line loop = asyncio..

> still not clear about the plugin_send -- stream_id parameter?

> What about plugin_shutdown?

> Also I don't think so with that directory structure and with above changes requested, we won't be able to test with sending process (See _TRANSLATOR_PATH)

NOTE: Shouldn't we follow same approach as HTTP_South and COAP plugin does? (i.e naming convention for defs and its expecting args, plugin info in init.sql etc.)

ashish-ScaleDB:
@MarkRiddoch @sstefano69 

Fixed as per Stefano's latest feedback:

1) I reverted `src/python/foglamp/plugins/north/http` directory structure and placed inside `translators` directory to test with existing sendingprocess

2) As I can see there are deviations from the [specs ](https://docs.google.com/document/d/1J6c7J4bPjqamqVwoYV93SOZJtTJr-GXl-zSgUDA-T8s/edit#heading=h.ed8mk4kyqr91)

a) Instead of plugin_retrieve_info we should use plugin_info 

b) To implement the specs, we will need to create initial configuration in init.sql which I have 
       implemented for http_translator. We also need to do so for OMF translator as well.

c) Writing/reading configuration and using storage should not be responsibility of the Plugin itself but 
      should be done by the parent process. This calls for changes in sending_process.py and as such 
       also in omf_translator.py

 These are some of the main deviations I can see as of now and which need to be addressed at the 
earliest.

we have experienced that curently translators value has to be hardcoded in sendingprocess _CONFIG_DEFAULT. How to choose it dynamically? Also when translator set to any valid value, stats are being updated for all translators (SENT_1, SENT_2, SENT_3). seems that we need to tweak sending_process.py appropriately.

Please also note that, I have successfully tested http_translator plugin using fogbench to send readings, to a simulated http server.

Please advise.

ashish-ScaleDB:
Jira created https://scaledb.atlassian.net/browse/FOGL-730

PR #288, opened 2017-11-10
FOGL-687 - Implement HTTP southern  plugin

amarendra-dianomic:
@MarkRiddoch @ashwinscale @praveen-garg, three major changes have been made in this PR along with the new http_south plugin:

1) I have removed Coap related Schedules init data from forlamp_init_data.sql. With a little change in device/server.py/_start(), lines 98-106, this is no longer required. And it is more in tune with the Architecture doc.

2) In device plugins (coap_device.py and http_south.py), except for plugin_info() method, all other plugin...() methods  have been made async because the reason as under:

	For http_south plugin, we need to implement an aiohttp server BUT we cannot use the default server provided by aiohttp as it goes into its own loop.run_forever() WHEREAS we need to execute code after initialisation of this server via plugin i.e. line#125 of device/server.py. Hence, we are forced to use the already running event loop. Since the app, handler and server created in http_south plugin are coro, they need to be awaited and as such method containing these lines must be an async method. Since device/server.py is common for both http_south and coap, it must be modified to await an async method and as such the corresponding methods in coap_device.py also changed to be async methods (although it has nil impact on running of the coap server).

	I already tried passing the current running event_loop into plugin_run and then apply a run_until_complete() but it did not work. 

	It may be noted that this does not impact plugins in other areas e.g. omf_translator, sending_process etc as they work independent of device plugins.

3) device/server.py registers each plugin with core management api via a service_registration_payload. This payload has service_port as 0 and it remains same for all the plugins it calls. This causes Service.AlreadyExistsWithTheSameAddressAndPort exception to be raised at the time of service registration and as such I have temporarily disabled it in instance.py till a solution is worked out.

4) Tests will follow.

### DIFF
--- a/src/python/foglamp/device/coap_device.py
+++ b/src/python/foglamp/device/coap_device.py
@@ -76,6 +76,7 @@ def plugin_shutdown(data):
     pass
 
 
+# TODO: Implement FOGL-701 (implement AuditLogger which logs to DB and can be used by all ) for this class
 class CoAPIngest(aiocoap.resource.Resource):
     """Handles incoming sensor readings from CoAP"""
 
--- a/src/python/foglamp/device/coap_device.py
+++ b/src/python/foglamp/device/coap_device.py
@@ -46,7 +46,7 @@ def plugin_info():
             'interface': '1.0', 'config': _DEFAULT_CONFIG}
 
 
-async def plugin_init(config):
+def plugin_init(config):
     """Registers CoAP handler to accept sensor readings"""
 
     uri = config['uri']['value']
@@ -64,15 +64,15 @@ async def plugin_init(config):
     return {}
 
 
-async def plugin_run(data):
+def plugin_run(data):
     pass
 
 
-async def plugin_reconfigure(config):
+def plugin_reconfigure(config):
     pass
 
 
-async def plugin_shutdown(data):
+def plugin_shutdown(data):
     pass
 
 
--- a/src/python/foglamp/device/coap_device.py
+++ b/src/python/foglamp/device/coap_device.py
@@ -64,7 +64,7 @@ def plugin_init(config):
     return {}
 
 
-def plugin_run(data):
+def plugin_start(data):
     pass
 
 
--- a/src/python/foglamp/device/http_south_device.py
+++ b/src/python/foglamp/device/http_south_device.py
@@ -4,68 +4,78 @@
 # See: http://foglamp.readthedocs.io/
 # FOGLAMP_END
 
-"""CoAP handler for sensor readings"""
+"""HTTP Listener handler for sensor readings"""
 
+from aiohttp import web
 import asyncio
-import json
-
-import aiocoap.resource
-import cbor2
-
 from foglamp import logger
+from foglamp.web import middleware
 from foglamp.device.ingest import Ingest
 
-__author__ = "Terris Linenbach"
+__author__ = "Amarendra K Sinha"
 __copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
 __license__ = "Apache 2.0"
 __version__ = "${VERSION}"
 
 _LOGGER = logger.setup(__name__)
 
+_CONFIG_CATEGORY_NAME = 'http_south'
+_CONFIG_CATEGORY_DESCRIPTION = 'South Plugin HTTP Listener'
 _DEFAULT_CONFIG = {
     'plugin': {
-         'description': 'Python module name of the plugin to load',
+         'description': 'http_south_device',
          'type': 'string',
-         'default': 'coap'
+         'default': 'http_south'
     },
     'port': {
         'description': 'Port to listen on',
         'type': 'integer',
-        'default': '5683',
+        'default': '6683',
     },
     'uri': {
         'description': 'URI to accept data on',
         'type': 'string',
-        'default': 'sensor-values',
+        'default': '0.0.0.0',
     }
 }
 
 
 def plugin_info():
-    return {'name': 'CoAP Server', 'version': '1.0', 'mode': 'async', 'type': 'device',
+    return {'name': 'http_south', 'version': '1.0', 'mode': 'async', 'type': 'device',
             'interface': '1.0', 'config': _DEFAULT_CONFIG}
 
 
 async def plugin_init(config):
-    """Registers CoAP handler to accept sensor readings"""
+    """Registers HTTP Listener handler to accept sensor readings"""
+
+    _LOGGER.info("Retrieve HTTP Listener Configuration %s", config)
 
     uri = config['uri']['value']
     port = config['port']['value']
 
-    root = aiocoap.resource.Site()
-
-    root.add_resource(('.well-known', 'core'),
-                      aiocoap.resource.WKCResource(root.get_resources_as_linkheader))
-
-    root.add_resource(('other', uri), CoAPIngest())
-
-    asyncio.ensure_future(aiocoap.Context.create_server_context(root, bind=('::', int(port))))
-
-    return {}
+    return {'uri': uri, 'port': port}
 
 
 async def plugin_run(data):
-    pass
+    """
+
+    Args:
+        data: dict returned from plugin_init()
+
+    Returns: (server, handler) tuple
+
+    """
+    host = data['uri']
+    port = data['port']
+
+    loop = asyncio.get_event_loop()
+
+    app = web.Application(middlewares=[middleware.error_middleware])
+    app.router.add_route('POST', '/device', HttpSouthIngest.render_post)
+    handler = app.make_handler()
+    coro = loop.create_server(handler, host, port)
+    server = await coro
+    return server, handler
 
 
 async def plugin_reconfigure(config):
@@ -76,8 +86,8 @@ async def plugin_shutdown(data):
     pass
 
 
-class CoAPIngest(aiocoap.resource.Resource):
-    """Handles incoming sensor readings from CoAP"""
+class HttpSouthIngest():
+    """Handles incoming sensor readings from HTTP Listner"""
 
     @staticmethod
     async def render_post(request):
@@ -102,54 +112,52 @@ class CoAPIngest(aiocoap.resource.Resource):
                             }
                         }
                     }
+        Example: curl -X POST http://localhost:6683 -d '{"timestamp": "2017-01-02T01:02:03.23232Z-05:00", "asset": "pump1", "key": "80a43623-ebe5-40d6-8d80-3f892da9b3b4", "readings": {"velocity": "500", "temperature": {"value": "32", "unit": "kelvin"}}}'
         """
-        # aiocoap handlers must be defensive about exceptions. If an exception
-        # is raised out of a handler, it is permanently disabled by aiocoap.
-        # Therefore, Exception is caught instead of specific exceptions.
-
         # TODO: The payload is documented at
         # https://docs.google.com/document/d/1rJXlOqCGomPKEKx2ReoofZTXQt9dtDiW_BHU7FYsj-k/edit#
         # and will be moved to a .rst file
 
-        code = aiocoap.numbers.codes.Code.INTERNAL_SERVER_ERROR
+        code = web.HTTPInternalServerError.status_code
         increment_discarded_counter = True
         message = ''
 
         try:
             if not Ingest.is_available():
-                message = '{"busy": true}'
+                message = {"busy": True}
             else:
-                payload = cbor2.loads(request.payload)
+                payload = await request.json()
 
                 if not isinstance(payload, dict):
                     raise ValueError('Payload must be a dictionary')
 
                 asset = payload.get('asset')
                 timestamp = payload.get('timestamp')
-
                 key = payload.get('key')
 
                 # readings and sensor_readings are optional
                 try:
-                    readings = payload['readings']
+                    readings = payload.get('readings')
                 except KeyError:
                     readings = payload.get('sensor_values')  # sensor_values is deprecated
 
                 increment_discarded_counter = False
 
-                await Ingest.add_readings(asset=asset, timestamp=timestamp, key=key,
-                                          readings=readings)
+                await Ingest.add_readings(asset=asset, timestamp=timestamp, key=key, readings=readings)
 
                 # Success
-                code = aiocoap.numbers.codes.Code.VALID
+                message = readings
+                code = web.HTTPOk.status_code
         except (ValueError, TypeError) as e:
-            code = aiocoap.numbers.codes.Code.BAD_REQUEST
-            message = json.dumps({message: str(e)})
-        except Exception:
-            _LOGGER.exception('Add readings failed')
+            code = web.HTTPBadRequest.status_code
+            message = {'error': str(e)}
+            _LOGGER.exception('ValueError/TypeError in Add readings - failed')
+        except Exception as e:
+            message = {'error': str(e)}
+            _LOGGER.exception('Exception in Add readings - failed')
 
         if increment_discarded_counter:
             Ingest.increment_discarded_readings()
 
-        return aiocoap.Message(payload=message.encode('utf-8'), code=code)
+        return web.json_response({'message':message, 'status':code})
 
--- a/src/python/foglamp/device/http_south_device.py
+++ b/src/python/foglamp/device/http_south_device.py
@@ -5,7 +5,7 @@
 # FOGLAMP_END
 
 """HTTP Listener handler for sensor readings"""
-
+import sys
 from aiohttp import web
 import asyncio
 from foglamp import logger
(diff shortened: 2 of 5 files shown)


amarendra-dianomic:
@MarkRiddoch Please also note that I have used a fixed port 6683 for http_south. Is this ok? Or shall I make it dynamic? Since service registration is done with service_port as 0, this information is not available with management api. Then if made dynamic, how clients will discover it?

src/python/foglamp/microservice_management/service_registry/instance.py
```diff
@@ -97,8 +97,8 @@ def register(cls, name, s_type, address, port, management_port,  protocol='http'
             else:
                 raise Service.AlreadyExistsWithTheSameName
 
-            if cls.check_address_and_port(address, port):
-                raise Service.AlreadyExistsWithTheSameAddressAndPort
+            # if cls.check_address_and_port(address, port):
```
MarkRiddoch: This is a known bug, Jira-692

src/python/foglamp/device/server.py
```diff
@@ -87,45 +88,45 @@ class Server:
...
-                raise
+            # TODO: Examine if below code is as per Architecture
+            # try:
+            #     plugin_module_name = config['plugin']['value']
+            # except KeyError:
+            #     _LOGGER.warning("Unable to obtain configuration of module for plugin {}".format(cls._plugin_name))
+            #     raise
 
+            # Retrieve the plugin name and hence plugin module name, from the params
+            plugin_module_name = cls._plugin_name
```
MarkRiddoch: Why has this changed? Work was done in the previous sprint to pick up the correct plugin name, this appears to be undoing it.

src/python/foglamp/device/coap_device.py
```diff
@@ -46,7 +46,7 @@ def plugin_info():
             'interface': '1.0', 'config': _DEFAULT_CONFIG}
 
 
-def plugin_init(config):
+async def plugin_init(config):
```
MarkRiddoch: Do all of these need to be async? This is effectively changing the API.

src/python/foglamp/device/server.py
amarendra-dianomic: @MarkRiddoch please note that previous code required that the plugin config pre-existed in the configration table, which is contrary to the Architecture specs. Hence I have restored the commented lines AND have moved these lines in the except block. In this way, the code will work first time when there is no config in the db, via except block, and will work second time onwards via the main try block when the config is already available in db. Hope this is ok.

src/python/foglamp/device/coap_device.py
amarendra-dianomic: @MarkRiddoch I have removed async from defs and have tried to implement the feature _plugin_data in device/server.py/_start(). Please let me know if this is ok.

src/python/foglamp/device/server.py
```diff
@@ -87,45 +92,50 @@ class Server:
...
             # TODO: Register for config changes
 
             cls._plugin_data = cls._plugin.plugin_init(config)
+
+            if isinstance(cls._plugin_data, tuple):
+                # If plugin is an aiohttp listener, then return tuple (coro, handler) from plugin_init
+                coro, cls._plugin_server_handler = cls._plugin_data
+                if coro is not None:
+                    cls._plugin_server = await coro
+
```
praveen-garg: instead of this, should not we pass current event_loop as an arg? and let http listener plugin to add aiohttp start corutine added to it; 

Understood that device >server> start def run loop forever, but here adding it as coro should not be an issue; as it happens before!

src/python/foglamp/device/server.py
```diff
@@ -167,17 +177,19 @@ def _register_microservice(cls):
...
         if r.status in range(500, 600):
-            _LOGGER.error("Server error code: %d", r.status)
+            _LOGGER.error("Server error code: %d. Plugin {}".format(cls._plugin_name), r.status)
         res = r.read().decode()
         conn.close()
         response = json.loads(res)
         try:
             cls._microservice_id = response["id"]
             _LOGGER.info('Device - Registered Service %s', response["id"])
+        except (Service.AlreadyExistsWithTheSameAddressAndManagementPort, Service.AlreadyExistsWithTheSameAddressAndPort) as ex:
```
praveen-garg: don;t expect now `AlreadyExistsWithTheSameAddressAndPort`

src/sql/foglamp_init_data.sql
```diff
@@ -75,8 +75,8 @@ INSERT INTO foglamp.configuration ( key, description, value )
 
 -- COAP:  CoAP device server
 --        plugin: python module to load dynamically
-INSERT INTO foglamp.configuration ( key, description, value )
-     VALUES ( 'COAP', 'CoAP Plugin Configuration', ' { "plugin" : { "type" : "string", "value" : "coap", "default" : "coap", "description" : "Python module name of the plugin to load" } } ');
+--INSERT INTO foglamp.configuration ( key, description, value )
```
praveen-garg: TODO https://scaledb.atlassian.net/browse/FOGL-698 ?

src/python/foglamp/device/http_south_device.py
```diff
@@ -0,0 +1,151 @@
...
+                    readings = payload.get('readings')
+                except KeyError:
+                    readings = payload.get('sensor_values')  # sensor_values is deprecated
+
+                increment_discarded_counter = False
+
+                await Ingest.add_readings(asset=asset, timestamp=timestamp, key=key, readings=readings)
+
+                # Success
+                message = readings
```
praveen-garg: please don't respond back with readings; just say 'recieved' or success

src/python/foglamp/device/http_south_device.py
```diff
@@ -0,0 +1,151 @@
...
+
+def plugin_run(data):
+    pass
+
+def plugin_reconfigure(config):
+    pass
+
+
+def plugin_shutdown(data):
+    pass
```
praveen-garg: shutdown aiohttp instance?

`loop.run_until_complete(shutdown(server, handler, app)) ` ?

```

async def shutdown(server, app, handler):
    server.close()
    await server.wait_closed()
    await app.shutdown()
    await handler.finish_connections(10.0) # TODO configurable ? 
    await app.cleanup()
```

src/python/foglamp/device/http_south_device.py
```diff
@@ -0,0 +1,151 @@
...
+
+    app = web.Application(middlewares=[middleware.error_middleware])
+    app.router.add_route('POST', '/', HttpSouthIngest.render_post)
+    handler = app.make_handler()
+    coro = loop.create_server(handler, uri, port)
+
+    return coro, handler
+
+
+def plugin_run(data):
```
praveen-garg: is this plugin_start or run; seems we say it run now! though specs says `start`. TRIVIAL

src/python/foglamp/device/http_south_device.py
```diff
@@ -0,0 +1,151 @@
...
+    pass
+
+def plugin_reconfigure(config):
+    pass
```
praveen-garg: should we add TODO that it will be called on configuration manager callback run?

src/python/foglamp/device/http_south_device.py
```diff
@@ -0,0 +1,151 @@
...
+        Example: curl -X POST http://localhost:6683 -d '{"timestamp": "2017-01-02T01:02:03.23232Z-05:00", "asset": "pump1", "key": "80a43623-ebe5-40d6-8d80-3f892da9b3b4", "readings": {"velocity": "500", "temperature": {"value": "32", "unit": "kelvin"}}}'
+        """
+        # TODO: The payload is documented at
+        # https://docs.google.com/document/d/1rJXlOqCGomPKEKx2ReoofZTXQt9dtDiW_BHU7FYsj-k/edit#
+        # and will be moved to a .rst file
+
+        code = web.HTTPInternalServerError.status_code
+        increment_discarded_counter = True
+        message = ''
+
```
praveen-garg: Be positive :)
`code = web.HTTPOk.status_code`
`message = {'result': 'success'}`

`web.HTTPInternalServerError.status_code` should be assigned in `except Exception as e:`

amarendra-dianomic:
@MarkRiddoch I have incorporated some of the above feedback in the PR. Please note that although I have retained asyc defs for plugin_start() and plugin_shutdown() but now device/server.py has been modified to be flexible enough to use sync/async start and shutdown as per definition. So coap_device.py remains unchanged. Hope this is ok.

amarendra-dianomic:
@MarkRiddoch for http_south plugin, I have created root endpoint e.g. http://localhost:6683/ only. Please let me know if this ok otherwise please suggest another route.

src/python/foglamp/device/server.py
amarendra-dianomic: @MarkRiddoch related to this is another query. How a plugin is supposed to supply its config info to DB the very first time the system is started? This line serves that purpose only. Second time onwards, this line will never be executed.

amarendra-dianomic: @MarkRiddoch related to this is another query. How a plugin is supposed to supply its config info to DB the very first time the system is started? This line serves that purpose only. Second time onwards, this line will never be executed.

src/python/foglamp/device/http_south_device.py
```diff
@@ -0,0 +1,166 @@
...
+def plugin_reconfigure(config):
+    pass
+
+
+def plugin_shutdown(data):
+    app = data['app']
+    handler = data['handler']
+    server = data['server']
+
+    server.close()
```
praveen-garg: what about this?

src/python/foglamp/device/http_south_device.py
```diff
@@ -0,0 +1,166 @@
...
+    asyncio.ensure_future(server.wait_closed())
+    asyncio.ensure_future(app.shutdown())
+    asyncio.ensure_future(handler.shutdown(60.0))
+    asyncio.ensure_future(app.cleanup())
```
praveen-garg: can not we wrap this into a _shutdown and call from entry point plugin_shutdown as _shutdown with ensure future / wait for task done

src/sql/foglamp_init_data.sql
```diff
@@ -75,8 +75,9 @@ INSERT INTO foglamp.configuration ( key, description, value )
 
 -- COAP:  CoAP device server
 --        plugin: python module to load dynamically
-INSERT INTO foglamp.configuration ( key, description, value )
-     VALUES ( 'COAP', 'CoAP Plugin Configuration', ' { "plugin" : { "type" : "string", "value" : "coap", "default" : "coap", "description" : "Python module name of the plugin to load" } } ');
+-- FOGL-698 - Fix configuration entry for CoAP plugin (init.sql vs via configuration manager)
+--INSERT INTO foglamp.configuration ( key, description, value )
```
praveen-garg: keep the TODO but revert! and add your entry too.

src/python/foglamp/device/http_south_device.py
amarendra-dianomic: server will stop accepting requesting

praveen-garg: I mean  should not be ensured?

amarendra-dianomic: Not required as per aiohttp doc example at http://aiohttp.readthedocs.io/en/stable/web.html

amarendra-dianomic:
@ashwinscale @MarkRiddoch @praveen-garg changes done per feedback. Please check.

src/python/foglamp/device/http_south_device.py
```diff
@@ -0,0 +1,166 @@
...
+    host = data['host']
+    port = data['port']
+
+    loop = asyncio.get_event_loop()
+
+    app = web.Application(middlewares=[middleware.error_middleware])
+    app.router.add_route('POST', '/', HttpSouthIngest.render_post)
+    handler = app.make_handler()
+    coro = loop.create_server(handler, host, port)
+    server = asyncio.ensure_future(coro)
```
praveen-garg: move aiohttp start and shutdown block to try / exepct and write to exception

src/python/foglamp/device/http_south_device.py
```diff
@@ -0,0 +1,188 @@
...
+    'host': {
+        'description': 'URI to accept data on',
+        'type': 'string',
+        'default': '0.0.0.0',
+    },
+    'uri': {
+        'description': 'URI to accept data on',
+        'type': 'string',
+        'default': 'sensor-reading',
+    }
```
praveen-garg: I'm not sure, if we should have only 1 key for url as `http(s)://host:port/uri_handler` or url composing units; do we have the (scaling) capability here to be SSL enabled later or sooner? (and hence the protocol can also be httpS vs http)

src/python/foglamp/device/http_south_device.py
```diff
@@ -0,0 +1,188 @@
...
+        handler = app.make_handler()
+        coro = loop.create_server(handler, host, port)
+        server = asyncio.ensure_future(coro)
+
+        data['app'] = app
+        data['handler'] = handler
+        data['server'] = server
+    except Exception as e:
+        _LOGGER.exception(str(e))
+        sys.exit(1)
```
praveen-garg: should be just raise? and let caller/ starter handle it?

src/python/foglamp/device/http_south_device.py
```diff
@@ -0,0 +1,188 @@
...
+def plugin_shutdown(data):
+    try:
+        app = data['app']
+        handler = data['handler']
+        server = data['server']
+
+        server.close()
+        asyncio.ensure_future(server.wait_closed())
+        asyncio.ensure_future(app.shutdown())
+        asyncio.ensure_future(handler.shutdown(60.0))
```
praveen-garg: not sure, it should be hard-coded or configurable  ?

src/python/foglamp/device/http_south_device.py
amarendra-dianomic: This calls for an architecture level decision and discussion and this issue can be taken up separately.

src/python/foglamp/device/http_south_device.py
amarendra-dianomic: Once starting the plugin fails, this is the best course of action. May be this task then will be restarted manually by Admin.

src/python/foglamp/device/http_south_device.py
amarendra-dianomic: This calls for an architecture level decision and discussion and this issue can be taken up separately.

src/python/foglamp/device/http_south_device.py
MarkRiddoch: This fits the same model as CoAP ands hence matches the requirement in the Jira.

src/python/foglamp/device/server.py
```diff
@@ -18,6 +18,7 @@
 from foglamp.device.ingest import Ingest
 from foglamp.microservice_management import routes
 from foglamp.web import middleware
+from foglamp.microservice_management.service_registry.instance import Service
```
MarkRiddoch: I don't see any changes that means that this new import is needed. Is this a hangover from changes that have been backed out?

src/python/foglamp/microservice_management/service_registry/instance.py
```diff
@@ -97,8 +97,9 @@ def register(cls, name, s_type, address, port, management_port,  protocol='http'
             else:
                 raise Service.AlreadyExistsWithTheSameName
 
-            if cls.check_address_and_port(address, port):
-                raise Service.AlreadyExistsWithTheSameAddressAndPort
+            # TODO: FOGL-692 - Service Registry should support services that do not have a service port
+            # if cls.check_address_and_port(address, port):
+            #     raise Service.AlreadyExistsWithTheSameAddressAndPort
```
MarkRiddoch: This is being dealt with in another Jira task and will cause merge difficulties, please back this change out before checking in. It will be resolved in the merge later.

amarendra-dianomic:
@MarkRiddoch all changes done as advised. Please check.

src/python/foglamp/device/http_south_device.py
```diff
@@ -0,0 +1,188 @@
...
+         'type': 'string',
+         'default': 'http_south'
+    },
+    'port': {
+        'description': 'Port to listen on',
+        'type': 'integer',
+        'default': '6683',
+    },
+    'host': {
+        'description': 'URI to accept data on',
```
ashwinscale: duplicate description

src/python/foglamp/device/http_south_device.py
```diff
@@ -0,0 +1,188 @@
...
+from foglamp.device.ingest import Ingest
+
+__author__ = "Amarendra K Sinha"
+__copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
+__license__ = "Apache 2.0"
+__version__ = "${VERSION}"
+
+_LOGGER = logger.setup(__name__)
+
+_CONFIG_CATEGORY_NAME = 'http_south'
```
ashwinscale: Note that this should be consistent with the original category name HTTP_SOUTH. However, currently server.py currently does not utilize this part of plugin_info (which is an existing issue). COAP suffers the same.

praveen-garg:
Can we rebase (/retest) and merge now?

amarendra-dianomic:
Closed in favour of [new PR](https://github.com/foglamp/FogLAMP/pull/335).

PR #289, opened 2017-11-10
FOGL-684 - move purge and statistics history to new locations

MarkRiddoch:


sstefano69:
I tried the branch  and it works fine.

PR #290, opened 2017-11-10
FOGL-679_FOGL-680 south service and coap_listen plugin

ashwinscale:
also a small fix on init sql for statistics


izoratti-dianomic:
Just a reminder, I will add a task on Jira - once everything is in place, I must merge and manually handle the conflicts with FOGL-610 - https://github.com/foglamp/FogLAMP/tree/FOGL-610
It's not a big deal, it's only the SQL init script, but since there have been several changes, it needs to be adapted. I commented here just because I saw Ashwin's comments.

PR #293, opened 2017-11-15
added init.sql to comply with the new structure.

izoratti-dianomic:


C/plugins/storage/postgres/init.sql
```diff
@@ -45,42 +45,30 @@
...
-  REPLICATION;
-
-ALTER USER foglamp ENCRYPTED PASSWORD 'foglamp';
 
 
 -- Create the foglamp tablespace
 CREATE TABLESPACE foglamp
   OWNER foglamp
-  LOCATION '/var/lib/postgresql/9.6/';
+  LOCATION '/usr/local/foglamp/data/storage/postgres';
```
MarkRiddoch: What if we have not set FOGLAMP_HOME to /usr/local? Can we make this work with FOGLAMP_HOME? Or in fact should it not be FOGLAMP_DATA? Thinking in particular of how this would work in a snap package.

izoratti-dianomic: Sure, I thought about that. 
In this first version, the plugin works with Linux user foglamp and FOGLAMP_ROOT: /usr/local/foglamp (so FOGLAMP_DATA: /usr/local/foglamp/data)
The CREATE TABLESPACE command accepts only an absolute path, therefore the package installer must change it in a post-install.
There are many options to avoid this, but they are fairly complicated, I would suggest to keep it simple for the moment and leave it like this, then change it in one of the next sprints.

MarkRiddoch: Sure, let's add a Jira for a future sprint

PR #305, opened 2017-11-18
FOGL-703 foglamp.common.process and FoglampProcess class

ashwinscale:
- [x] Parsing the mandatory command line arguments --name, --address and --port
- [x] Tracking the management interface of the core microservice
- [x] Maintaining the micro service manage api client used to talk with the core micro service and in particular the service registry
- [x] The connection to the storage micro service and the clients for talking to the storage and readings interface
- [x] This class will exist in foglamp.common.process and will be called FoglampProcess.
- [x] The class should have a pure virtual method, run, which will be overridden by each process implementation and will provide the method that starts the task specific portion of the FogLAMP process.

Questions:
How should the logger be treated here?
Currently the StorageClient and ReadingsStorageClient accept host and port as parameters and manages the connection to storage. Does this need to be move out as well, or can it stand as is?
Should this PR also include changes to services and tasks as users of the new class?


MarkRiddoch:
In answer to your questions Ashwin:

_How should the logger be treated here?_
   Adding the logger here is a good idea, if only so that we have common logging when we start processes
_Currently the StorageClient and ReadingsStorageClient accept host and port as parameters and manages the connection to storage. Does this need to be move out as well, or can it stand as is?_
   This can stay as it is for now. The new class should manage the passing of the host name port etc.
_Should this PR also include changes to services and tasks as users of the new class?_
   Yes, changes to the tasks certainly. Don't do the services as they should use the child class that is designed for use by services, see FOGL-704

Makefile
```diff
@@ -61,6 +61,7 @@ default : c_build $(SYMLINK_SERVICES_DIR) $(SYMLINK_PLUGINS_DIR) python_build py
 # This first does a make to build anything needed for the installation.
 install : $(INSTALL_DIR) c_install python_install python_requirements
 
+###############################################################################
```
MarkRiddoch: Not really related to this change

python/foglamp/common/process.py
```diff
@@ -0,0 +1,152 @@
...
+        elif self._name is None:
+            raise ValueError("--name is not specified")
+
+        self._m_client = self.MicroserviceManagementClient(self._core_management_host,self._core_management_port)
+        self._readings_storage = ReadingsStorageClient(self._core_management_host, self._core_management_port)
+        self._storage = StorageClient(self._core_management_host, self._core_management_port)
+
+    # pure virtual method run() to be implemented by child class
+    @abstractmethod
+    def run(self):
```
praveen-garg: @ashwinscale  will not this collide with configuration change callback? (Rr we are going to change that to `reconfigure`

python/foglamp/tasks/statistics/__main__.py
```diff
@@ -0,0 +1,23 @@
...
+
+from foglamp.tasks.statistics.statistics_history import StatisticsHistory
+from foglamp.common import logger
+
+__author__ = "Terris Linenbach, Vaibhav Singhal"
+__copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
+__license__ = "Apache 2.0"
+__version__ = "${VERSION}"
+
+if __name__ == '__main__':
```
praveen-garg: as `__main__.py` is package entry point, we can avoid `if __name__ == '__main__':`

praveen-garg:
Left 2 comments:
1. Will not `run` collide with configuration change callback? (Rr we are going to change that to reconfigure
2. As __main__.py is package entry point, we can avoid if __name__ == '__main__':

python/foglamp/common/process.py
ashwinscale: I don't think it would collide. A callback's run() method would exist in it's own module. The code is basically:
...
cb = import_module(callback)
...
cb.run(category_name)
...

PR #306, opened 2017-11-20
FOGL-729 explicitly use --system or --user flag on pip install commands

ashwinscale:
Note than pip uninstall does NOT allow --user or --system to be specified.
@izoratti @izoratti-dianomic please let me know if this fixes the issue you were having.


Makefile
```diff
@@ -121,14 +127,16 @@ $(INSTALL_DIR) :
...
 ###############################################################################
 # clean
-clean : 
+clean : uninstall_python_requirements
 	-$(RM_DIR) $(CMAKE_BUILD_DIR)
 	-$(RM_DIR) $(PYTHON_BUILD_DIR)
 	-$(RM) $(SYMLINK_SERVICES_DIR)
 	-$(RM) $(SYMLINK_PLUGINS_DIR)
 
 # uninstall
```
izoratti-dianomic: I don't think we should have an uninstall - I have never seen it in make. 
I would remove this whole section.

Makefile
```diff
@@ -99,7 +101,11 @@ python_build : $(PYTHON_SETUP_FILE)
 
 # install python requirements
```
izoratti-dianomic: I think pip should be executed in this way:
- with _make_, execute _pip --user_
- with _make install_, execute _pip --system_

Makefile
```diff
@@ -6,6 +6,8 @@ MKDIR_PATH := mkdir -p
 CD := cd
 LN := ln -sf
 CMAKE := cmake
+PIP_USER_FLAG = --user
+PIP_SYSTEM_FLAG = --system
 PIP_INSTALL_REQUIREMENTS := pip3 install -Ir
 PIP_UNINSTALL_REQUIREMENTS := pip3 uninstall -yr
```
izoratti-dianomic: I don't think we must "uninstall" anything.

izoratti-dianomic: Are lines 30 and 31 correct? I thought we did not have _src_ anymore.

PR #318, opened 2017-11-21
Rescue of FOGL-676 branch

MarkRiddoch:


ashwinscale:
Issue 1:
Another issue when installing with a non-foglamp linux user and export DB_CONNECTION="dbname=foglamp user=foglamp" isn't specified.
ususer@us1704:~/fogl676/FogLAMP$ curl -X POST http://0.0.0.0:41009/storage/table/statistics_history -d @payloac
{ "entryPoint" : "insert", "message" : "ERROR:  permission denied for sequence statistics_history_id_seq
", "retryable" : false}

Maybe:
https://stackoverflow.com/questions/9325017/error-permission-denied-for-sequence-cities-id-seq-using-postgres
stackoverflow.com
ERROR: permission denied for sequence cities_id_seq using Postgres
I'm new at postgres (and at database info systems all in all). I ran following sql script on my database: create table cities ( id serial primary key, name text not null ); create table reports (...

but it all works if I do export DB_CONNECTION

Issue 2:
Also line C/plugins/storage/postgres/connection.cpp: ln 582
-       sql.append("DELETE from ");
+       sql.append("DELETE from foglamp.");

izoratti-dianomic:
@ashwinscale: DB_CONNECTION="dbname=foglamp user=foglamp" 
This is incorrect. The connection string must contain only the dbname (i.e. foglamp), the user is set as the Linux user or by the PGUSER env var.
Now that we have reverted to any version of PostgreSQL, i.e. we do not control how Postgres is configured, we cannot predict how FogLAMP will connect to PG, so we leave it to the developer.
In the standard Ubuntu installation, the Linux and PG users are peered, hence the superuser must create a PG user with the same name of the Linux user who will start FogLAMP.
In the Debian installation from PG.org. PG authentication is via MD5, hence the FogLAMP user must specify the UID and PWD, either via env vars (PGUSER/PGPASS) or PGPASSFILE. Other issues may be related to the type of connectivity, i.e. which socket file or host or port, all handled via PGHOST env var.

PR #325, opened 2017-11-22
FOGL-695 - Use log levels defined in init.sql for the audit trail

MarkRiddoch:


praveen-garg:
it would be great if have a consistent log (severity( level across services and defined in init.sql and a common Db(Audit)Logger=>Severity class/ enum; e.g. python has 0 as NOT-SET, 10-50; not sure about storage service writing to DB logs as well? and what levels. common usage implementation will help all https://scaledb.atlassian.net/browse/FOGL-701

PR #335, opened 2017-11-24
FOGL-687 - Implement HTTP southern plugin per new dir structure

amarendra-dianomic:


python/foglamp/plugins/south/http_south/http_south.py
```diff
@@ -0,0 +1,190 @@
...
+            'interface': '1.0', 'config': _DEFAULT_CONFIG}
+
+
+def plugin_init(config):
+    """Registers HTTP Listener handler to accept sensor readings"""
+
+    _LOGGER.info("Retrieve HTTP Listener Configuration %s", config)
+
+    host = config['host']['value']
+    port = config['port']['value']  # Make port dynamic? its unattended env! expose as service port?
```
MarkRiddoch: There is nothing to gain my making this dynamic as it is a port that has to be known by the sensors sending to it.

This is not the service port as the service port id the port for REST API services.

python/foglamp/plugins/south/http_south/http_south.py
```diff
@@ -0,0 +1,190 @@
...
+class HttpSouthIngest(object):
+    """Handles incoming sensor readings from HTTP Listener"""
+
+    @staticmethod
+    async def render_post(request):
+        """Store sensor readings from CoAP to FogLAMP
+
+        Args:
+            request:
+                The payload is a cbor-encoded array that decodes to JSON
```
MarkRiddoch: Strictly speaking it is not cbor

PR #339, opened 2017-11-27
FOGL-573 (Part 1) statistics test 

praveen-garg:
(stats history test ~will be in new PR~ are in https://github.com/foglamp/FogLAMP/pull/343)

For this note the deviation https://docs.google.com/document/d/1JJDP7g25SWerNVCxgff02qp9msHbqA9nt3RAFx8-Qng/edit#heading=h.wxe1fbyc0ddz

but (there are 2 sent keys now; SENT_1 and SENT_2; and  all keys are upper-case) // see comments in doc and open the issue to fix mapping etc.

### DIFF
--- a/tests/unit-tests/python/foglamp_test/services/core/api/test_statistics.py
+++ b/tests/unit-tests/python/foglamp_test/services/core/api/test_statistics.py
@@ -1,0 +1,129 @@
+# -*- coding: utf-8 -*-
+
+# FOGLAMP_BEGIN
+# See: http://foglamp.readthedocs.io/
+# FOGLAMP_END
+
+import asyncpg
+import requests
+import pytest
+
+__author__ = "Praveen Garg"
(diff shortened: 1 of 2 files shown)


tests/unit-tests/python/foglamp_test/services/core/api/test_statistics.py
```diff
@@ -0,0 +1,129 @@
...
+
+__DB_NAME = "foglamp"
+BASE_URL = 'http://localhost:8081/foglamp'
+headers = {"Content-Type": 'application/json'}
+
+pytestmark = pytest.mark.asyncio
+
+
+async def add_statistics_test_data():
+    conn = await asyncpg.connect(database=__DB_NAME)
```
MarkRiddoch: @izoratti - you started to discuss ways of using the database in the unit tests that did not directly access it like this. We need to move that forward as we are getting tests making direct access like this and will need to go back and change then when the generic solution is implemented.

PR #343, opened 2017-11-28
FOGL-573 (Part 2) statistics history REST api test; 

praveen-garg:
ugly! takes 45+ seconds once test db is there, will make better :]

```
tests/unit-tests/python/foglamp_test/services/core/api/test_statistics_history.py ...

========================== 3 passed in 45.30 seconds ===========================
```

### DIFF
--- a/python/foglamp/services/core/api/statistics.py
+++ b/python/foglamp/services/core/api/statistics.py
@@ -105,5 +105,5 @@ async def get_statistics_history(request):
     # Append the last set of records which do not get appended above
     results.append(temp_dict)
 
-    # TODO: find out where from this "interval" will be picked and what will be its role in query?
-    return web.json_response({"interval": 5, 'statistics': results})
+    # TODO: find this "interval" from statistics task schedule
+    return web.json_response({"interval": 15, 'statistics': results})
--- a/tests/unit-tests/python/foglamp_test/services/core/api/test_statistics_history.py
+++ b/tests/unit-tests/python/foglamp_test/services/core/api/test_statistics_history.py
@@ -1,0 +1,126 @@
+# -*- coding: utf-8 -*-
+
+# FOGLAMP_BEGIN
+# See: http://foglamp.readthedocs.io/
+# FOGLAMP_END
+
+import asyncpg
+import asyncio
+from datetime import datetime
+import requests
+import pytest
+
+__author__ = "Praveen Garg"
+__copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
+__license__ = "Apache 2.0"
+__version__ = "${VERSION}"
+
+__DB_NAME = "foglamp"
+BASE_URL = 'http://localhost:8081/foglamp'
+
+pytestmark = pytest.mark.asyncio
+
+async def set_statistics_test_data(val):
+    conn = await asyncpg.connect(database=__DB_NAME)
+    await conn.execute('''UPDATE foglamp.statistics SET value = $1, previous_value = $2''', val, 0)
+    await conn.close()
+
+
+@pytest.allure.feature("api")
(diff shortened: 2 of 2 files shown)


python/foglamp/services/core/api/statistics.py
```diff
@@ -105,5 +105,5 @@
     # Append the last set of records which do not get appended above
     results.append(temp_dict)
 
-    # TODO: find out where from this "interval" will be picked and what will be its role in query?
-    return web.json_response({"interval": 5, 'statistics': results})
+    # TODO: find this "interval" from statistics task schedule
+    return web.json_response({"interval": 15, 'statistics': results})
```
MarkRiddoch: Ashish has now resolved this I believe (see FOGL-777)

tests/unit-tests/python/foglamp_test/services/core/api/test_statistics_history.py
```diff
@@ -0,0 +1,127 @@
...
+        r = requests.get(BASE_URL + '/statistics/history')
+        res = r.json()
+
+        assert 200 == r.status_code
+        last_count = len(res['statistics']) * 8
+
+        # use fixture
+        await set_statistics_test_data(10)
+
+        assert 15 == res['interval']
```
MarkRiddoch: This may not always  be 15 seconds, it is configurable. We should check it matches the configuration not a fixed value.

tests/unit-tests/python/foglamp_test/services/core/api/test_statistics_history.py
```diff
@@ -0,0 +1,127 @@
...
+
+        # Wait for 15 (as per the task schedule) seconds
+        # to get 1 more batch of statistics updated value in statistics_history
+        # FIXME: we should not wait in actual; but execute the task itself
+        await asyncio.sleep(15)
```
MarkRiddoch: Again this is configurable

praveen-garg:
@MarkRiddoch @ashish-ScaleDB  ready for review again!

python/foglamp/services/core/api/statistics.py
```diff
@@ -59,24 +59,25 @@
...
+            # Remove python side handling date_trunc and use
+            # SELECT date_trunc('second', history_ts::timestamptz)::varchar as history_ts
+
+            payload = PayloadBuilder().AGGREGATE(["count", "*"]).payload()
+            result = storage_client.query_tbl_with_payload("statistics", payload)
+            key_count = result['rows'][0]['count_*']
+
+            payload = PayloadBuilder().SELECT(("history_ts", "key", "value")).LIMIT(limit * key_count).payload()
+        else:
+            raise web.HTTPBadRequest(reason="limit must be an integer")
```
ashish-ScaleDB: Thanks for adding the BadRequest handling

tests/unit-tests/python/foglamp_test/services/core/api/test_statistics_history.py
```diff
@@ -0,0 +1,147 @@
+# -*- coding: utf-8 -*-
+
+# FOGLAMP_BEGIN
+# See: http://foglamp.readthedocs.io/
+# FOGLAMP_END
+
+import asyncpg
+import asyncio
+from datetime import datetime, timedelta
```
ashish-ScaleDB: unused timedelta

praveen-garg: its not unused; https://github.com/foglamp/FogLAMP/pull/343/files#diff-96e9d24e7d9d1665b7a7ab8153e8ef87R45 (in actual code you are using storage layer which gives you string hence you do split with `:` and then convert to seconds; Here its timedelta itself

PR #353, opened 2017-12-03
Implemented foglamp and postgres scripts to run with managed and unmanaged database

izoratti-dianomic:


izoratti-dianomic:
Extra changes needed.

PR #359, opened 2017-12-04
FOGL-784 Significant changes in Managing FogLAMP microservices / start/stop/status, configuration and make/make install.

izoratti-dianomic:


Makefile
```diff
@@ -212,6 +242,22 @@ $(FOGBENCH_PYTHON_INSTALL_DIR) :
 $(EXTRAS_INSTALL_DIR) :
 	$(MKDIR_PATH) $@
 
+###############################################################################
+####################### DATA INSTALL TARGETS ################################
+###############################################################################
```
ashwinscale: Data directories should be a product of init rather than install. I understand that we don't have init yet.

Makefile
```diff
@@ -212,6 +242,22 @@ $(FOGBENCH_PYTHON_INSTALL_DIR) :
...
+# install data
+data_install : $(DATA_INSTALL_DIR) install_data
+
+install_data : $(DATA_INSTALL_DIR) $(DATA_SRC_DIR)
+	$(CP_DIR) $(DATA_SRC_DIR) $(INSTALL_DIR)
+ifdef SUDO_USER
+	chown -R ${SUDO_USER}:${SUDO_USER} $(INSTALL_DIR)/$(DATA_SRC_DIR)
```
ashwinscale: chown is causing a conflict on ubuntu core snap creation.

scripts/common/write_log.sh
```diff
@@ -0,0 +1,72 @@
...
+#             $3 - Output:
+#                  - logonly : send the message only to syslog
+#                  - all     : send the message to syslog and stdout
+#                  - outonly " send the message only to stdout
+#             $4 - Format
+#                - pretty : Do not show the date and priority
+#
+write_log() {
+
+    MODULE="foglamp.storage.postgres"
```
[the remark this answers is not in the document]
izoratti-dianomic: Fixed.

scripts/foglamp
```diff
@@ -48,25 +82,27 @@ fi
 #             $4 - Format
 #                - pretty : Do not show the date and priority
 #
-foglamp_log() {
+write_log() {
```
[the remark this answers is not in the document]
izoratti-dianomic: Fixed.

Makefile
MarkRiddoch: Yes, but only if it dynamic data. Where we are installing data files that come as part of the package then it is right to build them in the make install. If the package needs them to run then they must be, because the package must be able to run following a make install.

Makefile
MarkRiddoch: Ordinarily the chown would be done as part of an install. I can see why you want to do it here, but maybe make it conditional or ignore error returns for the snap creation case

data/etc/foglamp.json
```diff
@@ -0,0 +1,4 @@
+{ "default storage plugin" : "postgres",
```
MarkRiddoch: If we are going to have this here then we should update the storage service to take this into account. It already has its own notion of what is default built into the code. I don't like duplicating this. I suggest raise a Jira that this has caused duplication and we fix it in a future sprint.

data/etc/foglamp.json
```diff
@@ -0,0 +1,4 @@
...
+  "storage plugins"        : { "plugin"  : "postgres",
```
MarkRiddoch: This should be an array to allow for multiple plugins to be tracked.

"storage plugins" : [ { "plugin" : "postgres", "managed" : true }, { "plugin" : "mysql", "managed" : false } ]

scripts/common/get_engine_management.sh
```diff
@@ -0,0 +1,36 @@
...
+
+get_engine_management() {
+
+    # Remove new lines and store in a variable
+    json_foglamp=`tr -d '\n' < $FOGLAMP_DATA/etc/foglamp.json`
+
+    # Remove tabs
+    json_foglamp=`echo $json_foglamp | tr -d '\t'`
+
+    middle_grep="\"plugin\" *: *\"${1}\" *, *\"managed\" *:.*"
```
MarkRiddoch: This would not work if you have multiple plugins. It may be better to invoke python to pull things out of the JSON than try to do it with grep, awk etc. It is likely to be more reliable and flexible.

scripts/storage
```diff
@@ -0,0 +1,106 @@
...
+## Get the default storage database plugin from the foglamp config file
+get_default_storage_plugin() {
+
+    # Remove new lines and store in a variable
+    json_foglamp=`tr -d '\n' < $1`
+
+    # Remove tabs
+    json_foglamp=`echo $json_foglamp | tr -d '\t'`
+
+    echo `echo $json_foglamp | grep -o '"default storage plugin" *:.*' | grep -o ':.*' | grep -o '".*' | cut -d'"' -f2`
```
MarkRiddoch: Using a Python script to extract this would be much easier and less error prone.

scripts/storage
```diff
@@ -0,0 +1,106 @@
...
+# Get the storage database plugin from the Storage microservice cache file
+get_plugin_from_storage() {
+
+    # Remove new lines and store in a variable
+    json_storage=`tr -d '\n' < $1`
+
+    # Remove tabs
+    json_storage=`echo $json_storage | tr -d '\t'`
+
+    echo `echo $json_storage | grep -o '"plugin" *: *{.*' | grep -o '{.*' | grep -o '"value" *:.*' | grep -o ':.*' | grep -o '".*' | cut -d'"' -f2`
```
MarkRiddoch: As above

Makefile
ashwinscale: In this case, this is a directory that must exist in FOGLAMP_DATA, which has variable location. Installing the directory in the default FOGLAMP_DATA directory is okay as a first step. However, if someone is not using the default FOGLAMP_DATA directory, then the user must manually move the directory contents into the non-default FOGLAMP_DATA  directory.

PR #376, opened 2017-12-11
FOGL-812 - Native sensortag poll mode plugin improvements

amarendra-dianomic:
Addressed below:
1) Examined print statements
2) When bluetooth bad address or not really connected, it should not create any more future task etc. and should always check is_connected
3) Battery level reading added

Not addressed:
1) Up/Down Keypress as this works on notification only and not in polling. This feature will be added in async mode plugin.
2) Shutdown - FOGL-764 will take care of this. However, it has been addressed partially here.

### DIFF
--- a/python/foglamp/plugins/south/cc2650poll/cc2650poll.py
+++ b/python/foglamp/plugins/south/cc2650poll/cc2650poll.py
@@ -72,25 +72,29 @@ def plugin_init(config):
     Raises:
     """
     global sensortag_characteristics
+    data = copy.deepcopy(config)
 
     bluetooth_adr = config['bluetoothAddress']['value']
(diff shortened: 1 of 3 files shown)


ashish-ScaleDB:
@amarendra-dianomic [Latest commit](https://github.com/foglamp/FogLAMP/pull/376/commits/8867b18deb1011cefd6d7eab8677cf6c73830e71) looks much better. Readings have been compared with that of sensor tag app on iPhone and Android respectively. Except for movement data, rest of the characterstics accuracy is very close to apps. It may be noted that  even android and iPhone apps differ on few reading values. Later  we should point some benchmark to compare the accuracy of reading values.

PR #387, opened 2017-12-14
Fogl 758 restore local

sstefano69:
- restore
- backup
- Api integration
- code refactored 

commands for either API or command line execution available here :

https://docs.google.com/document/d/1okUiEoafdxcMlgUgINO295Ihlaj3hIXPFS36wPWcvnY/edit#


python/foglamp/plugins/storage/postgres/backup_restore/lib.py
```diff
@@ -7,11 +7,16 @@
 import subprocess
 import time
 import os
+import asyncio
+import json
 from enum import IntEnum
+import psycopg2
```
MarkRiddoch: Can I suggest we have a comment somewhere explain why this script uses the database directly and does not go via storage

izoratti-dianomic: @MarkRiddoch has a very valid point.
I know our testing and build environment is not ready, but we must be 100% sure whatever we develop it works in:
1. development env (i.e. after make)
2. deployment env (i.e. after sudo make install in /usr/local/foglamp)
3. snap on Ubuntu x86_64
4. snap on Ubuntu Core armhf

I know it sounds long and complicated, but it is not - really. Let's talk Monday about that.
/iz

sstefano69: @MarkRiddoch, done

PR #401, opened 2017-12-20
FOGL-851 unregister the (failed) service if no ping (heartbeat) uptime

praveen-garg:
- Log 

`[FOGLAMP] 12-20-2017 14:04:13 - INFO :: monitor: SMNTR: Unregistered the failed micro-service service instance id=4fc79c7d-adf3-4445-bca9-a71f81606087: <Foo, type=Southbound, protocol=http, address=0.0.0.0, service port=40120, management port=40120, status=0>`

**Note:** I assume scheduler should wait for 5s (actually as per configured monitoring interval) before retrying restart of failed service; as heartbeat monitoring is being done every 5s per default config.

### DIFF
--- a/python/foglamp/services/common/microservice_management/service_registry/monitor.py
+++ b/python/foglamp/services/common/microservice_management/service_registry/monitor.py
@@ -8,8 +8,8 @@
 
 import asyncio
 import aiohttp
-import logging
 import json
+import logging
 
 from foglamp.common import logger
 from foglamp.common.configuration_manager import ConfigurationManager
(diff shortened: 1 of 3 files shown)


MarkRiddoch:
It would be better to wait for more than the ping period as it might take some time for a response to come back. It is fairly normal not to take action until two ping periods have expired to compensate for false triggering

praveen-garg:
Yupe, I should have mentioned with some delta; I already put 1s+ in test to that interval and mentioned in comment too

PR #405, opened 2017-12-20
FOGL-884 - South micro service misuses the plugin handle

amarendra-dianomic:


python/foglamp/services/south/server.py
```diff
@@ -154,15 +154,13 @@ class Server(FoglampMicroservice):
...
 
     
     async def _exec_plugin_poll(self, config) -> None:
         """Executes poll type plugin
         """
         await Ingest.start(self._core_management_host, self._core_management_port)
-        self._plugin_handle['ingest'] = Ingest
         max_retry = 3
         try_count = 1
         while True and self._plugin_handle['is_connected'] is True and try_count <= max_retry:
```
MarkRiddoch: This is also an issue. The handle should be opaque so that we can use it in the plugin.

amarendra-dianomic: handle['is_connected'] is being populated in plugin only and it is being used in server.py as an indicator as to whether the device is connected as this info can possibly come from plugin only. Can we achieve this b

MarkRiddoch: Yes, this was never meant to be like this. Handle is not a way for communication between the server and the plugin, it is for the plugin to have instant data of its own. I will merge your change, but keep the Jira as open of this - with a modified description.

PR #414, opened 2017-12-22
FOGL-711 Fixed error return from browser code

MarkRiddoch:


python/foglamp/services/core/api/browser.py
```diff
@@ -82,7 +82,10 @@ def setup(app):
     _storage = connect.get_storage()
     results = _storage.query_tbl_with_payload('readings', payload)
 
-    return web.json_response(results['rows'])
+    if 'rows' in results:
+        return web.json_response(results['rows'])
+    else:
+        raise web.HTTPBadRequest(reason=results['message'])
```
ashish-jabble: Its NOT the case of `Bad Request error` but 200 with No data available for non-numeric readings. We may handle like below:

```
if 'rows' not in results:
     response = results['message']
else:
    response = results['rows']

return web.json_response({reading: response})
```

python/requirements.txt
```diff
@@ -2,6 +2,9 @@
 aiocoap==0.3
 cbor2==4.0.0
 
+# Postgres for backup
+psycopg2==2.7.1
+
```
ashish-jabble: Assuming this addition because we are using it for backup restore process. So all good 👍

MarkRiddoch: Correct, a little naughty to sneak it in with this change.

python/foglamp/services/core/api/browser.py
MarkRiddoch: We could, or as it currently is it says Bad Request with the message saying the data type is not compatible with the request. I'm not sure we should give 200 OK unless we return real values.

PR #416, opened 2017-12-26
(don't merge) var name fix; routing def calls vs internal def separation; few test fixes

praveen-garg:


praveen-garg:
FOGL-904 and other related things will fix. I assume tests will be taken care of with that too.

PR #418, opened 2017-12-27
FOGL-779 Admin API encryption

praveen-garg:
Admin REST Server api (https) TLS 1.2'

Also carry an important fix (equality operators) in storage client `== None` vs `is None`

**Note:** certificate store `<x>`/etc/certs is needed (see  `get_certificates()` def) it goes to search via paths as defined in FOGL-780


python/foglamp/services/core/server.py
```diff
@@ -70,9 +75,22 @@ class Server:
...
+
+    is_rest_server_tls_enabled = False
+
+    _REST_API_DEFAULT_CONFIG = {
+        'port': {
+            'description': 'REST API service port',
+            'type': 'integer',
+            'default': '8081'
+        },
+        'isTlsEnabled': {
```
MarkRiddoch: The spec [The FogLAMP REST API](https://docs.google.com/document/d/1JJDP7g25SWerNVCxgff02qp9msHbqA9nt3RAFx8-Qng/edit#heading=h.v60x3i70chpa) has a different set of configuration options.

praveen-garg: @MarkRiddoch  fixed, please check now.

PR #419, opened 2017-12-28
FOGL-799: Add per asset Ingres statistics

Singhal-Vaibhav:
This is the kind of values you will get in statistics table when this pull is in:

```
{"key" "TI SENSORTAG CC2650/MAGNETOMETER",
"description":"The number of readings received by FogLAMP since startup for sensor TI SENSORTAG CC2650/MAGNETOMETER",
"value":23,
"previous_value":22,
"ts":"2017-12-28 17:31:26.789132+05:30"}
```

This required change in foglamp schema.
Automated tests will be added in a separate pull, https://scaledb.atlassian.net/browse/FOGL-910


python/foglamp/services/south/ingest.py
```diff
@@ -457,6 +460,12 @@ def increment_discarded_readings(cls):
                 cls._discarded_readings_stats += readings
                 _LOGGER.exception('An error occurred while writing discarded statistics')
 
+            try:
+                await stats.add_update(cls._sensor_stats)
+                cls._sensor_stats = {}
+            except Exception:  # TODO catch real exception
```
praveen-garg: except Exception as ex:  and log with str(ex) ? or you mean to handle specific known exceptions stats.add_update can raise

python/foglamp/common/statistics.py
```diff
@@ -53,6 +53,39 @@ def __init__(self, storage):
...
+        """
+        for key, value_increment in sensor_stat_dict.items():
+            try:
+                payload = PayloadBuilder().INSERT(key=key,
+                                                  description="The number of readings received by "
+                                                              "FogLAMP since startup for sensor {}".format(key),
+                                                  value=0, previous_value=0).payload()
+                result = self._storage.insert_into_tbl("statistics", payload)
+                if result['response'] == "inserted":
+                    payload = PayloadBuilder() \
```
praveen-garg: why update after insert? is not it being handled in KeyError

python/foglamp/services/south/ingest.py
```diff
@@ -560,6 +569,12 @@ def is_available(cls) -> bool:
             if cls._stop:
                 raise RuntimeError('The device server is stopping')
 
+        # Increment the count of received readings to be used for statistics update
+        if asset in cls._sensor_stats:
+            cls._sensor_stats[asset.upper()] += 1
+        else:
+            cls._sensor_stats[asset.upper()] = 1
```
praveen-garg: in a batch of readings it will be always 1? or  can be n number of for a specific asset sensor

Singhal-Vaibhav: This code is inside add_readings which are updated for every read we get, so no point of batch here.

ashish-jabble:
As statistics_history REST API tests are failing against this PR, so fix those into FOGL-910 as well. Hence merging.

PR #420, opened 2017-12-28
service announcement fixes for name+type based discovery over  the network

praveen-garg:


python/foglamp/services/core/server.py
```diff
@@ -177,7 +178,8 @@ def _start_core(cls, loop=None):
             # see http://<core_mgt_host>:<core_mgt_port>/foglamp/service for registered services
 
             _logger.info('Announce management API service')
-            cls.management_announcer = ServiceAnnouncer('FogLAMP-Core', '_foglamp_core._tcp', cls.core_management_port,
+            unique_id = str(uuid.uuid4())
+            cls.management_announcer = ServiceAnnouncer('FogLAMP-Core-{}'.format(unique_id), '_foglampcore._tcp', cls.core_management_port,
```
MarkRiddoch: It's important that we have the same name each time we start, therefore the use of a changing uuid is not the best approach. the intention is to allow the name to be configured, with the default being FogLAMP. This configuration item would replace the FogLAMP portion of the current name.

python/foglamp/services/core/server.py
```diff
@@ -203,9 +205,11 @@ def _start_core(cls, loop=None):
             address, service_server_port = service_server.sockets[0].getsockname()
             _logger.info('Rest Server started on http://%s:%s', address, service_server_port)
 
-            cls.admin_announcer = ServiceAnnouncer('FogLAMP', '_foglamp._tcp', service_server_port, ['The FogLAMP Admin REST API'])
-            cls.user_announcer = ServiceAnnouncer('FogLAMP', '_foglamp_app._tcp', service_server_port,
-                                              ['The FogLAMP Application  REST API'])
+            cls.admin_announcer = ServiceAnnouncer('FogLAMP-{}'.format(unique_id), '_foglamp._tcp', service_server_port,
+                                                   ['The FogLAMP Admin REST API'])
+            # do we need this to announce as another service?
+            cls.user_announcer = ServiceAnnouncer('FogLAMP-App-{}'.format(unique_id), '_foglampapp._tcp', service_server_port,
```
MarkRiddoch: As above, we need a consistent name across reboots. We should use a configuration item to replace the FogLAMP part of the name.

praveen-garg:
@MarkRiddoch  please look into FOGL-916 and update desc / specs

praveen-garg:
closing in favor of https://github.com/foglamp/FogLAMP/pull/453

PR #429, opened 2018-01-03
FOGL-780 self signed certificate generating script

praveen-garg:
**TODO: BEFORE MERGE**

- [x] allow to check env var and put these in $FOGLAMP_DATA/etc/certs, $FOGLAMP_ROOT/etc/certs or /usr/local/foglamp/etc/certs

- [x] check openssl is available

- [ ] check minimum openssl version required ?!

- [x] on error/fail write to syslog and exit

- [x] work fine on ubuntu core, raspbian, ubuntu VM and 16.04LTS


scripts/core/selfssl
```diff
@@ -0,0 +1,89 @@
...
+# organizational_unit=
+common_name=FogLAMP
+email=info@dianomic.com
+
+
+# add fail_if_error and log
+
+if [ -z $1 ]; then
+  printf "Enter the SSL Certificate Name:"
+  read SSL_NAME
```
MarkRiddoch: I don't think we want to have interactive use like this - FogLAMP is meant to be standalone/unattended. This script will be used during installation and during an Admin API request, but not be a user sitting at a terminal.

scripts/core/selfssl
```diff
@@ -0,0 +1,89 @@
...
+fi
+
+if [ -z $2 ]; then
+  printf "How many days the Certificate will be valid:"
+  read SSL_EXPIRATION_DAYS
```
MarkRiddoch: As above re interactive use

scripts/core/selfssl
```diff
@@ -0,0 +1,89 @@
...
+fi
+
+echo "Creating a new Certificate ..."
+
+# Add more info /C=$country/ST=$state/L=$locality/O=$organization/OU=$organizational_unit/CN=$common_name/emailAddress=$email
+SUBJ="/C=$country/ST=$state/O=$organization/CN=$common_name/emailAddress=$email"
+
+openssl genrsa -des3 -passout pass:x -out server.pass.key 2048
```
MarkRiddoch: This will add another prerequisite, @izoratti should be aware of this for the documentation and also to look at hw it is integrated into the install procedure.

scripts/core/selfssl
```diff
@@ -0,0 +1,89 @@
...
+openssl req -new -key $SSL_NAME.key -subj $SUBJ -out $SSL_NAME.csr
+openssl x509 -req -sha256 -days $SSL_EXPIRATION_DAYS -in $SSL_NAME.csr -signkey $SSL_NAME.key -out $SSL_NAME.cert
+
+
+# The server.cert file is the certificate suitable for use along with the server.key private key.
+# Put these in $FOGLAMP_DATA/etc/certs, $FOGLAMP_ROOT/etc/certs or /usr/local/foglamp/etc/certs
+# Make directory to place SSL Certificate if it doesn't exists
+
+#if [[ ! -d $SSL_DIR ]]; then
+#  sudo mkdir -p $SSL_DIR
```
MarkRiddoch: We don't assume foglamp has sudo rights

praveen-garg:
@MarkRiddoch few questions questions:

- certificate name for Admin REST will be configurable? of fixed name? what should be default name and expiration_days should be used for this script when plugged to start

- it should always create certificate or only when enable_http is false and start/ rebooted (i.e. then script is called only from python code)

- if for any reason it failed to create/ load certs; admin rest api should be failed to start (with logs) or just log and fallback to http (but configuration may show false state unless its verified with log)

- Any minimum openssl version thing check is needed? as of today this script is checking for openssl existence only

praveen-garg:
@amarendra-dianomic there are 2 options 1. use passphrase (unset on error/ exit) 2. remove use of -des3

scripts/core/selfssl
izoratti: Yes, it makes sense.

scripts/core/selfssl
```diff
@@ -0,0 +1,125 @@
...
+    fi
+fi
+
+if [[ ! -d "${SSL_LOCATION}" ]]; then
+    # write to log?
+    echo "Could not find the directory ${SSL_LOCATION} to put SSL certificates."
+    exit 1
+fi
+
+# include logger script
```
izoratti: This should go up and capture the echo. (see general comment).

praveen-garg:
@MarkRiddoch @izoratti feedback changes are in. Please let me know the integration plan of it (as per above questions)

praveen-garg:
@amarendra-dianomic verify please

amarendra-dianomic:
@praveen-garg please note that,
- the earlier reported issue of '-des3' option now stands resolved. Now the certificates are being generated OK.
- foglamp server, rest server are all working fine with both http(port 8081) and https(port 1995) when the server is started vide "scripts/foglamp start" command.
- however, I am not able to start the server vide "python3 -m foglamp.services.core" with https enabled as in this case, logs indicate that server is not able to locate the certficate. Therefore, it will be better that in such cases, the server should default back to http and resume normally.

praveen-garg:
@amarendra-dianomic please note above the open questions (may be we want them to handle next)
also script make a call at deployment time; for python3 -m we may use a admin rest api call to create (code snippet posted in ticket FOGL-780)

PR #435, opened 2018-01-05
FOGL-904 microservicement management client

ashwinscale:
created reusable class for a microservice_management_client 
process.py now uses this class to connect to core


praveen-garg:
What is the plan for routes?
shutdown and unregister ( and start vs register)

PR #438, opened 2018-01-05
FOGL-842 MicroserviceClient code for register / unregister config interest

praveen-garg:
TODO: 

- [x] It fixes FOGL-915

- [x]  microservice test (see foo.py in tests) and exception handling fix

- [x] implement interest stuff

- [ ] test_service_deregister_on_fail.py FOGL-922


python/foglamp/common/service_record.py
```diff
@@ -39,10 +38,10 @@ def __init__(self, s_id, s_name, s_type, s_protocol, s_address, s_port, m_port):
         self._protocol = s_protocol
         self._address = s_address
         self._port = None
-        if s_port != None:
+        if s_port is not None:
             self._port = int(s_port)
         self._management_port = int(m_port)
-        self._status = 0
+        self._status = 1
```
praveen-garg: I believe, default status should be 1 (when service_record is created) if no ping let monitor loop set it to 0 and unregister

tests/unit-tests/python/foglamp_test/common/test_service_deregister_on_fail.py
```diff
@@ -19,12 +21,12 @@
...
-core_port = "37061"
+core_port = "36793"
 
 
-@pytest.allure.feature("common")
-@pytest.allure.story("process")
-class TestMicroservice:
+@pytest.allure.feature("core")
+@pytest.allure.story("monitor")
+class TestMonitoring:
```
praveen-garg: quick fix; FOGL-922 will handle it better; also need to see why ping does not work for `foo.py`

praveen-garg:
@MarkRiddoch @ashish-jabble  ready for review and merge. /cc @amarendra-dianomic

PR #452, opened 2018-01-10
FOGL-800 : Sensor payload modification filter

Singhal-Vaibhav:
Integration test results:
```
===================================================== test session starts ======================================================
platform linux -- Python 3.5.2, pytest-3.1.1, py-1.5.2, pluggy-0.4.0 -- /usr/bin/python3
cachedir: .cache
rootdir: /home/foglamp/Development/FogLAMP/tests/unit-tests/python/foglamp_test/common, inifile:
plugins: mock-1.6.0, cov-2.5.1, asyncio-0.6.0, allure-adaptor-1.7.7
collected 4 items 

test_jq_filter.py::TestJQFilter::test_default_filter_configuration PASSED
test_jq_filter.py::TestJQFilter::test_default_filterRule PASSED
test_jq_filter.py::TestJQFilter::test_custom_filter_configuration PASSED
test_jq_filter.py::TestJQFilter::test_invalid_filter_configuration PASSED

=================================================== 4 passed in 0.19 seconds ===================================================
```
Integration test will help to write a north side plugin using JQ filter


README.rst
```diff
@@ -61,6 +62,7 @@ On Ubuntu-based Linux distributions the packages can be installed with *apt-get*
    apt-get install python3-pip
    apt-get install python-dbus
    apt-get install postgresql
+   apt-get install autoconf automake build-essential libtool python-dev
```
MarkRiddoch: I take it these are needed to build the pyjq library and not for runtime? This makes.a difference for snap production.

Singhal-Vaibhav: Right,not required for runtime. I will get rid of them

Singhal-Vaibhav:
@MarkRiddoch ,
If i understood properly, FOGL-800 requirement was to provide a mechanism to use the JQ filter. This pull does that and it also demonstrate HOW with the supporting test.
If we want to integrate this filter with north plugins (every north plugin will have its own filter configuration), we can do it, however not in this pull. That should be a separate task for every north plugin.

izoratti:
I am ok with the code. I think the long commented def main should be removed and:
1. Add a comment to refer to a test to see an example
2. Add a reference of what the class does in README. 

I think once we have added this we are good to go.

README.rst
MarkRiddoch: We need to know about build requirements as well, so this is good to have. I will speak to Ivan about how to deal with requirements like this.

python/foglamp/common/jqfilter.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+# Other Transformation examples
+# No transformation - "."
+# First element of the JSON array - ".[0]"
+# Get only certain field from JSON (e.g., descr) - ".[]|{descr}"
+# Change the name of a JSON key (e.g., descr to Description - ".[]|{Description: .descr}"
+
+
+# def main():
```
MarkRiddoch: Are we better putting this in a Unit test - I have a personal dislike of commented out code!

PR #453, opened 2018-01-10
FOGL-916 Add configuration for advertising the FogLAMP services

MarkRiddoch:


python/foglamp/services/core/server.py
```diff
@@ -60,6 +60,34 @@ class Server:
...
+            'default': 'FogLAMP'
+        },
+        'description': {
+            'description': 'The description of this FogLAMP service',
+            'type': 'string',
+            'default': 'The FogLAMP administrative API'
+        }
+    }
+
+    _MANAGEMENT_SERVICE = '_foglamp_manage._tcp'
```
praveen-garg: please remove `_` between foglamp and manage `'_foglampmanage._tcp'`

python/foglamp/services/core/server.py
```diff
@@ -60,6 +60,34 @@ class Server:
...
+    """ The management service we advertise """
+
+    _ADMIN_API_SERVICE = '_foglamp_admin._tcp'
```
praveen-garg: please remove _ between foglamp and admin

python/foglamp/services/core/server.py
```diff
@@ -60,6 +60,34 @@ class Server:
...
+    """ The admin REST service we advertise """
+
+    _USER_API_SERVICE = '_foglamp_user._tcp'
```
praveen-garg: please remove _ between foglamp and user 

when these admin + user api is a single service; then why are we announcing twice; only diff will be authentication (sooner) + role (in future)

python/foglamp/services/core/server.py
```diff
@@ -213,6 +241,33 @@ def get_certificates():
...
+                _logger.error("No configuration manager available")
+            await cls._configuration_manager.create_category(category, config, 'The FogLAMP service configuration', True)
+            config = await cls._configuration_manager.get_category_all_items(category)
+
+            try:
+                cls._service_name = config['name']['value']
+            except KeyError:
+                cls._service_name = 'FogLAMP'
+            try:
+                cls._service_descriptiopn = config['description']['value']
```
praveen-garg: wrong var name

python/foglamp/services/core/server.py
```diff
@@ -340,15 +391,23 @@ def _start_core(cls, loop=None):
                 _logger.info('Loading certificates %s and key %s', cert, key)
                 ssl_ctx.load_cert_chain(cert, key)
 
+            # Get the service data and advertise the management port of the core to allow other microservices to find FogLAMP
+            loop.run_until_complete(cls.service_config())
+            _logger.info('Announce management API service')
+            cls.management_announcer = ServiceAnnouncer('core.{}'.format(cls._service_name), cls._MANAGEMENT_SERVICE, cls.core_management_port,
```
praveen-garg: Major concern: so by default if we deployed 100 (unattended) devices, all will have same name and type; unless configuration changed and REBOOTED!

I would prefer to use MAC address Or a uuid (saving it as config item will keep it same on reboot too) | and admin can always change it as needed to see something per their choice.

python/foglamp/services/core/server.py
```diff
@@ -340,15 +391,23 @@ def _start_core(cls, loop=None):
...
             _logger.info('REST API Server started on %s://%s:%s', 'http' if cls.is_rest_server_http_enabled else 'https',
                          address, service_server_port)
 
-            cls.admin_announcer = ServiceAnnouncer('FogLAMP', '_foglamp._tcp', service_server_port, ['The FogLAMP Admin REST API'])
-            cls.user_announcer = ServiceAnnouncer('FogLAMP', '_foglamp_app._tcp', service_server_port,
-                                              ['The FogLAMP Application  REST API'])
+            # All services are up so now we can advertise the Admin and User REST API's
+            cls.admin_announcer = ServiceAnnouncer(cls._service_name, cls._ADMIN_API_SERVICE, service_server_port,
+                                              [cls._service_description])
+            cls.user_announcer = ServiceAnnouncer(cls._service_name, cls._USER_API_SERVICE, service_server_port,
```
praveen-garg: 1 or 2 announcement, as above

python/foglamp/services/core/server.py
MarkRiddoch: They have always been separate logical services. At the moment they are the same port but they may no the forever.

python/foglamp/services/core/server.py
MarkRiddoch: As above they are separate logical services.

python/foglamp/services/core/server.py
MarkRiddoch: No, this follows well known conventions.

praveen-garg: `_foglamp_manage` is standard instead of `_foglampmanage` you mean? I will look more then.

python/foglamp/services/core/server.py
izoratti: I honestly think the use of UUID would be very confusing. 
The choice of the MAC address is better, but then we should use the IP address, which makes the whole concept of auto-discoverability pretty pointless.

praveen-garg:
for description txt record, Better is to use  ['description= a short description text']

python/foglamp/services/core/server.py
praveen-garg: is mac address dynamic? the way IP is on DHCP network?

izoratti: Let's make it clear: there is no perfect solution here.
1. UUID: it changes every time you reboot FogLAMP: you should change name accordingly
2. MAC address: there are multiple mac addresses (a Raspberry PI already has 2!), you must arbitrary select one of them
3. IP address: slightly better than UUID, but also bad for the multiple interfaces.
4. Default name: you may have machines advertising the same name.

In all this, I think the easiest and simplest way of handling this for a sysadmin is #4. Scripts can be easily set up to handle it, and so we may do with GUIs. 

From a product perspective, let's close this PR with #4 please.

izoratti:
@MarkRiddoch and @praveen-garg - see my comments above, I think the PR makes sense as is from a product perspective. The change of the names to "-" is also consistent with other services, so all good.

Could you please re-review and approve, so we close it?

python/foglamp/services/core/server.py
praveen-garg: Okay! I believe we could have picked any of #2 (to advertise better) but as decided let's go ahead as decided.

MarkRiddoch:
Mistakenly pushed FOGL-937 changes on this branch. Dismissing the pull request to clean up the branches.

python/foglamp/services/core/server.py
praveen-garg: For 4) we may, but that will not give expected results 
Please try `avahi-browse -d local _foglamp-manage._tcp --resolve -t`

Also on win/mac, we may want to use bonjour. hence proper txtrecord is good to have.

izoratti: I was thinking at a more simple solution, i.e. scripting for foglamp and the suffix added by avahi automatically. I honestly think we are good with this.

PR #454, opened 2018-01-10
Fogl 934 - it addresses the problem of having OMF translator handle numbers in string format

sstefano69:
it addresses the problem of having OMF translator handle numbers in string format


Singhal-Vaibhav:
This caused FOGL-960

Change in any code in sending process should be tested with all north plugins. I am not sure whether this was the case with this pull.

@billehunt, @MarkRiddoch , @izoratti As we discussed can we have the review process more transparent where reviewers post their comments on what was reviewed in the pull? Architecture/code/functionality, etc?

PR #455, opened 2018-01-11
Fogl 933 - Fix disabling South service schedule in Scheduler

amarendra-dianomic:


python/foglamp/services/core/scheduler/scheduler.py
```diff
@@ -1078,6 +1077,50 @@ def _schedule_row_to_schedule(cls,
             self._schedule_first_task(schedule_row, now)
             self._resume_check_schedules()
 
+    async def ping_service(self, service):
```
MarkRiddoch: Don't we already have a way to ping a service?

amarendra-dianomic: This is actually makes a call to the ping method of the individual service, via the defined route, which is part of the microservice instance.

praveen-garg: @amarendra-dianomic if its required, can it be a utils def; so that monitoring loop can also use it

python/foglamp/services/core/scheduler/scheduler.py
```diff
@@ -1078,6 +1077,50 @@ def _schedule_row_to_schedule(cls,
...
+                    attempt_count += 1
+                    time.sleep(1)
+            if attempt_count <= _MAX_ATTEMPTS:
+                self._logger.info('Ping received for Service %s id %s at url %s', service._name, service._id, url_ping)
+                return True
+        self._logger.error('Ping not received for Service %s id %s at url %s attempt_count %s', service._name, service._id,
+                      url_ping, attempt_count)
+        return False
+
+    async def shutdown_service(self, service):
```
praveen-garg: again, utils? I believe each scheduled (microservice) plugin shall need it. 
And/Or core server shutdown (just to ensure all microservices are shutdown except storage and core) as in Pr for FOGL-653

python/foglamp/services/core/scheduler/scheduler.py
```diff
@@ -1078,6 +1077,50 @@ def _schedule_row_to_schedule(cls,
...
+        _ping_timeout = 5  # type: int
+        """Timeout for a response from any given micro-service"""
+
+        try:
+            url_shutdown = "{}://{}:{}/foglamp/service/shutdown".format(service._protocol, service._address,
+                                                                        service._management_port)
+            async with aiohttp.ClientSession() as session:
+                async with session.post(url_shutdown) as resp:
+                    status_code = resp.status
+                    assert 200 == resp.status
```
praveen-garg: we should avoid assert practice in production code, you can keep else block with 200 check and return false and log error as in generic exception block

praveen-garg:
does this PR handles other than CC2650 sensor tag plugins schedule as well?

python/foglamp/services/core/scheduler/scheduler.py
amarendra-dianomic: Assertions are supposed to pass 100% of the time. An exception is only thrown when the condition unexpectedly fails. There's nothing wrong with running assertions in production code. They're not harmful, or bad design. They're good design, in fact. They confirm that the system is running correctly.

amarendra-dianomic:
@praveen-garg changes done per feedback. Pl check. Also, this PR does take care of **all startup** schedules enabling/disabling tasks.

MarkRiddoch:
I suspect the issue Vaibhav has found is a bug in the shutdown of the sensor tag poll plugin rather than in the change itself. I would propose we test the plugin and if we confirm it does not shutdown correctly we tackle the problem as a bug in that plugin and allow this change to move forward rather than delay this change because of a bug in the plugin.

amarendra-dianomic:
@MarkRiddoch I have checked all three South plugins and have found that while async mode plugins (COAP and CC2650ASYN) terminate without any trouble, poll mode plugin (CC2650POLL) need explicit kill signal to terminate. **This is because of the while True loop that is used to continuously poll the data from the device.** Therefore, I have now modified the shutdown method a bit and now using kill signal to terminate the poll plugin and which works fine for async mode plugins shutdown as well.

amarendra-dianomic:
@praveen-garg changes made as discussed. Pl check.

amarendra-dianomic:
Will be taken up later after the known issues as reported during testing are resolved first.

python/foglamp/plugins/south/cc2650async/cc2650async.py
```diff
@@ -165,148 +162,159 @@ def plugin_start(handle):
                 try:
                     pattern_index = tag.con.expect('Notification handle = .*? \r', timeout=4)
                 except pexpect.TIMEOUT:
-                    _LOGGER.error("SensorTagCC2650 async timeout")
-                    print("TIMEOUT exception!")
-                    break
-
-                # expect, if successful, will return the index of the pattern "Notification handle = " which ideally
+                    attempt_count += 1
+                    if attempt_count > 15:
```
praveen-garg: so max timeout is 15 seconds? for POLL it seems to be 10.

python/foglamp/plugins/south/cc2650async/cc2650async.py
```diff
@@ -165,148 +162,159 @@ def plugin_start(handle):
...
+                        _LOGGER.error("SensorTagCC2650 {} async timeout")
+                        print("TIMEOUT!!")
```
praveen-garg: if this this for python3 -m, can we please put the same message as in logger error

python/foglamp/plugins/south/cc2650async/cc2650async.py
```diff
@@ -165,148 +162,159 @@ def plugin_start(handle):
...
+                # will help in splitting at line#190.
+                if pattern_index != 0:
+                    attempt_count += 1
+                    if attempt_count > 15:
+                        _LOGGER.error("SensorTagCC2650 {} async timeout")
+                        print("TIMEOUT!!")
+                        break
+                    else:
+                        await asyncio.sleep(1)
+                        continue
```
praveen-garg: again? we have same block above

python/foglamp/plugins/south/cc2650async/cc2650async.py
```diff
@@ -165,148 +162,159 @@ def plugin_start(handle):
...
+
+                after = tag.con.after
+                hex_string = after.split()[3:]
+
+                cnt += 1
+                # Used only for debugging. debug_cnt should be set to 0 in production
+                if debug_cnt > 0:
+                    if cnt >= debug_cnt:
+                        break
+                    print(cnt, "****", hex_string)
```
praveen-garg: please remove

python/foglamp/plugins/south/cc2650poll/cc2650poll.py
```diff
@@ -243,12 +241,10 @@ def plugin_poll(handle):
 
     except (Exception, RuntimeError) as ex:
         _LOGGER.exception("SensorTagCC2650 {} exception: {}".format(bluetooth_adr, str(ex)))
-        raise exceptions.DataRetrievalError(ex)
```
praveen-garg: why we are not raising it? how this error will be handled then?

python/foglamp/services/common/utils.py
```diff
@@ -0,0 +1,77 @@
...
+    _ping_timeout = 15  # type: int
+    """Timeout for a response from any given micro-service"""
+
+    try:
+        url_shutdown = "{}://{}:{}/foglamp/service/shutdown".format(service._protocol, service._address,
+                                                                    service._management_port)
+        async with aiohttp.ClientSession() as session:
+            async with session.post(url_shutdown) as resp:
+                status_code = resp.status
+                assert 200 == resp.status
```
praveen-garg: just FYI assert does not work if python is optimized mode

python/foglamp/services/common/utils.py
```diff
@@ -0,0 +1,77 @@
...
+            _logger.info('Ping received for Service %s id %s at url %s', service._name, service._id, url_ping)
+            return True
+    _logger.error('Ping not received for Service %s id %s at url %s attempt_count %s', service._name, service._id,
+                       url_ping, attempt_count)
+    return False
+
+
+async def shutdown_service(service):
+    _ping_timeout = 15  # type: int
+    """Timeout for a response from any given micro-service"""
```
praveen-garg: is this a doc-string or explanation of _ping_timeout

python/foglamp/services/common/utils.py
```diff
@@ -0,0 +1,77 @@
...
+                status_code = resp.status
+                assert 200 == resp.status
+    except (Exception, AssertionError) as ex:
+        _logger.exception('Error in Service shutdown %s, %s', service._name, str(ex))
+        return False
+    else:
+        _logger.info('Service %s, id %s at url %s successfully shutdown', service._name, service._id, url_shutdown)
+        return True
+
+def terminate_child_processes(parent_id):
```
praveen-garg: why in Utils , Only scheduler shall use it.

python/foglamp/services/core/scheduler/scheduler.py
```diff
@@ -860,7 +866,16 @@ def _schedule_first_task(self, schedule, current_time):
...
         if self._task_processes:
-            raise TimeoutError("Timeout Error: Could not stop scheduler")
+            task_count = 0
+            for task_id in list(self._task_processes.keys()):
+                try:
+                    task_process = self._task_processes[task_id]
+                    schedule = task_process.schedule
+                    task_count += 1 if schedule.type != Schedule.Type.STARTUP else 0
+                except KeyError:
+                    continue
```
praveen-garg: what above block is doing; just counting?

python/foglamp/services/core/scheduler/scheduler.py
```diff
@@ -1092,24 +1107,23 @@ def _schedule_row_to_schedule(cls,
...
-            enabled=False,
-            process_name=schedule.process_name)
+        # Disable Schedule - update the schedule in memory
+        self._schedules[schedule_id] = self._schedules[schedule_id]._replace(enabled=False)
+
+        # Update database
+        update_payload = PayloadBuilder() \
+            .SET(enabled='f') \
+            .WHERE(['id', '=', str(schedule_id)]) \
+            .payload()
```
praveen-garg: better we make it 1 liner

python/foglamp/services/core/scheduler/scheduler.py
```diff
@@ -1092,24 +1107,23 @@ def _schedule_row_to_schedule(cls,
...
+        try:
+            self._logger.debug('Database command: %s', update_payload)
```
praveen-garg: Database > update

python/foglamp/services/core/scheduler/scheduler.py
```diff
@@ -1123,49 +1137,56 @@ def _schedule_row_to_schedule(cls,
...
-                task_future = task_process.future
-                if schedule.type != Schedule.Type.STARTUP and task_future.cancel() is True:
-                    await self._wait_for_task_completion(task_process)
+            if schedule.type == Schedule.Type.STARTUP: # If schedule is a service e.g. South services
+                try:
+                    found_services = ServiceRegistry.get(name=schedule.process_name)
+                    service = found_services[0]
+                    if await utils.ping_service(service) is True:
+                        # Shutdown will take care of unregistering the service from core
+                        await utils.shutdown_service(service)
```
praveen-garg: so now no need to call microservices shutdown in core/server.py? and scheduler stop shall handle it?

python/foglamp/services/core/scheduler/scheduler.py
```diff
@@ -1123,49 +1137,56 @@ def _schedule_row_to_schedule(cls,
...
-            try:
-                # KNOWN ISSUE: Does not work well with microservices e.g. COAP etc
-                # TODO: If schedule is a microservice, then deal with shutdown, unregister etc.
-                task_process.process.terminate()
-                task_future = task_process.future
-                if schedule.type != Schedule.Type.STARTUP and task_future.cancel() is True:
-                    await self._wait_for_task_completion(task_process)
+            if schedule.type == Schedule.Type.STARTUP: # If schedule is a service e.g. South services
+                try:
+                    found_services = ServiceRegistry.get(name=schedule.process_name)
```
praveen-garg: shall it always be `name=schedule.process_name`?

python/foglamp/services/core/scheduler/scheduler.py
```diff
@@ -1123,49 +1137,56 @@ def _schedule_row_to_schedule(cls,
...
+                del schedule_execution.task_processes[task_process.task_id]
+                del self._task_processes[task_process.task_id]
+            else: # else it is a Task e.g. North tasks
+                # Terminate process
+                try:
+                    # We need to terminate the child processes because now all tasks are started vide a script and
+                    # this creates two unix processes. Scheduler can store pid of the parent shell script process only
+                    # and on termination of the task, both the script shell process and actual task process need to
+                    # be stopped.
+                    utils.terminate_child_processes(task_process.process.pid)
```
praveen-garg: instead if utils, let's have it internal def of scheduler

python/foglamp/services/core/scheduler/scheduler.py
```diff
@@ -1234,10 +1224,11 @@ def _schedule_row_to_schedule(cls,
 
         # Start schedule
         await self.queue_task(schedule_id)
```
praveen-garg: not str(schedule_id) ?

python/foglamp/services/core/scheduler/scheduler.py
```diff
@@ -1234,10 +1224,11 @@ def _schedule_row_to_schedule(cls,
...
+
         self._logger.info(
             "Enabled Schedule '%s/%s' process '%s'\n",
             schedule.name,
-            schedule.schedule_id,
+            schedule_id,
```
praveen-garg: not str(schedule_id) ?

python/foglamp/services/core/server.py
```diff
@@ -520,7 +520,7 @@ def stop(cls, loop=None):
                 return
 
             tasks = [cls._request_microservice_shutdown(svc) for svc in services_to_stop]
-            await asyncio.wait(tasks)
+            await asyncio.gather(*tasks)
```
praveen-garg: what diff it will make?

python/foglamp/services/core/service_registry/monitor.py
```diff
@@ -46,7 +46,7 @@ def __init__(self):
     async def _monitor_loop(self):
         """Main loop for the scheduler"""
         # check health of all micro-services every N seconds
-        _MAX_ATTEMPTS = 5
+        _MAX_ATTEMPTS = 15
```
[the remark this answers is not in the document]
praveen-garg: acceptable? only POLL wants it that way.. what is optimal

python/foglamp/services/south/server.py
```diff
@@ -204,16 +204,18 @@ def run(self):
...
+        done, pending = await asyncio.wait(asyncio.Task.all_tasks(), timeout=5)
+        for task_pending in pending:
+            task_pending.cancel()
 
         # This deactivates event loop and
         # helps aiohttp microservice server instance in graceful shutdown
         _LOGGER.info('Stopping South Service Event Loop')
         loop.stop()
 
+        _LOGGER.info("Stopped plugin '{}'".format(self._name))
```
praveen-garg: Plugin shutdown logs should be (are) in its own module file; no need to duplicate again; please rebase and review

python/foglamp/services/south/server.py
```diff
@@ -233,8 +235,23 @@ def run(self):
...
         _LOGGER.info('Configuration has changed for South plugin {}'.format(self._name))
 
-        # plugin shutdown before re-configure
-        self._plugin.plugin_shutdown(self._plugin_handle)
+        # Shutdown plugin and Ingest before re-configure
+        try:
+            self._plugin.plugin_shutdown(self._plugin_handle)
+        except Exception:
+            _LOGGER.exception("Unable to stop plugin '{}' during reconfigure".format(self._name))
+            return web.json_response({"south": "reconfigure error - unable to stop plugin"})
```
praveen-garg: its an HTTPInternalError not 200

python/foglamp/services/south/server.py
```diff
@@ -233,8 +235,23 @@ def run(self):
...
+        try:
+            await Ingest.stop()
+            _LOGGER.info('Stopped the Ingest server.')
+        except Exception as ex:
+            _LOGGER.exception('Unable to stop the Ingest server. %s', str(ex))
+            return web.json_response({"south": "reconfigure error - unable to stop ingest"})
```
praveen-garg: its an HTTPInternalError not 200

python/foglamp/services/south/server.py
```diff
@@ -244,4 +261,10 @@ def run(self):
         new_handle = self._plugin.plugin_reconfigure(self._plugin_handle, new_config)
         self._plugin_handle = new_handle
 
+        # Executes the requested plugin type with new config
+        if self._plugin_info['mode'] == 'async':
+            await  self._exec_plugin_async(new_config)
+        elif self._plugin_info['mode'] == 'poll':
+            asyncio.ensure_future(self._exec_plugin_poll(new_config))
+
```
praveen-garg: I think restart is responsibility of reconfigure, how it works for CoAP/ HTTP south; if no other possibilty .. add a docstring with FIXME?

praveen-garg:
I think we are very close. :+1:

python/foglamp/services/core/scheduler/scheduler.py
praveen-garg: Okay Yes!

python/foglamp/services/core/scheduler/scheduler.py
amarendra-dianomic: added the doc string

praveen-garg:
Approving based on 3 known things:

- monitor loop _MAX_ATTEMPTS  increased to 15, earlier it was set to 5.
- plugin reconfigure, does not call start of plugin but change endpoint itself after getting new handle from plugin_reconfigure
- scheduler has 1 more instance usage of `FOGLAMP_ENV = TEST` now for tasks. (earlier it had 1 only.

python/foglamp/services/south/server.py
```diff
@@ -233,8 +234,23 @@ def run(self):
...
-        # plugin shutdown before re-configure
-        self._plugin.plugin_shutdown(self._plugin_handle)
+        # Shutdown plugin and Ingest before re-configure
+        try:
+            self._plugin.plugin_shutdown(self._plugin_handle)
+        except Exception:
+            _LOGGER.exception("Unable to stop plugin '{}' during reconfigure".format(self._name))
+            raise web.HTTPInternalServerError(reason="reconfigure error - unable to stop plugin {}".format(self._name))
+        try:
+            await Ingest.stop()
```
praveen-garg: we should not need that if (re)start is responsibility of reconfigure itself

python/foglamp/services/south/server.py
```diff
@@ -233,8 +234,23 @@ def run(self):
...
+            _LOGGER.info('Stopped the Ingest server.')
+        except Exception as ex:
+            _LOGGER.exception('Unable to stop the Ingest server. %s', str(ex))
+            raise web.HTTPInternalServerError(reason="reconfigure error - unable to stop ingest for plugin {}".format(self._name))
+
+        # Cancel all pending asyncio tasks after a timeout occurs
+        done, pending = await asyncio.wait(asyncio.Task.all_tasks(), timeout=5)
```
praveen-garg: better could have been, plugin collect its own tasks and cancel pending on it shutdown. **It looks like, here we have just simulated the south service stop start (except its aiohttp instance), not the child plugin of it**

amarendra-dianomic:
@MarkRiddoch with reference to review feedback given  by @praveen-garg and @Singhal-Vaibhav I would like to mention that,
- Though monitor loop _MAX_ATTEMPTS increased to 15, earlier it was set to 5 but currently 15 seems to a practical choice given the variance in the runtime environment. So let us stick to 15 for now.
- Restarting a STARTUP schedule is not working properly and I have already created an issue [FOGL-1004](https://scaledb.atlassian.net/browse/FOGL-1004) for the same. This involves rather heavy fixing and testing and as such it will be taken up separately.
- Both Praveen and yourself pointed out use of environment variable in scheduler.py. I have added TODO for both these occurrences and these will be removed when we fully implemented mocking in tests. This will also be taken up separately.
- I have noted Praveen's concerns regarding reconfigure feature for South plugins. I will also address these separately vide [FOGL-989](https://scaledb.atlassian.net/browse/FOGL-1004). I would also like to mention that via repo a temp repo FOGL-933.b, I have successfully tested reconfigure of all South plugins but this has been kept separately in view of FOGL-989. You commented on the ticket that plugins would not restart on reconfigure. Can you please elaborate this, preferably on FOGL-989? Because, currently, we are restarting the plugins in the same process (meaning that no new process is created upon restart) on reconfigure and it has been tested quite successfully. I hope this is acceptable.

PR #457, opened 2018-01-11
FOGL-924 - cc2650poll plugin fails with error  [DO NOT MERGE]

amarendra-dianomic:


python/foglamp/plugins/south/common/sensortag_cc2650.py
```diff
@@ -136,24 +136,31 @@ class SensorTagCC2650(object):
...
+        while attempt_count <= 5:
+            try:
+                self.bluetooth_adr = bluetooth_adr
+
+                self.con = pexpect.spawn('gatttool -b ' + bluetooth_adr + ' --interactive')
+                self.con.expect('\[LE\]>', timeout=int(timeout))
+
+                msg_debug = 'SensorTagCC2650 {} Connecting... If nothing happens, please press the power button. Attempt no. {}'.\
+                            format(self.bluetooth_adr, attempt_count)
+                print(msg_debug)
```
MarkRiddoch: Can we loose the prints please?

python/foglamp/plugins/south/common/sensortag_cc2650.py
```diff
@@ -136,24 +136,31 @@ class SensorTagCC2650(object):
...
+                self.con.sendline('connect')
+                self.con.expect('.*Connection successful.*\[LE\]>', timeout=int(timeout))
+                self.is_connected = True
+                msg_success = 'SensorTagCC2650 {} connected successfully'.format(self.bluetooth_adr)
+                print(msg_success)
+                _LOGGER.debug(msg_success)
+                break
+            except (ExceptionPexpect, EOF, TIMEOUT, Exception) as ex:
+                attempt_count += 1
+                time.sleep(1)
```
MarkRiddoch: I would be tempted to backoff the sleeps here - the problem of course is that the init method will block and eventually the monitor will kick in and declare the service as dead. Not sure we have a good solution for this in the poll mode. It may be we have to say the poll mode can not work with gatttool like this.

python/foglamp/plugins/south/common/sensortag_cc2650.py
amarendra-dianomic: This message appears only when we start server vide "python3 -m foglamp.services.core".

python/foglamp/plugins/south/common/sensortag_cc2650.py
amarendra-dianomic: Reduced the sleep time from 1 second to 0.5 second. Hope this is acceptable. Secondly, the alternate to gatttool is hcitool. May be we should experiment with hcitool in next sprint.

amarendra-dianomic:
@ashwinscale changes made to address the concerns raised. Please check.

python/foglamp/services/core/service_registry/monitor.py
```diff
@@ -46,7 +46,7 @@ def __init__(self):
     async def _monitor_loop(self):
         """Main loop for the scheduler"""
         # check health of all micro-services every N seconds
-        _MAX_ATTEMPTS = 5
+        _MAX_ATTEMPTS = 15
```
praveen-garg: I believe its just too high 👎

amarendra-dianomic: it is warranted to take care of network latency. I have experienced ping at 9th attempt even, hence this value to assure that service has really shutdown.

amarendra-dianomic:
Will be taken up later after the known issues as reported during testing of PR#455, for FOGL-933, are resolved first.

PR #458, opened 2018-01-11
FOGL 916, 937 rolled together

MarkRiddoch:


python/foglamp/services/core/server.py
```diff
@@ -213,6 +241,33 @@ def get_certificates():
...
+        """
+        Get the service level configuration
+        """
+        try:
+            config = cls._SERVICE_DEFAULT_CONFIG
+            category = 'service'
+
+            if cls._configuration_manager is None:
+                _logger.error("No configuration manager available")
+            await cls._configuration_manager.create_category(category, config, 'The FogLAMP service configuration', True)
```
praveen-garg: +            if cls._configuration_manager is None:  then we should not proceed? else it will give error NoneType has no ...

PR #459, opened 2018-01-11
FOGL-939, FOGL-940: Add JQ filter configuration on north plugins and update sending process

Singhal-Vaibhav:
~~Looked from http receiving side:~~
~~Earlier response (where we are manipulating each read tuple in code):~~
```
[{'asset_code': 'fogbench/pressure', 'readings': [{'reading': {'pressure': 901.3}, 'user_ts': '2018-01-11 14:13:18.157493+05:30', 'read_key': 'c452a4e1-c9ed-4023-92ee-ca37d0c354dc'}]}, {'asset_code': 'fogbench/humidity', 'readings': [{'reading': {'temperature': 20, 'humidity': 14}, 'user_ts': '2018-01-11 14:13:18.157535+05:30', 'read_key': 'd0f364e4-f43b-45cd-8553-01d0715c8a27'}]}]
```

~~Using JQ Filter:~~
```
[{'asset_code': 'fogbench/pressure', 'readings': [{'reading': {'pressure': 901.3}, 'user_ts': '2018-01-11 14:13:18.157493+05:30', 'read_key': 'c452a4e1-c9ed-4023-92ee-ca37d0c354dc'}]}, {'asset_code': 'fogbench/humidity', 'readings': [{'reading': {'temperature': 20, 'humidity': 14}, 'user_ts': '2018-01-11 14:13:18.157535+05:30', 'read_key': 'd0f364e4-f43b-45cd-8553-01d0715c8a27'}]}]
```

~~No change in output~~

As per the latest discussion, the scope of using jq filter is manipulating in data only. Plugins assume that they get the right set of keys and payloads keys are not altered. Default jq configuration will be False

### DIFF
--- a/python/foglamp/plugins/north/http_north/http_north.py
+++ b/python/foglamp/plugins/north/http_north/http_north.py
@@ -12,7 +12,6 @@ import json
 
 from foglamp.common import logger
 from foglamp.plugins.north.common.common import *
-from foglamp.common.jqfilter import JQFilter
 
 __author__ = "Ashish Jabble"
 __copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
@@ -48,12 +47,12 @@ _DEFAULT_CONFIG = {
     "applyFilter": {
         "description": "Whether to apply filter before processing the data",
         "type": "boolean",
-        "default": "True"
+        "default": "False"
     },
     "filterRule": {
         "description": "JQ formatted filter to apply (applicable if applyFilter is True)",
         "type": "string",
-        "default": ".[] | {readings: [{reading: .reading,user_ts: .user_ts,read_key: .read_key}], asset_code: .asset_code}"
+        "default": ".[]"
     }
 }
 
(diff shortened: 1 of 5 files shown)


python/foglamp/plugins/north/http_north/http_north.py
```diff
@@ -43,6 +44,16 @@
         'description': 'how long (x seconds) the plugin should wait for pending tasks to complete or cancel otherwise',
         'type': 'integer',
         'default': '10'
+    },
+    "applyFilter": {
+        "description": "Whether to apply filter before processing the data",
+        "type": "boolean",
+        "default": "True"
```
MarkRiddoch: Before we check in I think we should make the filter disabled by default

praveen-garg:
a quick question: if we only change filter rule from config; shall not mapping for send block shall break?

Singhal-Vaibhav:
@praveen-garg Yes and No :) , 
If we set applyFilter to true and change the filterRule for any plugin, it depends on what the filter rule is. If the filter rule manipulates the keys of the payload dictionary the send block of plugins will break as currently both the omf and http_north are manipulating the reading block based on the payload keys. 
On the other hand, we use filter to manipulate the data of the payload (without touching the keys), then it will not break. e.g's can be adding a unit to the data, and so on.

python/foglamp/plugins/north/http_north/http_north.py
izoratti: I agree, the filter must be disabled.

PR #464, opened 2018-01-15
fix aiohttp dependency

praveen-garg:
until https://github.com/aio-libs/aiohttp/issues/2662 is fixed.

We need this in 1.1 and master too


pintomax:
Temporary fix for now.

praveen-garg:
for 2.3.6 released version yarl 0.18.0 (already released as well) is good; that way we can ensure it always using that tested compatible version.

PR #465, opened 2018-01-15
FOG-949: force ORDER BY DESC to asset/series REST API call

pintomax:
The query related to asset/value/series now returns data order by timestamp desc:

>$ curl -s 'http://localhost:8081/foglamp/asset/fogbench%2Fpressure/pressure/series' | jq
```json
[
  {
    "min": "804",
    "average": "949.84",
    "timestamp": "2018-01-15 11:10:43",
    "max": "1086.5"
  },
  {
    "min": "821.3",
    "average": "911.63",
    "timestamp": "2018-01-15 11:10:40",
    "max": "1088.6"
  },
  {
    "min": "812",
    "average": "886.97",
    "timestamp": "2018-01-15 11:10:34",
    "max": "1021.9"
  },
  {
    "min": "823.4",
    "average": "958.06",
    "timestamp": "2018-01-15 11:10:29",
    "max": "1094.1"
  }
```

### DIFF
--- a/python/foglamp/services/core/api/browser.py
+++ b/python/foglamp/services/core/api/browser.py
@@ -343,7 +343,7 @@ async def asset_averages(request):
     _limit_skip_payload = validate_limit_skip(request, d)
     d.update(_limit_skip_payload)
 
-    # FOGL-949: add ORDER BY timestamp DESC
+    # Add ORDER BY timestamp DESC
     _sort_payload = PayloadBuilder(_limit_skip_payload).ORDER_BY(["timestamp", "desc"]).chain_payload()
     d.update(_sort_payload)
 
--- a/python/foglamp/services/core/api/browser.py
+++ b/python/foglamp/services/core/api/browser.py
@@ -294,8 +294,18 @@ async def asset_averages(request):
     The amount of time covered by each returned value is set using the
     query parameter group. This may be set to seconds, minutes or hours
 
-    Return the result of the query
-    SELECT user_ts AVG((reading->>'reading')::float) FROM readings WHERE asset_code = 'asset_code' GROUP BY user_ts
+    Return the result of the query:
+
+    SELECT min((reading->>'reading')::float) AS "min",
+           max((reading->>'reading')::float) AS "max",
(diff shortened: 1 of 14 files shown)


python/foglamp/services/core/api/browser.py
```diff
@@ -294,8 +294,18 @@ def validate_limit_skip(request, _dict):
...
-    SELECT user_ts AVG((reading->>'reading')::float) FROM readings WHERE asset_code = 'asset_code' GROUP BY user_ts
+    Return the result of the query:
+
+    SELECT min((reading->>'reading')::float) AS "min",
+           max((reading->>'reading')::float) AS "max",
+           avg((reading->>'reading')::float) AS "average",
+           to_char(user_ts, 'YYYY-MM-DD HH24:MI:SS') AS "timestamp"
+    FROM foglamp.readings
+           WHERE asset_code = 'asset_code' AND
+             reading ? 'reading'
```
ashish-jabble: If `AND` required then we should add into payload itself otherwise remove it from here.

pintomax: This is just the example query, taken from SQL seen by the storage

MarkRiddoch: The AND reading ? 'reading' is internal top the storage layer to check that the JSON property exists and should not be in the payload passed to the storage layer.

python/foglamp/services/core/api/browser.py
```diff
@@ -333,6 +343,10 @@ def validate_limit_skip(request, _dict):
     _limit_skip_payload = validate_limit_skip(request, d)
     d.update(_limit_skip_payload)
 
+    # Add ORDER BY timestamp DESC
+    _sort_payload = PayloadBuilder(_limit_skip_payload).ORDER_BY(["timestamp", "desc"]).chain_payload()
+    d.update(_sort_payload)
+
     payload = json.dumps(d)
```
ashish-jabble: @pintomax FYI, Instead of above code snippet we could have been use `payload = PayloadBuilder(d).ORDER_BY(["timestamp", "desc"]).payload()`

PR #473, opened 2018-01-17
FOGL-547 Fix returned payload for asset summary

MarkRiddoch:


MarkRiddoch:
Ashish  has also submitted some fix. So closing this.

PR #477, opened 2018-01-18
FOGL-959: Documentation on FOGLAMP scripted tests

Singhal-Vaibhav:


tests/README.rst
```diff
@@ -0,0 +1,68 @@
...
+
+FogLAMP uses pytest as the test runner for testing python based code. For more information on pytest please refer
+|pytest docs|
+To execute a single test, navigate to the directory where your test code is placed and execute the command
+``pytest test_filename.py`` where test_filename.py is the name of the file which contains your tests.
+
+pytest also supports running of a complete test suite. To execute the complete test suite, navigate to the directory
+which contains the tests and run the command ``pytest`` . This will execute all the python tests in the given directory
+and sub-directories.
+
```
praveen-garg: should we list, how to run a specific (single) test or specific class in module?

tests/README.rst
```diff
@@ -0,0 +1,68 @@
...
+- `Integration`_ - Tests that checks the integration of different FogLAMP units that work as a single component.
+- `System`_ - Tests that checks the end to end flows in FogLAMP
+
+
+Running FogLAMP scripted tests
+==============================
+
+Test Prerequisites
+------------------
+
```
MarkRiddoch: I believe there are more prerequisites than this - for example the tests seem to need a system running before the tests are run. Do we not need to set PYTHONPATH before running the tests?

We may consider having a script to bundle up all the things we need to do to run test in the same way as we do with scripts/foglamp start

Singhal-Vaibhav:
@MarkRiddoch & @ashwinscale ,

While I agree to most of you points but do we want to fix the issues with testing first and then document or document first and then FIX testing infra?

MarkRiddoch:
I would like to have instructions on what we have now first. These instructions should allow somebody who has never seen FogLAMP before to clone it from GitHub and have the tests run in the same way the instructions for installing and running FogLAMP aim to. So we need to have all the steps documented to go from a freshly cloned GitHub repository to test execution.

Once we have that I think we need to improve the testing mechanism and add more tests, but should always make sure that documentation tracks the changes.

If we can simplify the setup by making small changes now, then that is probably worth it, but we should not make changes that will delay the roll out of testing instructions that all can follow. We should aim too have this is place by the end of today ideally.

Singhal-Vaibhav:
@ashwinscale , @MarkRiddoch 
Feedback comments incorporated.

tests/unit/README.rst
```diff
@@ -0,0 +1,48 @@
+******************
+FogLAMP Unit Tests
+******************
+
+Unit tests are the first category of test in FogLAMP. These test ensures that every single function of your code under
+test is generating the desired output.
+
+FogLAMP unit tests heavily use test doubles to replace a production object. A typical example is a code fragment that
```
ashwinscale: This file itself is in the unit parent directory yet it holds details on python unit tests (mocking in particular). Python specific information should be moved into the python subdirectory

tests/integration/README.rst
```diff
@@ -0,0 +1,26 @@
...
+Since these kinds of tests interacts between two or more heterogeneous systems, these are often slow in nature.
+
+**NOTE:** *It is necessary to run FogLAMP for integration tests to work*
+
+Currently integration tests can be executed only once at a time, going forward it will be possible to run integration
+tests as a suite. To run any integration test, you need to replace the _core_management_port in the code. The core
+management port is exposed by the FogLAMP Core service.
+
+core management port can be found out from /var/log/syslog
+::
```
ashwinscale: Please outline step by step instructions here. 
Something such as:
$FOGLAMP_ROOT/scripts/start
# check /var/log/syslog to get core_mng_port
# specify port in FogLAMP/tests/integration/foglamp/conftest.py
cd FogLAMP/tests/integration/
pytest

I tried these exact instructions with no success. Please fill in anything I am missing.

Singhal-Vaibhav: running pytest in a directory run all the python tests in that suite. For integration, the instructions are:
> Currently integration tests can be executed only once at a time, going forward it will be possible to run integration tests as a suite.

Step by step instructions added as requested.

tests/integration/foglamp/services/common/microservice_management/service_registry/test_services_registry_api.py
```diff
@@ -20,8 +20,7 @@
 # Module attributes
 __DB_NAME = "foglamp"
 # Needs foglamp to start,
-# replace 37935 with core_management_port
-BASE_URL = 'http://localhost:37935/foglamp'
+BASE_URL = 'http://localhost:'+str(pytest.test_env.core_mgmt_port)+'/foglamp'
```
praveen-garg: use .format or %s

tests/integration/foglamp/services/core/test_scheduler_get_tasks.py
```diff
@@ -56,9 +56,7 @@
                                        sqlalchemy.Column('reason', sqlalchemy.types.VARCHAR(255)))
 
 # TODO: To run this test,
-#       1) Do 'scripts/foglamp start' and note the management_port from syslog
-#       2) Change _m_port below with the management_port
-#       3) Execute this command: FOGLAMP_ENV=TEST pytest -s -vv tests/integration/foglamp/services/core/test_scheduler_get_tasks.py
+#       1) Execute this command: FOGLAMP_ENV=TEST pytest -s -vv tests/integration/foglamp/services/core/test_scheduler_get_tasks.py
```
praveen-garg: explictly say: its 'FOGLAMP_ENV=TEST' only used by scheduler

tests/integration/foglamp/services/core/test_scheduler.py
```diff
@@ -24,9 +24,7 @@
```
praveen-garg: this file should move to unit?

tests/unit/python/__template__.py
```diff
@@ -0,0 +1,44 @@
...
+    @pytest.fixture(scope="", params="", autouse=False, ids=None, name=None)
+    def _module_fixture(self):
+        """Test fixtures that is specific for this class. This fixture can be used with any test definition"""
+        pass
+
+    @pytest.mark.parametrize("test_input, expected", [
+        ("input_data1", "expected_data_1"),
+        ("input_data1", "expected_data_2")
+    ])
+    def test_some_unit(self, _module_fixture, test_input, expected):
```
praveen-garg: add an example for non-parameterized

tests/integration/foglamp/services/core/test_scheduler_get_tasks.py
```diff
@@ -56,9 +56,7 @@
                                        sqlalchemy.Column('reason', sqlalchemy.types.VARCHAR(255)))
```
praveen-garg: this file should be moved to unit?

tests/integration/foglamp/services/core/test_scheduler.py
```diff
@@ -24,9 +24,7 @@
 
 _CONNECTION_STRING = "dbname='foglamp' user='foglamp'"
 # TODO: To run this test,
-#       1) Do 'scripts/foglamp start' and note the management_port from syslog
-#       2) Change _m_port below with the management_port
-#       3) Execute this command: FOGLAMP_ENV=TEST pytest -s -vv tests/integration/foglamp/services/core/test_scheduler.py
+# 1) Execute this command: FOGLAMP_ENV=TEST pytest -s -vv tests/integration/foglamp/services/core/test_scheduler.py
```
praveen-garg: explictly say: its 'FOGLAMP_ENV=TEST' only used by scheduler

tests/README.rst
```diff
@@ -0,0 +1,89 @@
...
+Follow the instructions mentioned `here`_  to install and run FogLAMP on your machine.
+You can test FogLAMP from your development environment or after installing FogLAMP.
+
+In order to execute the scripted tests, run the following command from FOGLAMP_ROOT
+::
+   pip3 install -r python/requirements-test.txt --user
```
MarkRiddoch: This does not run any tests, it is getting the dependencies the tests need - please reword as it stand it makes it sound as if the above command executes tests.

tests/integration/foglamp/services/core/test_scheduler.py
Singhal-Vaibhav: This will be addressed when dependencies of running foglamp is removed and the test is a pure unit test.

tests/integration/foglamp/services/core/test_scheduler_get_tasks.py
Singhal-Vaibhav: This will be addressed when dependencies of running foglamp is removed and the test is a pure unit test.

PR #487, opened 2018-01-18
Fogl 935- Creates a North plugin for OCS

sstefano69:


sstefano69:
@MarkRiddoch, I would like to propose this : 
1) merge as is (so the OMF will not be impacted for now)
2) handle the timezone problem
3) create a new Jira task for handling OMS/OCS refactoring managed after 2

MarkRiddoch:
That sounds like a good plan

PR #490, opened 2018-01-18
FOGL-966 fix 

praveen-garg:
instead of creating certs (and temp key) and moving; create directly on the PATH decided with env var settings

> tip: use `sudo -H make install` to allow pip caching

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,6 @@ RM_FILE := rm
 MAKE_INSTALL = $(MAKE) install
 CP     := cp
 CP_DIR := cp -r
-RSYNC  := rsync
 
 ###############################################################################
 ################################### DIRS/FILES ################################
@@ -64,7 +63,7 @@ FOGLAMP_SCRIPT_SRC         := scripts/foglamp
 
 # SCRIPTS TO INSTALL IN SCRIPTS DIR
 COMMON_SCRIPTS_SRC         := scripts/common
-CORE_SCRIPTS_SRC           := scripts/core
+CORE_SCRIPTS_SRC           := scripts/core/selfssl
 POSTGRES_SCRIPT_SRC        := scripts/plugins/storage/postgres
 SOUTH_SCRIPT_SRC           := scripts/services/south
 STORAGE_SERVICE_SCRIPT_SRC := scripts/services/storage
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -182,7 +188,10 @@ $(SCRIPTS_INSTALL_DIR) :
 
 install_common_scripts : $(SCRIPT_COMMON_INSTALL_DIR) $(COMMON_SCRIPTS_SRC)
 	$(CP) $(COMMON_SCRIPTS_SRC)/*.sh $(SCRIPT_COMMON_INSTALL_DIR)
-	
+
+install_core_scripts : $(SCRIPT_CORE_INSTALL_DIR) $(CORE_SCRIPTS_SRC)
+	$(RSYNC) --exclude *.rst $(CORE_SCRIPTS_SRC)/* $(SCRIPT_CORE_INSTALL_DIR)
```
ashwinscale: Rsync shouldn't be needed. If we are excluding a file, it means the directory we are going to install may have artifacts of a foglamp run (I am guessing). Typically any generated files shouldn't exist in a scripts/source directory (as it is readonly).

Makefile
```diff
@@ -268,6 +286,7 @@ install_data : $(DATA_INSTALL_DIR) $(DATA_SRC_DIR)
 ifdef SUDO_USER
 ifeq ("$(USER)","root")
 	chown -R ${SUDO_USER}:${SUDO_USER} $(INSTALL_DIR)/$(DATA_SRC_DIR)
```
ashwinscale: Why don't we just run chown on the INSTALL_DIR. Then we wouldn't need multiple steps. @izoratti 
 is there a reason to specify the DATA_SRC_DIR rather than the general INSTALL_DIR?

Makefile
praveen-garg: > Why does this file appear within this directory?
you mean `.rst` file?

Makefile
```diff
@@ -182,7 +188,10 @@ $(SCRIPTS_INSTALL_DIR) :
 
 install_common_scripts : $(SCRIPT_COMMON_INSTALL_DIR) $(COMMON_SCRIPTS_SRC)
 	$(CP) $(COMMON_SCRIPTS_SRC)/*.sh $(SCRIPT_COMMON_INSTALL_DIR)
-	
+
+install_core_scripts : $(SCRIPT_CORE_INSTALL_DIR) $(CORE_SCRIPTS_SRC)
```
ashwinscale: Why not follow the example set by install_postgres_script  where we specify the single script we want to copy over? I think rsync is a valid method, but for makefile readability and consistency, I would prefer to use set patterns (install_postgres_script for example). If we switch to use rsync, I would prefer all similar targets to use rsync (for the sake of consistency), and it should be done in a separate issue/PR.

Makefile
```diff
@@ -61,6 +63,7 @@ FOGLAMP_SCRIPT_SRC         := scripts/foglamp
 
 # SCRIPTS TO INSTALL IN SCRIPTS DIR
 COMMON_SCRIPTS_SRC         := scripts/common
+CORE_SCRIPTS_SRC           := scripts/core/selfssl
```
ashwinscale: Please rename this to SELFSSL_SCRIPT_SRC as it refers to a very particular source file.

praveen-garg:
I was not planning to touch this file.

praveen-garg:
@izoratti @MarkRiddoch @pintomax Please review if any other functional or naming etc. issue?

PR #496, opened 2018-01-23
FOGL-970_Timezone_Handling

sstefano69:


python/foglamp/tasks/north/sending_process.py
```diff
@@ -493,12 +487,15 @@ def _transform_in_memory_data_readings(raw_data):
                     value = payload[key]
                     payload[key] = plugin_common.convert_to_type(value)
 
+                # Adds timezone UTC
+                timestamp = row['user_ts'] + "+00"
```
ashish-jabble: See the timestamp O/p `2018-01-23 19:29:20.081109+05:30+00`. Do we need this, please elaborate?

sstefano69:
The storage layer has changed a 'make' is required also.

python/foglamp/tasks/north/sending_process.py
sstefano69: the transformation code of the plug-in is expecting the UTC expressed (and I like to have a timezone expressed instead of having naive datetime) - plugin piece of code :

`step1 = datetime.datetime.strptime(timestamp_raw, '%Y-%m-%d %H:%M:%S.%f+00')

PR #504, opened 2018-01-30
FOGL-1012 core api ping test + HttpSouthIngest fixes

praveen-garg:
* Need to add readme (conftest etc. @ashish-jabble ) in another REST api test's PR
* http(s) SSL context based test to be done in next Pull request ( a part of completion of this JIRA)


tests/unit/python/foglamp/services/core/api/test_common_ping.py
```diff
@@ -0,0 +1,32 @@
...
+from aiohttp import web
+import json
+import pytest
+
+from foglamp.services.core import routes
+
+
+@pytest.allure.feature("unit")
+@pytest.allure.story("api", "common")
+async def test_ping(test_server, test_client, loop):
```
praveen-garg: Readme shall clarify about:

 > test_server/test_client /loop loop  are available from pytest-aiohttp plugin ( with conftest based content too)

tests/unit/python/foglamp/services/core/api/test_common_ping.py
```diff
@@ -0,0 +1,32 @@
...
+    app = web.Application()
+    # fill route table
+    routes.setup(app)
+
+    server = await test_server(app)
+    server.start_server(loop=loop)
```
praveen-garg: @ashwinscale atleast server should be tested once, at ping. Later tests should use this tested server only (no need to use test_server fixture)

praveen-garg:
> How do the parameters to test_ping() get used?

1. test_server to start and test server and using the loop for loop context
2. test_client to send a ping request

I'm not sure, I'm following the question?!

tests/unit/python/foglamp/services/core/api/test_common_ping.py
```diff
@@ -0,0 +1,32 @@
...
+
+    client = await test_client(app)
+
+    resp = await client.get('/foglamp/ping')
+    assert resp.status == 200
```
ashish-jabble: actual == expected?

tests/unit/python/foglamp/services/core/api/test_common_ping.py
```diff
@@ -0,0 +1,32 @@
...
+
+from foglamp.services.core import routes
+
+
+@pytest.allure.feature("unit")
+@pytest.allure.story("api", "common")
+async def test_ping(test_server, test_client, loop):
+    app = web.Application()
+    # fill route table
+    routes.setup(app)
```
ashish-jabble: Shouldn't we use only GET ping route here instead of all routes? I think better to load routes file-wise or common code

```
from foglamp.services.core.api import common as api_common
app.router.add_route('GET', '/foglamp/ping', api_common.ping())
```

tests/unit/python/foglamp/services/core/api/test_common_ping.py
ashwinscale: What do you mean here @ashish-jabble ?

tests/unit/python/foglamp/services/core/api/test_common_ping.py
ashwinscale: Why not use both TestServer and TestClient fixtures for all tests?

tests/unit/python/foglamp/services/core/api/test_common_ping.py
ashwinscale: I would agree here. Do you see any problems maintaining routes in two places (source and test)?

tests/unit/python/foglamp/plugins/south/test_http_south.py
```diff
@@ -62,17 +59,17 @@ class TestHttpSouthDeviceUnit(object):
                 }
             }
         }"""
-        with patch.object(Ingest, 'add_readings', return_value=asyncio.ensure_future(asyncio.sleep(0.1))) as mock_method1:
-            with patch.object(Ingest, 'is_available', return_value=True) as mock_method2:
-                request = mock_request(data)
+        with patch.object(Ingest, 'add_readings', return_value=asyncio.ensure_future(asyncio.sleep(0.1))):
```
ashwinscale: I don't understand the return value.  We would need to patch add_readings with an awaitable. Am I missing something here?

tests/unit/python/foglamp/services/core/api/test_common_ping.py
praveen-garg: server shall be started behind the scene even with test_client fixture itself. Here the purpose is to test that with default start and config based https (To be done in next PR, as I pointed in PR description) 

Hence test_server fixture is explicitly used. And we don't need to test that server start again  with diff config, as that unit will be tested here itself.

tests/unit/python/foglamp/services/core/api/test_common_ping.py
praveen-garg: Okay, I will change to `assert 200 == resp.status` to keep stuff consistent and as a practice.

tests/unit/python/foglamp/services/core/api/test_common_ping.py
praveen-garg: My opinion is, that will add an overhead of another test to that static def (routes setup) test And will be an extra baggage to import and maintain.

tests/unit/python/foglamp/services/core/api/test_common_ping.py
ashish-jabble: Here I meant for general test file practice (assert actual == expected)
`E.g. assert 200 = resp.status`

tests/unit/python/foglamp/plugins/south/test_http_south.py
praveen-garg: asyncio.ensure_future(asyncio.sleep(0.1)) is an asyncio task itself. we just want to make sure its called and its an async. otherwise we shall need a wrapper to mock async.

tests/unit/python/foglamp/services/core/api/test_common_ping.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+@pytest.mark.skip(reason="To be done with self signed certs")
+async def test_ping_https(test_server, test_client, loop):
+    app = web.Application()
+    # fill route table
+    routes.setup(app)
+
+    server = await test_server(app)
+    # server.start_server(loop=loop, ssl=<ssl_ctx>)
```
praveen-garg: @ashwinscale  Added here with skipped test for better clarification on test_server fixture

praveen-garg: Once its tested here, we don;t need to do at other places e.g. api/test_audit.py

praveen-garg:
@ashwinscale @ashish-jabble @Singhal-Vaibhav 
can we merge https://github.com/foglamp/FogLAMP/pull/504 and move forward with next final version for ping with https + audit with side-effects and then close corresponding tickets

PR #510, opened 2018-01-31
FOGL-997 - System test suite and smoke test

izoratti:


ashwinscale:
We should have push to move a lot of the scripting into python.

izoratti:
I don't know how to reply to @ashwinscale comments, so I will write here...

>>> We should have push to move a lot of the scripting into python.
For system tests we must do the exact opposite.

The whole point is to have the simplest, easily available testing environment, and that is a simple shell script. The majority of the tests are one line long, usually a command like curl, foglamp or grep. There is no need to have python or else involved. 

That said, there may be a need for more complex steps that require Python or something else, and the suite is designed to accommodate this need.

As I mentioned in my message on slack, I looked at pytest and it is definitely not designed to have this kind of tests, but it looks perfect for unit tests of python modules.

ashwinscale:
I simply meant to use python as the scripting language rather than shell. Shell commands are often different across operating system versions, and operating systems themselves. Leveraging python to carry out some of these tasks makes it more portable. 

> Pytest
I was not talking referring to pytest here.

PR #512, opened 2018-02-01
Fogl 959.a

Singhal-Vaibhav:


tests/integration/__template__.py
```diff
@@ -46,3 +46,9 @@ def test_some_integration_component(self, _module_fixture, test_input, expected)
         """
         # assertions to verify that the actual output of a component is equal to the expected output
         assert test_input == expected
+
+    def test_other_integration_component(self, _module_fixture):
+        """Purpose of the test, This test is called once.
+        """
+        # assertions to verify that the actual output of a component is equal to the expected output
+        assert "actual" != "expected"
```
praveen-garg: != :)

praveen-garg: let's align with the doc https://docs.google.com/document/d/1AjHR-E7ZKn3bit21HXY4EG6eDgSloPS7ttl2fEcY4ec/edit#  assert expected = actual

praveen-garg:
Also please fix https://github.com/foglamp/FogLAMP/tree/develop/tests/unit/python per fix in https://github.com/foglamp/FogLAMP/pull/505 
1) We should not use explicit event_loop
2) As to be used in with context if used in next statements

Singhal-Vaibhav:
Changes as per feedback,
Re BASE_URL, other are using port in a different way like:
`_m_port = pytest.test_env.core_mgmt_port`

PR #513, opened 2018-02-01
FOGL-1004 - Start Schedule method is not working for type STARTUP

amarendra-dianomic:


amarendra-dianomic:
@MarkRiddoch the new API is required for a very specific use case which I have added in method documentation and I reproduce here for reference:

> This method caters to the use case when a microservice, e.g. South service, which has been started by the Scheduler and then is shutdown vide api and then is needed to be restarted. It removes the Scheduler's record of the task related to the STARTUP schedule (which is not removed when shutdown action is taken by the microservice api as the microservice is running in a separate process and hinders starting a schedule by Scheduler's queue_task() method).

Also, since microservice and scheduler are running in two **separate processes**, rest api has been chosen as the preferred method of communication between them. I must add that this very specific use case could be envisaged before and it was pointed out by @Singhal-Vaibhav during his testing of FOGL-933 PR.

Also, this problem does not arise when we disable/enable schedules or cancel/start tasks as all methods run in the same process.

Hope this answers your question.

praveen-garg:
I'm not sure, if we need a START method for STARTUP type schedules. Those are meant to be started with foglamp start itself.

amarendra-dianomic:
@MarkRiddoch absolutely right and I totally agree. That is why this clean up is being initiated by the shutdown method of the microservice. The only reason for providing this api is that Scheduler, part of Core process, maintains some internal state for the microservice when it starts a STARTUP type schedule and this never gets cleaned up when the South service, **running in a separate process**, shuts down. That is why South service needs to tell the Scheduler (in Core) to clean its internal state concerning that particular microservice aka STARTUP task **, with the help of this api**. Otherwise it will interfere with the restarting the microservice aka STARTUP task **because the microservice aka STARTUP task can be restarted vide Scheduler only**.

amarendra-dianomic:
@MarkRiddoch all changes done as advised. Now cleaning of scheduler internal state related to microservice is being done while unregistering the microservice. Hope this is OK. Please check.

amarendra-dianomic:
@praveen-garg changes done as discussed. Please check.

praveen-garg:
@Singhal-Vaibhav  please do a functional test, and merge!

amarendra-dianomic:
@Singhal-Vaibhav all done, pl check.

amarendra-dianomic:
Will be taken up after [FOGL-1093](https://scaledb.atlassian.net/browse/FOGL-1093) is resolved.

PR #518, opened 2018-02-02
Fogl 924 - cc2650poll plugin fails with error

amarendra-dianomic:
This PR replaces [PR#457](https://github.com/foglamp/FogLAMP/pull/457).

As required by @ashwinscale, this PR provides solution for retries, with plugin restart, in case CC POLL plugin hangs in middle or develops some trouble midway or needs a hard keypress restart midway. To test this PR,
- Start server with CC2650POLL plugin enabled but sensortag NOT switched on.
- Observe syslog, after 1-2 startup retries, switch on the sensortag.
- Observe readings table. You will see readings coming in.
- Switch off sensortag.
- Observe syslog and readings table. Readings will stop coming.
- Now Switch on sensortag again.
- Watch syslog and readings table. Data will start coming again.

### DIFF
--- a/python/foglamp/plugins/south/common/sensortag_cc2650.py
+++ b/python/foglamp/plugins/south/common/sensortag_cc2650.py
@@ -136,24 +136,30 @@ class SensorTagCC2650(object):
     is_connected = False
 
     def __init__(self, bluetooth_adr, timeout):
-        try:
-            self.bluetooth_adr = bluetooth_adr
+        attempt_count = 1
+        while attempt_count <= 5:
+            try:
+                self.bluetooth_adr = bluetooth_adr
 
-            self.con = pexpect.spawn('gatttool -b ' + bluetooth_adr + ' --interactive')
-            self.con.expect('\[LE\]>', timeout=int(timeout))
+                self.con = pexpect.spawn('gatttool -b ' + bluetooth_adr + ' --interactive')
+                self.con.expect('\[LE\]>', timeout=int(timeout))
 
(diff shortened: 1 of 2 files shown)


amarendra-dianomic:
@MarkRiddoch your concern is valid but please note that,
- I could never reproduce the scenario as mentioned in FOGL-924 by @ashwinscale. In order to address this issue, I had to manually swtich on/off Sensortag in order to simulate the reported scenarios.
- Please also see [FOGL-988](https://scaledb.atlassian.net/browse/FOGL-988?jql=reporter%20%3D%20ashish) reported by @ashish-jabble . This also could not be reproduced. Later, it was confirmed that Ashish's Sensortag misbehaved due to faulty battery. Similarly, Ashwin also found batter issue with the Sensortags and later it was resolved with the change of the battery. 

Therefore, I am forced to believe that both FOGL-924 and FOGL-988 occurred due to external reasons and not because of any shortcoming in the plugin code. Also, plugin code has undergone revisions since this issue was reported and I am sure the current code is a kind of more stable. However, this issue has definitely made us more aware of the use cases and scenarios that may arise in due course and therefore will help us to make our code more robust. This PR is an attempt in this direction.

Secondly, I also found one excellent library [bluepy](https://github.com/IanHarvey/bluepy) to be used exclusively with Sensortag. But it has got dependencies on C code and though I have not examined it in detail but I believe it uses hcitool instead of gatttool. I think, we can also examine this library as an alternative.

PR #523, opened 2018-02-05
FOGL-1012 (#2) https server start ability and ping test

praveen-garg:
```
(FOGL-1012.b)$ pytest -s tests/unit/python/foglamp/services/core/api
========================================================= test session starts ==========================================================
platform linux -- Python 3.5.2, pytest-3.3.2, py-1.5.2, pluggy-0.6.0
rootdir: /home/foglamp/foglamp/FogLAMP, inifile:
plugins: mock-1.6.0, cov-2.5.1, asyncio-0.6.0, allure-adaptor-1.7.10, aiohttp-0.3.0
collected 15 items                                                                                                                     

tests/unit/python/foglamp/services/core/api/test_audit.py .............                                                          [ 86%]
tests/unit/python/foglamp/services/core/api/test_common_ping.py ..                                                               [100%]

====================================================== 15 passed in 0.56 seconds =======================================================

(FOGL-1012.b)$ pytest -s tests/unit
========================================================= test session starts ==========================================================
platform linux -- Python 3.5.2, pytest-3.3.2, py-1.5.2, pluggy-0.6.0
rootdir: /home/foglamp/foglamp/FogLAMP, inifile:
plugins: mock-1.6.0, cov-2.5.1, asyncio-0.6.0, allure-adaptor-1.7.10, aiohttp-0.3.0
collected 92 items                                                                                                                     

tests/unit/python/foglamp/common/storage/test_payload_builder.py ...................................................s........... [ 68%]
                                                                                                                                 [ 68%]
tests/unit/python/foglamp/plugins/north/common/test_common.py .                                                                  [ 69%]
tests/unit/python/foglamp/plugins/south/test_http_south.py ...                                                                   [ 72%]
tests/unit/python/foglamp/services/common/microservice_management/test_instance.py ..........                                    [ 83%]
tests/unit/python/foglamp/services/core/api/test_audit.py .............                                                          [ 97%]
tests/unit/python/foglamp/services/core/api/test_common_ping.py ..                                                               [100%]

================================================= 91 passed, 1 skipped in 0.97 seconds =================================================
```


tests/unit/python/foglamp/services/core/api/certs/foglamp.cert
```diff
@@ -0,0 +1,15 @@
+-----BEGIN CERTIFICATE-----
```
ashwinscale: @izoratti what is your take on having keys in the repo for the purpose of testing?

praveen-garg: @izoratti  please confirm.

ashwinscale:
Using the certificates script makes it an integration test in my opinion. 
At the same time, I don't think we should keep private keys in the repo either.
We could use https://stackoverflow.com/questions/27164354/create-a-self-signed-x509-certificate-in-python potentially if it works.

praveen-garg:
> Using the certificates script makes it an integration test in my opinion.

I agree! 

> really dislike leaving certs and keys in the code, 
>  At the same time, I don't think we should keep private keys in the repo either.

For real code and/or Not self-signed certificates Yes (and we gitignored them too). For tests, having these self signed test certificates (via push force) does not look a bad idea to me.

But yes, as mentioned in FOGL-780 __create_ssl subprocess was tried and it had some dependency errors). May be we can have a setup script to set up unit test base and call that with some test env vars etc. if there is some config based thing

praveen-garg:
shall be handled with FOGL-1091 after core server tests.

PR #525, opened 2018-02-06
FOGL-765 fogbench send to HTTP south ability;  FOGL-1084 STANDARD CoAP/ HTTP south error handling with code and message

praveen-garg:
- better error handing for CoAP and Http south; 
- fogbench allows now to send to coap or http (-p | --payload {coap,http}), 
    - fixed stats counter based on real acknowledgement only,
    -  and removed extra loop for read block iterator 

> Tested with fogbench coap (and now http too) and node based coap+ http sender | looks fine
> key error are resulting in 4.00 or 400


python/foglamp/plugins/south/coap_listen/coap_listen.py
```diff
@@ -193,44 +199,43 @@ class CoAPIngest(aiocoap.resource.Resource):
...
         try:
             if not Ingest.is_available():
+                Ingest.increment_discarded_readings()
                 message = '{"busy": true}'
-            else:
-                payload = cbor2.loads(request.payload)
-
-                if not isinstance(payload, dict):
-                    raise ValueError('Payload must be a dictionary')
+                raise aiocoap.error.CommunicationKilled(message)
```
praveen-garg: for reference https://github.com/chrysn/aiocoap/blob/master/aiocoap/error.py#L176

praveen-garg:
@izoratti @izoratti-dianomic please  review.

extras/python/fogbench/__main__.py
```diff
@@ -172,15 +173,25 @@ def read_out_file(_file=None, _keep=False, _iterations=1, _interval=0):
...
+                is_sent = loop.run_until_complete(send_to_http(r))
+
+            if not is_sent:
+                break
+
             msg_transferred_itr += 1
             byte_transferred_itr += sys.getsizeof(r)
 
-        # Start sending the messages to server
-        _start_time.append(datetime.now())  # Start time of every iteration
```
izoratti: This start time must be reinstated at this point.

extras/python/fogbench/__main__.py
```diff
@@ -172,15 +173,25 @@ def read_out_file(_file=None, _keep=False, _iterations=1, _interval=0):
...
-        for r in readings_list:
```
izoratti: Same for this loop, it must be here.

extras/python/fogbench/__main__.py
```diff
@@ -172,15 +173,25 @@ def read_out_file(_file=None, _keep=False, _iterations=1, _interval=0):
...
         byte_transferred_itr = 0  # Bytes transferred in every iteration
+
+        # Start sending the messages to server
+        _start_time.append(datetime.now())  # Start time of every iteration
+
         for idx, r in enumerate(readings_list):
             readings_list[idx]["key"] = str(uuid.uuid4())
+
+            is_sent = False
+            if send_to == 'coap':
```
izoratti: This if must control the loops, i.e. the if will then allow a loop to coap or http on condition.

praveen-garg: @izoratti  please check https://github.com/foglamp/FogLAMP/pull/541 do we want the fix it *(back to) like that?!

Singhal-Vaibhav:
@izoratti 
For 1 sec, let us assume the code here is good, (which is believe it is, getting rid of multiple for loops)
what was the expectations in terms of output? 

I dont see any change in output, for say, earlier (end_time-start_time = x sec) and now its is (end_time-start_time = x+10 seconds) ? Do you see a performance degradation? If yes, please help me with the scenario and output.

praveen-garg:
@izoratti loop concern is resolved via #541. 

@MarkRiddoch @ashwinscale  please have a look and expedite the merge!

PR #527, opened 2018-02-07
701 exception handling fixes

praveen-garg:
exception handling fixes, 
using event loop fixture for test; 
other minor miss fixes

### DIFF
--- a/python/foglamp/services/south/server.py
+++ b/python/foglamp/services/south/server.py
@@ -186,8 +186,9 @@ class Server(FoglampMicroservice):
             try:
                 self._plugin.plugin_shutdown(self._plugin_handle)
             except Exception as ex:
-                _LOGGER.exception("Unable to stop plugin '{}'".format(self._name))
-                raise ex
+                _LOGGER.exception("Unable to stop plugin '%s' | reason: %s", self._name, str(ex))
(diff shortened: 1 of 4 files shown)


python/foglamp/services/south/server.py
```diff
@@ -187,6 +187,7 @@ def run(self):
                 self._plugin.plugin_shutdown(self._plugin_handle)
             except Exception as ex:
                 _LOGGER.exception("Unable to stop plugin '{}'".format(self._name))
+                raise ex
```
MarkRiddoch: This has been deliberately removed as it allows a plugin to prevent a south plugin from shutting down.

python/foglamp/services/south/server.py
```diff
@@ -196,6 +197,7 @@ def run(self):
             _LOGGER.info('Stopped the Ingest server.')
         except Exception as ex:
             _LOGGER.exception('Unable to stop the Ingest server. %s', str(ex))
+            raise ex
```
MarkRiddoch: Likewise this was deliberately removed

praveen-garg:
@MarkRiddoch feedback changes done! we should have at-least for Ingest and remove for plug-in
Because Ingest code has nothing to do with hardware integration and it must be in fine control of us.

PR #534, opened 2018-02-09
FOGL-925 Added registration mechanism for statistics in order to allo…

MarkRiddoch:
…w new stats to be added without chznging init.sql


python/foglamp/common/statistics.py
```diff
@@ -74,25 +84,37 @@ def __init__(self, storage):
...
+            return
+        if len(self._registered_keys) == 0:
+            self._load_keys()
+        try:
+            payload = PayloadBuilder().INSERT(key=key, description=description, value=0, previous_value=0).payload()
+            self._storage.insert_into_tbl("statistics", payload)
+            self._registered_keys.append(key)
+        except Exception as ex:
+            """ The error may be because the key has been created in another process, reload keys """
+            self._load_keys()
```
ashish-jabble: Will it not be an expensive overhead call? Can't we do it via upsert?
PostgreSQL added the ON CONFLICT (key) target action clause (do update) to the INSERT statement to support the upsert feature.

MarkRiddoch: No in both cases. Because the class is a singleton it will happen once per process. So it is not an overhead on ever statistic insertion.

We can do it at the storage layer because it does not have enough information to register the statistic, it does not have the description. The previous workaround used a fixed description that would not be applicable for statistics being added by anything other than a south service. This mechanism is more flexible.

ashish-jabble:
@MarkRiddoch Please resolve conflicts

python/foglamp/common/statistics.py
```diff
@@ -80,12 +90,36 @@ def __init__(self, storage):
...
+            if not key in _self.registered_key:
```
ashish-jabble: Right one `self._registered_keys`

PR #535, opened 2018-02-09
Fix unit suite (to use asyncio marking and fixture only as needed)

praveen-garg:


tests/unit/python/foglamp/plugins/south/test_http_south.py
```diff
@@ -59,7 +58,11 @@ class TestHttpSouthIngest(object):
                 }
             }
         }"""
-        with patch.object(Ingest, 'add_readings', return_value=asyncio.ensure_future(asyncio.sleep(0.1))):
+
+        async def mock_coro():
```
praveen-garg: just to make consistent and simple!

tests/unit/python/foglamp/plugins/south/test_http_south.py
```diff
@@ -46,6 +44,7 @@ class TestHttpSouthIngest(object):
     """Unit tests foglamp.plugins.south.http_south.http_south.HttpSouthIngest
     """
 
+    @pytest.mark.asyncio
     async def test_post_sensor_reading_ok(self, event_loop):
```
praveen-garg: need to fix this failure in **suite** only! Neither earlier nor now its failing in as a module test or single def test!

PR #551, opened 2018-02-13
FOG-979: PID File for foglamp core

pintomax:
PID File for foglamp core.

Pid file is created with JSON data in it at startup, just after REST API details are known.
```
Feb 13 16:11:25 ubuntu16-server-virt FogLAMP[1066] INFO: server: foglamp.services.core.server: Announce management API service
Feb 13 16:11:25 ubuntu16-server-virt FogLAMP[1066] INFO: server: foglamp.services.core.server: PID [1066] written in [/home/ubuntu/source/FogLAMP/data/var/run/foglamp.core.pid]
Feb 13 16:11:25 ubuntu16-server-virt FogLAMP[1066] INFO: server: foglamp.services.core.server: REST API Server started on http://0.0.0.0:8081
```

It's removed when FogLAMP core ends:

```
Feb 13 16:16:11 ubuntu16-server-virt FogLAMP[1066] INFO: server: foglamp.services.core.server: FogLAMP PID file [/home/ubuntu/source/FogLAMP/data/var/run/foglamp.core.pid] removed.
Feb 13 16:16:11 ubuntu16-server-virt FogLAMP Storage[1071]: Storage service shut down.
Feb 13 16:16:11 ubuntu16-server-virt FogLAMP [1328] INFO: script.foglamp: FogLAMP stopped.
```

PID file content:

```
{"processID": 27114, "adminAPI": {"protocol": "HTTP", "addresses": ["0.0.0.0"], "port": 8081}}
```
The same content, in a more readable version.
```json
{
  "processID": 27114,
  "adminAPI": {
    "protocol": "HTTP",
    "addresses": [
      "0.0.0.0"
    ],
    "port": 8081
  }
}
```


python/foglamp/services/core/server.py
```diff
@@ -356,6 +361,87 @@ def _start_app(cls, loop, app, host, port, ssl_ctx=None):
...
+                    raise
+            except Exception as ex:
+                errmsg = "FogLAMP PID file create error: [" + ex.__class__.__name__ + "], (" + format(str(ex)) + ")"
+                _logger.error(errmsg)
+                sys.stderr.write(errmsg + "\n")
+                raise
+
+            """ Build the JSON object to write into PID file """
+            info_data = {'processID' : pid,\
+                         'REST_info' : {\
```
MarkRiddoch: It might be better to call this adminAPI rather than REST_info as we have more than one REST API

python/foglamp/services/core/server.py
```diff
@@ -356,6 +361,87 @@ def _start_app(cls, loop, app, host, port, ssl_ctx=None):
...
+                             "protocol": "HTTP" if cls.is_rest_server_http_enabled else "HTTPS",\
+                             "address": api_address,\
```
MarkRiddoch: I would be tempted to make the address a list for future expansion. That way when we want to support binding to multiple interfaces we can. The current 0.0.0.0 approach is not great as this is not really the address we listen on, but the address we bind to and the meaning is subtly different.

PR #553, opened 2018-02-14
FOGL-1066: Unit test for jqfilter.py

Singhal-Vaibhav:
- [X] Unit test added

- Test Results:

 `4 passed in 0.04 seconds`

- Code coverage:

```
Coverage for /FogLAMP/python/foglamp/common/jqfilter.py : 100% 
19 statements   19 run 0 missing 0 excluded
```

- [X] Removed unused super().__init__() from jqfilter init

### DIFF
--- a/tests/unit/python/foglamp/common/test_jqfilter.py
+++ b/tests/unit/python/foglamp/common/test_jqfilter.py
@@ -7,13 +7,11 @@
 """ Test common/jqfilter.py
 
 """
-
+from unittest.mock import patch
 import pytest
 import pyjq
-from unittest.mock import patch
 from foglamp.common import logger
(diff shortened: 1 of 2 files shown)


tests/unit/python/foglamp/common/test_jqfilter.py
```diff
@@ -0,0 +1,55 @@
...
+# See: http://foglamp.readthedocs.io/
+# FOGLAMP_END
+
+""" Test common/jqfilter.py
+
+"""
+
+import pytest
+import pyjq
+from unittest.mock import patch
```
ashish-jabble: standard import "from unittest.mock import patch" should be placed before "import pytest"

PR #561, opened 2018-02-15
FOGL-724 Add mechanism to add new service with specified plugin

MarkRiddoch:


python/foglamp/services/core/api/service.py
```diff
@@ -0,0 +1,94 @@
...
+"""
+
+
+
+async def add_service(request):
+    """
+    Create a new service to run a specific plugin
+
+    :Example:
+             curl -X POST /foglamp/service -d '{"name": "furnace4", "type": "south", "plugin": "dht11"}'
```
ashish-jabble: `curl -d '{"name": "furnace4", "type": "south", "plugin": "dht11"}'  -X POST  http://localhost:8081/foglamp/service`

python/foglamp/services/core/api/service.py
```diff
@@ -0,0 +1,94 @@
...
+
+        # Finally add a schedule to run the new service at startup
+        schedule = StartUpSchedule()
+        schedule.name = name
+        schedule.process_name = name
+        schedule.repeat = datetime.timedelta(0)
+        schedule.exclusive = True
+        schedule.enabled = False
+        # Save schedule
+        await server.Server.scheduler.save_schedule(schedule)
```
ashish-jabble: Right now with the propose code multiple entries are possible with same schedule name, process name
Is this correct? I believe that schedule name should be unique. For example if we have multiple sensor tags installed then each will have a unique schedule name, unique configuration entry but may use same plugin name.

python/foglamp/services/core/api/service.py
```diff
@@ -0,0 +1,94 @@
...
+    :Example:
+             curl -X POST /foglamp/service -d '{"name": "furnace4", "type": "south", "plugin": "dht11"}'
+    """
+
+    try:
+        data = await request.json()
+
+        name = data.get('name', None)
+        plugin = data.get('plugin', None)
+        type = data.get('type', None)
```
ashish-jabble: Shadows built-in name 'type'

python/foglamp/services/core/api/service.py
MarkRiddoch: In reality it is fine because the uniqueness of processes names will catch it earlier and it will not get as far as adding a schedule

amarendra-dianomic: @MarkRiddoch this does not seem to be correct for below reasons:
- scheduled_processes table + configuration table do enforce uniqueness of process_name but this is not happening in case of schedules table. Schedule table enforces uniqueness of rows by schedule_id and not by schedule_name+process_name.
- @ashish-jabble has actually run the code and found duplicate entries being created in schedules table, with the same example provided in the code.
- Since there is no unique constraint defined in the database for schedule_name+process_name combo in schedules table, this uniqueness has to be enforced in code. A better alternative would be to enforce the constraint at database level.

MarkRiddoch: No, but if the scheduled_process can not be inserted the code stops at that point. Since thew schedule name and processName are the same this will give us what we want

amarendra-dianomic: @MarkRiddoch you are absolutely right about this desired behavior **but the problem is that it is not happening**. @ashish-jabble  has actually been able to insert duplicate entries into schedules table by running this code. You can also try this at your local. So we need to fix this.

MarkRiddoch: @izoratti-dianomic I would suggest a change to the data model here to enforce this. Doing it at this level would not be ideal. Let me have your thoughts?

MarkRiddoch: It is likely this is also a result of a bug in the storage client which is not reporting failures back to the caller.

PR #563, opened 2018-02-16
FOGL-1042 monitor unit tests

ashwinscale:


python/foglamp/services/core/service_registry/monitor.py
```diff
@@ -44,8 +44,11 @@ def __init__(self):
         self._ping_timeout = None  # type: int
         """Timeout for a response from any given micro-service"""
 
+    async def _sleep(self, sleep_time):
```
ashwinscale: Note that I put a wrapper around asyncio.sleep. The code runs an infinite loop in which asyncio.sleep is called routinely. I wanted to be able to use the sleep as a safe exit point (mock sleep to throw an exception when I desire it to do so). I also wanted to eliminate sleep time so that the unit test executes quickly. @MarkRiddoch  please let me know if this is okay. It is strictly a change to accommodate unit tests.

python/foglamp/services/core/service_registry/monitor.py
```diff
@@ -44,8 +44,11 @@ def __init__(self):
...
+        await asyncio.sleep(sleep_time)
+
     async def _monitor_loop(self):
-        """Main loop for the scheduler"""
+        """async Monitor loop to monitor registered services"""
         # check health of all micro-services every N seconds
         _MAX_ATTEMPTS = 15
         """Number of max attempts for finding a heartbeat of service"""
```
praveen-garg: I would suggest to make it module level variable, so that in *test* we can set to 2 or 3 and minimize the loop time.

python/foglamp/services/core/service_registry/monitor.py
MarkRiddoch: Yes certainly, in this case we are not creating a different code flow for test, so I think it is fine.

python/foglamp/services/core/service_registry/monitor.py
ashwinscale: We can make it part of the configuration as well. But let's consider it during refactoring rather than now.

.gitignore
```diff
@@ -49,3 +49,7 @@ docs/.cache/
 *.key
 
 .cache/
+
+# test
+.pytest_cache
+.coverage
```
ashish-jabble: fyi, test ignored files under `$FOGLAMP_ROOT/tests/.gitignore` and I think this is the right place to maintain it. 
We have to run pytest command under `$FOGLAMP_ROOT/tests/unit/python/foglamp`, though pytest autodiscover the tests but right now integration tests are creating a problem. If we want to run all the tests outside the tests dir path then do `pytest --ignore=tests/integration`

tests/unit/python/foglamp/services/core/service_registry/test_monitor.py
```diff
@@ -0,0 +1,109 @@
...
+
+            async def __aexit__(self, *args):
+                return None
+        # as monitor loop is as infinite loop, this exception is thrown when we need to exit the loop
+
+        class TestMonitorException(Exception):
+            pass
+        # register a service
+        s_id_1 = ServiceRegistry.register(
+            'sname1', 'Storage', 'saddress1', 1, 1,  'protocol1')
```
ashish-jabble: trivial but extra whitespace before 'protocol1'

tests/unit/python/foglamp/services/core/service_registry/test_monitor.py
```diff
@@ -0,0 +1,109 @@
...
```
ashish-jabble: I see most of the code is repeating for both test scenarios, just asking. refactor?

PR #564, opened 2018-02-16
FOGL-1007 Unit test for api/backup_restore.py

Singhal-Vaibhav:
- [X] Unit test added

- Test Results:

 `35 passed in 2.37 seconds`

- Code coverage:

```
Coverage for FogLAMP/python/foglamp/services/core/api/backup_restore.py : 100%
103 statements   103 run 0 missing 0 excluded
```

- [X] Fixed FOGL-1102

### DIFF
--- a/python/foglamp/plugins/storage/postgres/backup_restore/backup_postgres.py
+++ b/python/foglamp/plugins/storage/postgres/backup_restore/backup_postgres.py
@@ -394,8 +394,8 @@ class BackupProcess(FoglampProcess):
         if status != lib.BackupStatus.COMPLETED:
 
             self._logger.error(self._MESSAGES_LIST["e000007"])
-            raise exceptions.BackupFailed
             loop.run_until_complete(audit.information('BKEXC', {'status': 'failed'}))
+            raise exceptions.BackupFailed
         else:
             loop.run_until_complete(audit.information('BKEXC', {'status': 'completed'}))
 
--- a/python/foglamp/services/core/api/backup_restore.py
+++ b/python/foglamp/services/core/api/backup_restore.py
@@ -118,8 +118,6 @@ async def get_backup_details(request):
     :Example: curl -X GET http://localhost:8081/foglamp/backup/1
     """
     backup_id = request.match_info.get('backup_id', None)
-    if not backup_id:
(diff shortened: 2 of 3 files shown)


python/foglamp/plugins/storage/postgres/backup_restore/backup_postgres.py
```diff
@@ -394,8 +394,8 @@ def execute_backup(self):
         if status != lib.BackupStatus.COMPLETED:
 
             self._logger.error(self._MESSAGES_LIST["e000007"])
-            raise exceptions.BackupFailed
             loop.run_until_complete(audit.information('BKEXC', {'status': 'failed'}))
+            raise exceptions.BackupFailed
```
Singhal-Vaibhav: Fix for FOGL-1102

python/foglamp/services/core/api/backup_restore.py
```diff
@@ -118,8 +118,6 @@ def _get_status(status_code):
     :Example: curl -X GET http://localhost:8081/foglamp/backup/1
     """
     backup_id = request.match_info.get('backup_id', None)
-    if not backup_id:
-        raise web.HTTPBadRequest(reason='Backup id is required')
```
Singhal-Vaibhav: This code is never called, before call comes here, its handled by routes `404: Not Found`

python/foglamp/services/core/api/backup_restore.py
```diff
@@ -146,8 +144,6 @@ def _get_status(status_code):
     :Example: curl -X DELETE http://localhost:8081/foglamp/backup/1
     """
     backup_id = request.match_info.get('backup_id', None)
-    if not backup_id:
-        raise web.HTTPBadRequest(reason='Backup id is required')
```
Singhal-Vaibhav: This code is never called, before call comes here, its handled by routes `404: Not Found`

python/foglamp/plugins/storage/postgres/backup_restore/backup_postgres.py
Singhal-Vaibhav: Fix for FOGL-1102

PR #570, opened 2018-02-19
FOGL-1004 - Start Schedule method is not working after shutdown of a STARTUP schedule

amarendra-dianomic:
Replaces #513


amarendra-dianomic:
@Singhal-Vaibhav I am unable to reproduce the error at my local. The service is shutting down perfectly. Could you please check it again?

Singhal-Vaibhav:
Looks good, tested with scenarios listed here and in pull #513

PR #578, opened 2018-02-19
FOGL-1099 - Deleting a running schedule does not kills the process and/or re-registers the service

amarendra-dianomic:


amarendra-dianomic:
@Singhal-Vaibhav log message added. Pl check.

PR #580, opened 2018-02-19
removed eval usage; and consistent severity level 

praveen-garg:
- [x] removed eval usage;  (vulnerable for external inputs)
- [x] severity levels are widespread but consistent now…; 
- [x] other formatting fixes

**test results:**

`================================================== 324 passed, 1 skipped in 6.51 seconds ==================================================`

### DIFF
--- a/python/foglamp/services/core/api/audit.py
+++ b/python/foglamp/services/core/api/audit.py
@@ -57,7 +57,7 @@ async def create_audit_entry(request):
         POST /foglamp/audit
 
         {
-                "source"   : "LocalMonitor",
+                "source"   : "LMTR", # 5 char max
                 "severity" : "WARNING",
                 "details"  : {
                                 message" : "Engine oil pressure low"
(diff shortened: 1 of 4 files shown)


python/foglamp/services/core/api/audit.py
```diff
@@ -82,67 +84,52 @@ class Severity(IntEnum):
...
         if return_error is True:
-            raise web.HTTPNotFound(reason={"error": err_msg})
+            raise web.HTTPBadRequest(reason={"error": err_msg})
 
-        # Prepare the audit log method to call
-        audit_method = "audit." + str(severity).lower() + "(source, details)"
-
-        # Instantiate the audit log class
         audit = AuditLogger()
+        getattr(audit, str(severity).lower())(source, details)
```
pintomax: I see "await" is not in use with the code modification.

praveen-garg: great catch! thx

python/foglamp/services/core/api/audit.py
```diff
@@ -49,100 +50,83 @@ class Severity(IntEnum):
...
 
-        # Get input data
-        json_data = await request.json()
+        payload = await request.json()
 
-        # Set items to store
-        severity = json_data.get(_SEVERITY_PARAM)
-        source = json_data.get(_SOURCE_PARAM)
-        details = json_data.get(_DETAILS_PARAM)
+        severity = payload.get("severity")
```
MarkRiddoch: I think this was better before rather than having repeated strings in the code.

praveen-garg: same I believe. 3 vars + each used at 2 places only;

PR #587, opened 2018-02-20
FOGL-1045 unit tests for services/core/interest_registry/change_callback.py

ashwinscale:


praveen-garg:
Code cov is `88%` and on a [local] rebase with develop, unit suite looks good.

PR #593, opened 2018-02-21
FOGL-921: added FOGLAMP_VERSION

pintomax:
FOGL-921: added FOGLAMP_VERSION

VERSION file contais current FogLAMP version and it's loaded from Makefile

FOGLAMP_VERSION is added into scripts/foglamp via "sed" command

It's also passed to other scripts.

postgres script renamed to postgres.sh

FOGLAMP_VERSION is removed from scripts/foglamp via 'make clean'


pintomax:
Starting FogLAMP 1.1.2......
FogLAMP started.

C/plugins/storage/postgres/init.sql
```diff
@@ -1091,3 +1091,5 @@ INSERT INTO foglamp.streams ( id, destination_id, description, last_object, ts )
 INSERT INTO foglamp.destinations( id, description, ts ) VALUES ( 3, 'OCS', now() );
 INSERT INTO foglamp.streams( id, destination_id, description, last_object, ts ) VALUES ( 4, 3, 'OCS north', 0, now() );
 
+-- FogLAMP DB version
+CREATE TABLE foglamp.version (id CHAR(10));
```
MarkRiddoch: Is this the FogLAMP version or the schema version?

scripts/foglamp
```diff
@@ -121,7 +123,7 @@ foglamp_start() {
 
         "2")
             #FogLAMP not running
-            echo -n "Starting FogLAMP."
+            echo -n "Starting FogLAMP ${FOGLAMP_VERSION}."
```
praveen-garg: `Starting FogLAMP 1.1.2......`

better would be to echo `Starting FogLAMP v1.1.2 ......`

C/plugins/storage/postgres/init.sql
pintomax: This is the FogLAMP version, we are currently checking where to save the schema version

scripts/foglamp
pintomax: Added v${FOGLAMP_VERSION}

Makefile
```diff
@@ -303,4 +312,6 @@ clean :
 	-$(RM_DIR) $(PYTHON_BUILD_DIR)
 	-$(RM_DIR) $(DEV_SERVICES_DIR)
 	-$(RM) $(SYMLINK_PLUGINS_DIR)
+# Remove FOGLAMP_VERSION from 'scripts/foglamp'
```
MarkRiddoch: Why not simply source the file?

pintomax: We can source the file, but the file VERSION has to be copied via "make install" in a location, which can be $FOGLAMP_ROOT/VERSION.


Not tested with Snap yet

PR #594, opened 2018-02-21
FOGL-1035 unit tests for FoglampProcess class

ashwinscale:
Note that register_interest and unregister_interest do not have unit tests as they are currently not implemented. I have opened issue https://scaledb.atlassian.net/browse/FOGL-1119 for this.

### DIFF
--- a/python/foglamp/common/process.py
+++ b/python/foglamp/common/process.py
@@ -26,13 +26,13 @@ _logger = logger.setup(__name__)
 
 
 class ArgumentParserError(Exception):
-    """ Overwrite default exception to not terminate application """
+    """ Override default exception to not terminate application """
     pass
 
 class SilentArgParse(argparse.ArgumentParser):
 
     def error(self, message):
-        """ Overwrite default error functionality to not terminate application """
+        """ Override default error functionality to not terminate application """
         raise ArgumentParserError(message)
 
     def silent_arg_parse(self, argument_name):
--- a/python/foglamp/common/process.py
+++ b/python/foglamp/common/process.py
@@ -29,6 +29,13 @@ class ArgumentParserError(Exception):
     """ Overwrite default exception to not terminate application """
     pass
 
+class SilentArgParse(argparse.ArgumentParser):
+    def error(self, message):
+        raise ArgumentParserError(message)
+    def silent_arg_parse(self, argument_name):
+        self.add_argument(argument_name)
+        parser_result = self.parse_known_args()
+        return list(vars(parser_result[0]).values())[0]
 
 class FoglampProcess(ABC):
     """ FoglampProcess for all non-core python processes.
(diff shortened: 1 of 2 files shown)


python/foglamp/common/process.py
```diff
@@ -29,6 +29,13 @@ class ArgumentParserError(Exception):
     """ Overwrite default exception to not terminate application """
     pass
 
+class SilentArgParse(argparse.ArgumentParser):
+    def error(self, message):
+        raise ArgumentParserError(message)
+    def silent_arg_parse(self, argument_name):
+        self.add_argument(argument_name)
+        parser_result = self.parse_known_args()
```
praveen-garg: should not we move this to try except? how error def is being used. 

should not we just remove def error! and:

```  try:
          parser_result = parser.parse_known_args()
       except ArgumentParserError:
     raise
       else:
          return list(vars(parser_result[0]).values())[0]
```

tests/unit/python/foglamp/common/test_process.py
```diff
@@ -0,0 +1,102 @@
...
+            fp = FoglampProcess()
+        with pytest.raises(TypeError):
+            class FoglampProcessImp(FoglampProcess):
+                pass
+            fp = FoglampProcessImp()
+
+    @pytest.mark.parametrize('argslist',
+                             [([ArgumentParserError()]),
+                              (['corehost', ArgumentParserError()]),
+                              (['corehost', 0, ArgumentParserError()])
```
praveen-garg: I don;t think this covers the code path; we should test with  combinations of`address, host and port` itself and for error try some bad say address1

python/foglamp/common/process.py
ashwinscale: The regular error method from argparse exits the application if there is an error. We simply want to raise an exception, so I have overwritten the method.

tests/unit/python/foglamp/common/test_process.py
ashwinscale: I am not sure what you mean. I have mocked an error coming from the silent parser. As we are testing application arguments, and the application, in this case, is the unit test process itself, I chose to mock the parser to simulate actual command line arguments.

python/foglamp/common/process.py
praveen-garg: I see, this is overridden method. Please add a comment. thanks for clarification.

tests/unit/python/foglamp/common/test_process.py
praveen-garg: **Please add test for ValueError handling**

```
[None, None, None] > ValueError: --address is not specified

[None, 1, 'a name'] > ValueError: --address is not specified

['corehost', None, 'a name'] > ValueError: --port is not specified

['corehost', 1, None] > ValueError: --name is not specified
```

I think code should have handling of empty string as well, Not only `None`

e.g. `if self._core_management_host is None or self._core_management_host == "":`

Test data `["", 1, "a name"]`

python/foglamp/common/process.py
```diff
@@ -29,6 +29,13 @@ class ArgumentParserError(Exception):
     """ Overwrite default exception to not terminate application """
     pass
 
+class SilentArgParse(argparse.ArgumentParser):
+    def error(self, message):
```
praveen-garg: please add a comment that is overridden and line break between def blocks

tests/unit/python/foglamp/common/test_process.py
```diff
@@ -0,0 +1,102 @@
...
+                              ])
+    def test_constructor_missing_args(self, argslist):
+        class FoglampProcessImp(FoglampProcess):
+            def run():
```
praveen-garg: add `self` as first arg as its class's def (for all such occurrences)

tests/unit/python/foglamp/common/test_process.py
ashwinscale: I didn't add cases for this as I believe the code would never return None. I didn't want to refactor too much as the issue is about adding basic unit tests without refactoring. I have filed an issue for this improvement but I believe it is okay to proceed as is.

PR #599, opened 2018-02-24
FOGL-1133 - Circular import disallows many unit tests from running

amarendra-dianomic:
Test suite result:
```
a@a ~/Development/FogLAMP/tests/unit/python/foglamp (FOGL-1133) $ pytest
======================================================= test session starts ========================================================
platform linux -- Python 3.6.3, pytest-3.4.0, py-1.5.2, pluggy-0.6.0
rootdir: /home/a/Development/FogLAMP/tests/unit/python/foglamp, inifile:
plugins: mock-1.6.0, cov-2.5.1, asyncio-0.8.0, allure-adaptor-1.7.10, aiohttp-0.3.0
collected 466 items                                                                                                                

common/test_audit_logger.py ..........                                                                                       [  2%]
common/test_configuration_manager.py ........................................................................                [ 17%]
common/test_jqfilter.py ....                                                                                                 [ 18%]
common/test_logger.py .....                                                                                                  [ 19%]
common/test_process.py ........                                                                                              [ 21%]
common/test_service_record.py .......                                                                                        [ 22%]
common/test_statistics.py ................                                                                                   [ 26%]
common/microservice_management_client/test_microservice_management_client.py .......................                         [ 31%]
common/storage_client/test_payload_builder.py ...................................................s...........                [ 44%]
common/storage_client/test_sc_exceptions.py ..........                                                                       [ 46%]
common/storage_client/test_storage_client.py ...............                                                                 [ 50%]
common/storage_client/test_utils.py ..................                                                                       [ 53%]
plugins/common/test_plugins_common_utils.py ....                                                                             [ 54%]
plugins/north/common/test_common.py .                                                                                        [ 54%]
plugins/south/test_http_south.py ...                                                                                         [ 55%]
services/common/test_avahi.py ......                                                                                         [ 56%]
services/common/test_services_common_utils.py ....                                                                           [ 57%]
services/common/microservice_management/test_instance.py ..........                                                          [ 59%]
services/core/test_connect.py ...                                                                                            [ 60%]
services/core/test_main.py .                                                                                                 [ 60%]
services/core/api/test_audit.py .................                                                                            [ 64%]
services/core/api/test_backup_restore.py ...................................                                                 [ 71%]
services/core/api/test_browser_assets.py ..........................................                                          [ 80%]
services/core/api/test_common_ping.py ...                                                                                    [ 81%]
services/core/api/test_statistics_api.py ........                                                                            [ 83%]
[...]

PR #600, opened 2018-02-24
Revert "FOGL-1004 - Start Schedule method is not working after shutdown of a STARTUP schedule"

praveen-garg:
Reverts foglamp/FogLAMP#570

* we need to think how to remove scheduler removal call from service registry, <as its a call to be made only after a service is all shutdown>


praveen-garg:
Closing in favor of #599 ; we can have another ticket refactoring.

PR #606, opened 2018-02-26
FOGL-671 Add api to core to view health of microservices

ashwinscale:


python/foglamp/services/core/api/service.py
```diff
@@ -6,24 +6,56 @@
...
+    for service_record in ServiceRegistry.all():
+        sr_list.append(
+                {
+                    'name' : service_record._name,
+                    'type' : service_record._type,
+                    'address' : service_record._address,
+                    'management_port' : service_record._management_port,
+                    'service_port' : service_record._port,
+                    'protocol' : service_record._protocol,
+                    'status': 'running'
```
MarkRiddoch: In the future this status will come from the monitored state in the service registry. However for now we only have an entry for running services, therefore this is a reasonable first implementation.

PR #609, opened 2018-02-28
FOGL-1135 - Delete an enabled schedule returns status code 200 (should return 4xx)

amarendra-dianomic:


python/foglamp/services/core/api/scheduler.py
```diff
@@ -533,6 +533,8 @@ def _extract_args(data, curr_value):
         retval, message = await server.Server.scheduler.delete_schedule(uuid.UUID(schedule_id))
 
         return web.json_response({'message': message, 'id': schedule_id})
+    except TypeError:
```
praveen-garg: @amarendra-dianomic  why its a type error?

amarendra-dianomic: Because this incident matches closest to the definition of TypeError:
6 built in exceptions python 2 7 14 documentation

amarendra-dianomic: To avoid this misunderstanding, replaced this TypeError with RuntimeWarning. Pl check.

PR #614, opened 2018-03-01
FOGL-1127 - Fetch support bundle functionality along with unit tests

ashish-jabble:
**Coverage 89%**

```
35 statements   31 run || 4 missing || 0 excluded
```

**Tests O/P**
```
collected 10 items                                                                                                                            

services/core/api/test_support.py::TestBundleSupport::test_get_support_bundle[pyloop-data0-expected_content0-2] PASSED
services/core/api/test_support.py::TestBundleSupport::test_get_support_bundle[pyloop-data1-expected_content1-1] PASSED
services/core/api/test_support.py::TestBundleSupport::test_get_support_bundle[pyloop-data2-expected_content2-0] PASSED
services/core/api/test_support.py::TestBundleSupport::test_get_support_bundle[pyloop-data3-expected_content3-0] PASSED
services/core/api/test_support.py::TestBundleSupport::test_get_support_bundle_by_name[pyloop] PASSED
services/core/api/test_support.py::TestBundleSupport::test_get_support_bundle_by_name_not_found[pyloop-data0-xsupport-180301-133523.tar.gz] PASSED
services/core/api/test_support.py::TestBundleSupport::test_get_support_bundle_by_name_not_found[pyloop-data1-support-180301-133523.tar.gz] PASSED
services/core/api/test_support.py::TestBundleSupport::test_get_support_bundle_by_name_bad_request[pyloop] PASSED
services/core/api/test_support.py::TestBundleSupport::test_get_support_bundle_by_name_no_dir[pyloop] PASSED
services/core/api/test_support.py::TestBundleSupport::test_create_support_bundle[pyloop] PASSED

=== 10 passed in 0.73 seconds ==
```

**Suite**
```
collected 644 items

=== 639 passed, 5 skipped in 31.75 seconds ====
```

### DIFF
--- a/python/foglamp/services/core/api/support.py
+++ b/python/foglamp/services/core/api/support.py
@@ -1,0 +1,79 @@
+# -*- coding: utf-8 -*-
+
+# FOGLAMP_BEGIN
+# See: http://foglamp.readthedocs.io/
+# FOGLAMP_END
+
+import os
+from aiohttp import web
+from foglamp.common import logger
+
+
+__author__ = "Ashish Jabble"
+__copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
+__license__ = "Apache 2.0"
+__version__ = "${VERSION}"
+
+
+_FOGLAMP_DATA = os.getenv("FOGLAMP_DATA", default=None)
+_FOGLAMP_ROOT = os.getenv("FOGLAMP_ROOT", default='/usr/local/foglamp')
+
+
+_logger = logger.setup(__name__, level=20)
+
+_help = """
+    -------------------------------------------------------------------------------
+    | GET POST        | /foglamp/support                                          |
+    | GET             | /foglamp/support/{bundle}                                 |
+    -------------------------------------------------------------------------------
+"""
+
+
+async def fetch_support_bundle(request):
+    """ get list of available support bundles
+
+    :Example:
+        curl -X GET http://localhost:8081/foglamp/support
+    """
+    # Get support directory path
+    support_dir = _get_support_dir()
+    valid_extension = '.tar.gz'
+    found_files = []
+    for root, dirs, files in os.walk(support_dir):
+        found_files = [f for f in files if f.endswith(valid_extension)]
+
+    return web.json_response({"bundles": found_files})
+
+
+async def fetch_support_bundle_item(request):
+    """
+    :Example:
+        curl -X GET http://localhost:8081/foglamp/support/support-180301-13%3A35%3A23.tar.gz
+    """
+    bundle_name = request.match_info.get('bundle', None)
+
+    if not str(bundle_name).endswith('.tar.gz'):
+        return web.HTTPBadRequest(reason="Bundle file extension is invalid")
+
+    if not os.path.isdir(_get_support_dir()):
+        raise web.HTTPNotFound(reason="Support bundle directory does not exist")
+
+    for root, dirs, files in os.walk(_get_support_dir()):
+        if str(bundle_name) not in files:
+            raise web.HTTPNotFound(reason='{} not found'.format(bundle_name))
+
+    return web.json_response()
+
+
+async def create_support_bundle(request):
+    # TODO: FOGL-1126
+    raise web.HTTPNotImplemented(reason='Create support bundle method is not implemented yet')
+
+
+def _get_support_dir():
+    if _FOGLAMP_DATA:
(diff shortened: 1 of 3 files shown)


python/foglamp/services/core/api/support.py
```diff
@@ -0,0 +1,79 @@
...
+async def create_support_bundle(request):
+    # TODO: FOGL-1126
+    raise web.HTTPNotImplemented(reason='Create support bundle method is not implemented yet')
+
+
+def _get_support_dir():
+    if _FOGLAMP_DATA:
+        support_dir = os.path.expanduser(_FOGLAMP_DATA + '/tmp/support')
+    else:
+        support_dir = os.path.expanduser(_FOGLAMP_ROOT + '/data/tmp/support')
```
ashish-jabble: @MarkRiddoch Please confirm here, if FOGLAMP_DATA is not set and update the JIRA

python/foglamp/services/core/api/support.py
```diff
@@ -0,0 +1,79 @@
...
+    :Example:
+        curl -X GET http://localhost:8081/foglamp/support/support-180301-13%3A35%3A23.tar.gz
+    """
+    bundle_name = request.match_info.get('bundle', None)
+
+    if not str(bundle_name).endswith('.tar.gz'):
+        return web.HTTPBadRequest(reason="Bundle file extension is invalid")
+
+    if not os.path.isdir(_get_support_dir()):
+        raise web.HTTPNotFound(reason="Support bundle directory does not exist")
```
ashish-jabble: @MarkRiddoch If directory does not exist? HTTPNOTFOUND or HTTPINTERNALSERVER or nothing to raise

tests/unit/python/foglamp/services/core/api/test_support.py
```diff
@@ -0,0 +1,104 @@
...
+                resp = await client.get('/foglamp/support')
+                assert 200 == resp.status
+                res = await resp.text()
+                jdict = json.loads(res)
+                assert expected_count == len(jdict['bundles'])
+                assert expected_content == jdict
+            mockwalk.assert_called_once_with(path)
+
+    # FIXME: issue with URL having colon in it even after encoded with %3A manually or urllib parser
+    async def test_get_support_bundle_by_name(self, client, support_bundles_dir_path):
```
praveen-garg: https://github.com/aio-libs/aiohttp/issues/2670
https://github.com/aio-libs/aiohttp/commit/06ed15ecb2a4700a08f4f2ff4b03bd751ca979f6 /cc @ashish-jabble

python/foglamp/services/core/api/support.py
```diff
@@ -0,0 +1,79 @@
...
+    valid_extension = '.tar.gz'
+    found_files = []
+    for root, dirs, files in os.walk(support_dir):
+        found_files = [f for f in files if f.endswith(valid_extension)]
+
+    return web.json_response({"bundles": found_files})
+
+
+async def fetch_support_bundle_item(request):
+    """
```
praveen-garg: add docstring

python/foglamp/services/core/api/support.py
MarkRiddoch: Yes, this is correct.

python/foglamp/services/core/api/support.py
MarkRiddoch: If there is no directory then I think we should just raise a not found error as no support bundles have been created.

tests/unit/python/foglamp/services/core/api/test_support.py
ashish-jabble: @MarkRiddoch As per specs mentioned in FOGL-1126 (The file will be called support-<YYMMDD>-<HH:MM:SS>.tar.gz).So I had added tests and example with this naming convention but somehow there is aiohttp bug on `v2.3.8` and bug has fixed in `2.3.9` **BUT with yarl dep 1.0.0**

So if we need to follow the same file naming convention,  we need to fix below things
1) aiohttp dep set to 2.3.9
2) and yarl dep explicitly set to 1.0.0

Please let me know, should I need to do above steps to make it work? or simplest and easiest way to just change the file naming convention we had in specs and do it like "support-YYMMDD-HH-MM-SS" or "support_YYMMDD_HH_MM_SS" or anything.

MarkRiddoch: I'm not too worried about the exact naming, so rather than force this update let's go for one of the alternate naming conventions. I would suggest support-YYMMDD-HH-MM-SS.

PR #626, opened 2018-03-07
FOGL-1154: First implementation of SQLite 3 storage plugin

pintomax:
First implementation of SQLite 3 storage plugin

Note:

data/etc/foglamp.json must be manually changed:
=> { "default storage plugin" : "sqlite", ...

data/etc/storage.json must be manually changed:

==> {"plugin":{"value":"sqlite"} ...

DEFAULT_SQLITE_DB_FILE env var is used in both sqlite.sh script and in Connection::Connection()

The schema_update.sh is not imoplemented yet.
It will be added in next commits


pintomax:
https://github.com/foglamp/FogLAMP/commit/64d1810d88a26d1190c16ae92d635f6365574b1f

This commit in FOGL-1172 branch solves the init.sql for postgres

PR #628, opened 2018-03-08
FOGL-1158 basic version of auth using JWT 

praveen-garg:
and in memory user objects i.e. FOGL-1157


praveen-garg:
closing in favor of #636

PR #638, opened 2018-03-12
FOGL-1176 - create support bundle request gets hung and does not clean latest request tmp files when syslog file is huge in size

amarendra-dianomic:


amarendra-dianomic:
Closed in favour of #650

PR #645, opened 2018-03-15
fix CORS issue

praveen-garg:
when verb is options, just return handler as its a preflight request to allow CORS


praveen-garg:
Please refer: https://github.com/aio-libs/aiohttp-cors/blob/master/aiohttp_cors/urldispatcher_router_adapter.py#L137

PR #647, opened 2018-03-16
FOGL-1180 -  add syslog reader endpoint to REST api

amarendra-dianomic:
Facilitate  to see syslog entries via REST api (as we do for audit logs)


praveen-garg:
can you please merge https://github.com/foglamp/FogLAMP/pull/638 into this? and then may be we just don't need syslogparser module/class at all and so the pyparsing dependency.

python/foglamp/services/core/api/support.py
```diff
@@ -89,6 +91,42 @@
...
+    try:
+        cmd = "grep -n '{}\[' {} | tail -n {}".format(source, _SYSLOG_FILE, limit)
+        a = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE).stdout.readlines()
+        c = []
+        for b in a:
+            c.append(b.decode())
+    except (OSError, Exception) as ex:
+        raise web.HTTPException(reason=str(ex))
+
+    return web.json_response({'logs': c})
```
praveen-garg: please return count key with value too

python/foglamp/services/core/api/support.py
```diff
@@ -89,6 +91,42 @@
     return web.json_response({"bundle created": bundle_name})
 
 
+async def get_syslog_entries(request):
+    """ Returns a list of syslog trail entries sorted with most recent first and total count
+        (including the criteria search if applied)
```
praveen-garg: default should return  latest 20 enteries

python/foglamp/services/core/api/support.py
```diff
@@ -89,6 +91,42 @@
...
+
+    :Example:
+        curl -X GET http://localhost:8081/foglamp/syslog
+        curl -X GET "http://localhost:8081/foglamp/syslog?limit=5"
+        curl -X GET "http://localhost:8081/foglamp/syslog?source=Storage"
+        curl -X GET "http://localhost:8081/foglamp/syslog?limit=5&source=Storage"
```
praveen-garg: can we please make source case insensitive

python/foglamp/services/core/api/support.py
```diff
@@ -89,6 +91,42 @@
...
+            raise ValueError
+    except ValueError:
+        raise web.HTTPBadRequest(reason="{} is not a valid source".format(source))
+
+    try:
+        cmd = "grep -n '{}\[' {} | tail -n {}".format(source, _SYSLOG_FILE, limit)
+        a = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE).stdout.readlines()
+        c = []
+        for b in a:
+            c.append(b.decode())
```
praveen-garg: so above 3 lines ` [c = b.decode() for b in a]` ? please add a comment what `b.decode does here;

python/foglamp/services/core/api/support.py
```diff
@@ -20,6 +19,9 @@
 
 _FOGLAMP_DATA = os.getenv("FOGLAMP_DATA", default=None)
 _FOGLAMP_ROOT = os.getenv("FOGLAMP_ROOT", default='/usr/local/foglamp')
+_SYSLOG_FILE = '/var/log/syslog'
+__DEFAULT_LIMIT = 20
+__DEFAULT_OFFSET = 0
```
praveen-garg: no offset feature? however we defined this const

amarendra-dianomic: removed

python/foglamp/services/core/api/support.py
amarendra-dianomic: already there

python/foglamp/services/core/api/support.py
amarendra-dianomic: done

python/foglamp/services/core/api/support.py
amarendra-dianomic: done

python/foglamp/services/core/api/support.py
amarendra-dianomic: done

amarendra-dianomic:
Closed in favour of #650

PR #651, opened 2018-03-19
FOGL-1194 token based logout and logout for all active sessions for self or admin

praveen-garg:
`791 passed, 12 skipped in 45.65 seconds `


praveen-garg:
Fixed: added 2 more tests!

============================================== 791 passed, 12 skipped in 45.65 seconds ===============================================

PR #652, opened 2018-03-19
FOGL-1195: restore API enabled and fixed

sstefano69:


ashish-jabble:
test_backup_restore.py, one test is failing. Fixed in https://github.com/foglamp/FogLAMP/pull/663

PR #654, opened 2018-03-19
FOGL-1188 auth optional tests added with TODO comments

ashish-jabble:
From #659 

- [x] username should be of min 4 characters
- [x] Role updation restricted for Super Admin user (even if auth is mandatory)
- [x] case insensitive match for username when user login

**Fixed items**

- [x] auth tests when authentication is optional
- [ ] 8% code coverage will also be covered with authentication => mandatory tests

Code Coverage **91%**
```
235 statements   213 run 22 missing 0 excluded
```

**TestAuthOptional tests**

```
collected 78 items                                                                                                                            
== 78 passed in 4.40 seconds ==
```

### DIFF
--- a/tests/unit/python/foglamp/services/core/api/test_auth_optional.py
+++ b/tests/unit/python/foglamp/services/core/api/test_auth_optional.py
@@ -1,0 +1,541 @@
+# -*- coding: utf-8 -*-
+
+# FOGLAMP_BEGIN
+# See: http://foglamp.readthedocs.io/
+# FOGLAMP_END
+
+import json
+from unittest.mock import patch
+from aiohttp import web
+import pytest
+
+from foglamp.common.web import middleware
+from foglamp.services.core import routes
+from foglamp.services.core.user_model import User
+from foglamp.services.core.api import auth
+
+__author__ = "Ashish Jabble"
+__copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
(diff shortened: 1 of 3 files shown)


tests/unit/python/foglamp/services/core/api/test_auth_optional.py
```diff
@@ -0,0 +1,562 @@
...
+                                assert msg == actual['message']
+                                assert expected['id'] == actual['user']['userId']
+                                assert expected['uname'] == actual['user']['userName']
+                                assert expected['role_id'] == actual['user']['roleId']
+                            patch_audit_logger_info.assert_called_once_with(msg)
+                    patch_create_user.assert_called_once_with(request_data['username'], request_data['password'], int(expected['role_id']))
+            patch_role.assert_called_once_with(int(expected['role_id']))
+        patch_logger_info.assert_called_once_with('Received %s request for %s', 'POST', '/foglamp/user')
+
+    async def test_create_user_exception(self, client):
```
Singhal-Vaibhav: What does this test cover? 500 exception - "'type' object is not subscriptable".
API is not handling and generic exception.

ashish-jabble: `with patch.object(User.Objects, 'create', return_value=Exception) as patch_create_user:`

PR #659, opened 2018-03-21
auth bug fixes and improvements

ashish-jabble:
- [x] username should be of min 4 characters
- [x] Role updation restricted for Super Admin user (even if auth is mandatory)
- [x] case insensitive match for username when user login

(path of this branch is to merge via FOGL-1188 i.e. its base branch)


praveen-garg:
Merged into #654

PR #662, opened 2018-03-21
FOGL-1207: C south plugin default date format added

pintomax:
FOGL-1207: C south plugin default date format added:

Date passed to storage layer:
Y-m-d H:M:S + microseconds + UTC timezone "+00:00"

2018-03-20 17:50:10.101893+00:00

This format is natively handled by Postgres and SQLite3


C/common/reading.cpp
```diff
@@ -74,14 +74,42 @@ void Reading::addDatapoint(Datapoint *value)
...
+	 * datetime.datetime.now(tz=datetime.timezone.utc)
+	 */
+
+	// Create datetime with seconds
+	std::strftime(date_time, sizeof(date_time),
+		      DEFAULT_DATE_TIME_FORMAT,
+		      timeinfo);
+
+	// Add microseconds 
+        snprintf(micro_s,
```
MarkRiddoch: Don't really need the string manipulation stuff, this can be done in the C++ string and avoid some duplication

C/common/reading.cpp
```diff
@@ -74,14 +74,42 @@ void Reading::addDatapoint(Datapoint *value)
...
+		 sizeof(micro_s),
+		 ".%06lu",
+		 m_timestamp.tv_usec);
+
+	// Prepare string buffer
+	string ts(date_time);
+	ts.append(micro_s);
+
+	// Append timezone:
+	ts.append("+00:00");
```
MarkRiddoch: Just do

convert << date_time << micro_s << "+00:00";

C/common/reading.cpp
pintomax: usage of strftime and microseconds formatting will be part on next commits

C/common/reading.cpp
pintomax: yes, done

PR #668, opened 2018-03-23
FOGL-1189 - handle Ping route based on allowPing=>True config

amarendra-dianomic:
Test result:
`===================================================== 9 passed in 0.47 seconds =====================================================
`
Test suite result:
`============================================= 883 passed, 12 skipped in 42.10 seconds ==============================================
`
**Checklist:**
- [x] unit test modification

- [x] Works well on Raspbian with https + auth mandatory 

- [x] Works well  with https + auth optional

- [x] warning and error logs added

- [x]  proper HTTP status code added for raise

### DIFF
--- a/python/foglamp/services/core/api/common.py
+++ b/python/foglamp/services/core/api/common.py
@@ -48,7 +48,7 @@ async def ping(request):
     except AttributeError:
         cfg_mgr = ConfigurationManager(connect.get_storage())
         category_item = await cfg_mgr.get_category_item('rest_api', 'allowPing')
-        allow_ping = True if category_item['value'] == 'true' else False
+        allow_ping = True if category_item['value'].lower() == 'true' else False
         if request.is_auth_optional is False and allow_ping is False:
             raise web.HTTPForbidden
 
--- a/python/foglamp/services/core/api/common.py
+++ b/python/foglamp/services/core/api/common.py
@@ -46,10 +46,10 @@ async def ping(request):
     try:
         auth_token = request.token
     except AttributeError:
-        cf_mgr = ConfigurationManager(connect.get_storage())
-        category_item = await cf_mgr.get_category_item('rest_api', 'allowPing')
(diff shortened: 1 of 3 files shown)


python/foglamp/services/core/api/common.py
```diff
@@ -41,6 +43,15 @@
            curl -X GET http://localhost:8081/foglamp/ping
     """
 
+    try:
+        auth_token = request.token
+    except AttributeError:
+        cfg_mgr = ConfigurationManager(connect.get_storage())
+        category_item = await cfg_mgr.get_category_item('rest_api', 'allowPing')
+        allow_ping = True if category_item['value'] == 'true' else False
+        if request.is_auth_optional is False and allow_ping is False:
```
ashish-jabble: `request.is_auth_optional is True` # auth is Optional

python/foglamp/services/core/api/common.py
```diff
@@ -41,6 +43,15 @@
            curl -X GET http://localhost:8081/foglamp/ping
     """
 
+    try:
+        auth_token = request.token
+    except AttributeError:
+        cfg_mgr = ConfigurationManager(connect.get_storage())
+        category_item = await cfg_mgr.get_category_item('rest_api', 'allowPing')
+        allow_ping = True if category_item['value'] == 'true' else False
```
ashish-jabble: Configuration manager value saved as it is for below scenarios:
curl -X PUT -d '{"value": "True"}' http://localhost:8081/foglamp/category/rest_api/allowPing
curl -X PUT -d '{"value": "TRUE"}' http://localhost:8081/foglamp/category/rest_api/allowPing

So we should handle `category_item['value'].lower()`

amarendra-dianomic: done

praveen-garg:
In case of optional => allowPing should not matter

amarendra-dianomic:
@ashish-jabble changes done per feedback. Pl check.

PR #669, opened 2018-03-23
set default config as per specs

praveen-garg:


python/foglamp/services/core/server.py
```diff
@@ -151,7 +151,7 @@ class Server:
         'authentication': {
             'description': 'To make the authentication mandatory or optional for API calls',
             'type': 'string',
-            'default': 'optional'  # make mandatory
+            'default': 'mandatory'
```
MarkRiddoch: In FOGL-1156, which defined this, it called for authentication to be optional by default.

praveen-garg:
Done :) the same stuff what was done, but had running `mandatory` in mind.

PR #672, opened 2018-03-25
FOGL-1154: handle 'schedule_day' field type

pintomax:
FOGL-1154: handle 'schedule_day' field type


python/foglamp/services/core/scheduler/scheduler.py
```diff
@@ -677,11 +677,17 @@ def _schedule_first_task(self, schedule, current_time):
 
                 schedule_id = uuid.UUID(row.get('id'))
 
+                # Data from storage layer could be a string or int
+                if type(row.get('schedule_day')) is str:
```
amarendra-dianomic: row.get('schedule_day') always returns a string. A better check will be:
`if row.get('schedule_day').strip().isdigit():`

pintomax: This modification is required as Postgres returns a string for 'schedule_day', say it has value 0,

then "0" is returned as per current datatype "smallint" in table definitiion.

SQLite3 returns 0, as INTEGER.

This modification is needed in order to deal with both Postgres and Sqlite

python/foglamp/services/core/scheduler/scheduler.py
```diff
@@ -677,11 +677,22 @@ def _schedule_first_task(self, schedule, current_time):
...
+                # row.get('schedule_day') returns an int, say 0, from SQLite
+                # and "0", as a string, from Postgres
+                # We handle here this difference
+                #
+
+                if type(row.get('schedule_day')) is str:
+                    s_day = int(row.get('schedule_day')) if row.get('schedule_day').strip() else None
+                else:
+                    s_day = int(row.get('schedule_day'))
+
```
praveen-garg: Now as we have proper int values https://github.com/foglamp/FogLAMP/pull/729 ; do we still need above workaround? /cc @amarendra-dianomic @pintomax @MarkRiddoch

PR #683, opened 2018-03-29
gitignored data back up files

ashish-jabble:


.gitignore
```diff
@@ -52,6 +52,10 @@ docs/.cache/
 
 .cache/
 
+# Backup
+data/backup
+data/etc/backup_postgres_configuration_cache.json
```
praveen-garg: should  this be `data/etc/*.json`

PR #684, opened 2018-03-29
FOGL-1021: Removal of tox.ini

Singhal-Vaibhav:


ashish-jabble:
There is no use of `Jenkinsfile` anymore, if no `tox` and `.buid.sh` is alive :-1:

PR #685, opened 2018-03-29
FOGL-1123 - Improve Service Registry and Monitor

amarendra-dianomic:
Test suite result:
`==================================================== 950 passed, 13 skipped in 44.33 seconds ====================================================
`


python/foglamp/services/core/service_registry/monitor.py
```diff
@@ -31,60 +30,61 @@ class Monitor(object):
...
+                    except:  # TODO: Fix too broad exception clause
+                        service_record._status = ServiceRecord.Status.Doubtful
+                        service_record._check_count += 1
+                        self._logger.info("Marked as doubtful micro-service %s", service_record.__repr__())
+                    else:
+                        service_record._status = ServiceRecord.Status.Running
+                        service_record._check_count = 1
+
+                    if service_record._check_count > self._max_attempts:
+                        service_record._status = ServiceRecord.Status.Failed
```
MarkRiddoch: If we unregister then this will take the service out of the service registered and hence it will not show up as failed. Unregister is really now only used when the service does a graceful shutdown.

python/foglamp/services/core/service_registry/monitor.py
```diff
@@ -31,60 +30,61 @@ class Monitor(object):
...
-                """Number of current attempt to ping url"""
-
-                url = "{}://{}:{}/foglamp/service/ping".format(service_record._protocol, service_record._address,
-                                                               service_record._management_port)
-                async with aiohttp.ClientSession() as session:
-                    while attempt_count < _MAX_ATTEMPTS + 1:
-                        # self._logger.info("Attempting service %s with try count %d", url, attempt_count)
-                        try:
+                # No need to try ping if service status is either Unregistered or Failed
+                if service_record._status in [ServiceRecord.Status.Running, ServiceRecord.Status.Doubtful]:
```
MarkRiddoch: If we also ping failed services we will allow them to recover if there is a temporary issue.

amarendra-dianomic:
@MarkRiddoch apart from feedback changes, a few other modifications have also been done for reasons as under:
- a Use Case where a failed microservice instance is removed and then re-started has forced modifications notably in interest_registry.py and service_registry.py. Now if an interest registration is attempted with details as that of an existing record, no exception is thrown and the existing registration id is returned instead. 
- Foglamp server shutdown method for microservices have been modified to account for new Status codes which we missed earlier.
- It has been assumed that in case of a service failure, other than intentional shutdown, interests records are to be retained. Meaning that in case, the service restarts(with the same service id), the earlier recorded interests may again be needed.
- Now in case of proper and graceful shutdown, we are removing interests records as well. This is a new addition and was required to distinguish it from the failed service event.

python/foglamp/common/service_record.py
```diff
@@ -25,11 +25,23 @@ class Type(IntEnum):
         Core = 2
         Southbound = 3
 
+    class Status(IntEnum):
+        """Enumeration for Service Status"""
+
+        Running = 1
+        Unregistered = 2
```
praveen-garg: what is it? why an unregistered will be in Service Registry

python/foglamp/common/service_record.py
```diff
@@ -25,11 +25,23 @@ class Type(IntEnum):
...
     class InvalidServiceType(Exception):
         # TODO: tell allowed service types?
         pass
 
-    __slots__ = ['_id', '_name', '_type', '_protocol', '_address', '_port', '_management_port', '_status']
+    class InvalidServiceStatus(Exception):
+        # TODO: tell allowed service status?
+        pass
+
+    __slots__ = ['_id', '_name', '_type', '_protocol', '_address', '_port', '_management_port', '_status', '_check_count']
```
praveen-garg: _check_count MUST not be an attribute of ServiceRecord entity.

python/foglamp/common/service_record.py
```diff
@@ -41,7 +53,8 @@ def __init__(self, s_id, s_name, s_type, s_protocol, s_address, s_port, m_port):
         if s_port is not None:
             self._port = int(s_port)
         self._management_port = int(m_port)
-        self._status = 1
+        self._status = ServiceRecord.Status.Running
```
praveen-garg: even for inner class? is it not accessible as `Status.Running` itself

python/foglamp/common/service_record.py
```diff
@@ -41,7 +53,8 @@ def __init__(self, s_id, s_name, s_type, s_protocol, s_address, s_port, m_port):
...
+        self._check_count = 1
```
praveen-garg: Nope

python/foglamp/services/core/server.py
```diff
@@ -654,9 +654,13 @@ def stop(cls, loop=None):
             found_services = ServiceRegistry.get()
             services_to_stop = list()
 
+            from foglamp.common.service_record import ServiceRecord
```
praveen-garg: No! As discussed earlier it must be very very uncommon case. Let's not make it a practice. Figure out and refactor to remove it

python/foglamp/services/core/service_registry/monitor.py
```diff
@@ -31,60 +30,61 @@ class Monitor(object):
...
 
-    _logger = None  # type: logging.Logger
+    _DEFAULT_MAX_ATTEMPTS = 15
 
-    def __init__(self):
+    _logger = None
 
-        self._logger = logger.setup("SMNTR", level=logging.INFO)
+    def __init__(self):
+        self._logger = logger.setup(__name__, level=20)
```
praveen-garg: logging.INFO is far better than using 20

python/foglamp/common/service_record.py
amarendra-dianomic: Right. "Unregistered" changed to "Stopped".

python/foglamp/common/service_record.py
amarendra-dianomic: Nah. We need to keep the ok count with each service and there is no other proper logical location. Otherwise, we will have to maintain a separate reference list of check_counts, thereby making code error prone and adding unneccessary redundancy.

python/foglamp/common/service_record.py
amarendra-dianomic: Nah. It is not accessible via Status.Running. It has to be ServiceRecoord.Status.Running.

python/foglamp/common/service_record.py
amarendra-dianomic: Why not?

python/foglamp/services/core/server.py
amarendra-dianomic: Line moved to be with other import statements. Is this you meant?

python/foglamp/services/core/service_registry/monitor.py
amarendra-dianomic: logging.INFO is posing some issues if used with "from foglamp.common import logger". Do not know why. Can you pl check this at your end?

praveen-garg:
We need to see, a huge re-work is not needed as we need `GET /foglamp/status - return the status of FogLAMP and its microservice components
`  `Monitoring`  via REST API

python/foglamp/common/service_record.py
amarendra-dianomic: Via commit https://github.com/foglamp/FogLAMP/pull/685/commits/f15be0d7674ccb6cea4e1a8550f44c7428cc4335, a separate dict is now being maintained to hold check counts and _check_count now removed from service_record.

amarendra-dianomic:
@MarkRiddoch @praveen-garg @ashwinscale @Singhal-Vaibhav feedback changes complete. Pl check.

Singhal-Vaibhav:
@amarendra-dianomic As per the design docs, the status of the service can be running/down/unresponsive instead of numbers 1-4.

amarendra-dianomic:
@MarkRiddoch in light of feedback from @Singhal-Vaibhav, please advise the correct naming for below status:
```
    class Status(IntEnum):
        """Enumeration for Service Status"""

        Running = 1
        Stopped = 2
        Failed = 3
        Doubtful = 4

```

amarendra-dianomic:
@Singhal-Vaibhav @praveen-garg tests fixed, status displayed in small letters, status text changed per doc. Pl check.

PR #686, opened 2018-03-29
Fogl 1213 - password rotation policy

ashish-jabble:
**Fixed items:**
- [x] configuration entry **passwordChange** in the **rest_api** category 
- [x] **pwd_last_changed** column to the user table 
- [x] **password_history** table added for maintain the history of password
- [x] login code to check the password age at login and if the password is older than the number of days specified in configuration
- [x] Added a new API entry **PUT** /foglamp/user/{username}/password 
- [x] test_user_model.py `collected 50 items == 49 passed, 1 skipped in 0.41 seconds ==` with Code coverage **98%**
- [x] test_auth_optional.py `collected 38 items === 38 passed in 2.48 seconds ===`
- [x] test_auth_mandatory.py `collected 62 items == 62 passed in 4.55 seconds ==`
       with code coverage **97%**
- [x] Admin should be able to continue to reset user's password. In this case Admin does not need to know the current password of user in context - child [PR#688](https://github.com/foglamp/FogLAMP/pull/688) -- `Once approved will update auth  tests`


python/foglamp/services/core/user_model.py
```diff
@@ -142,16 +151,43 @@ def update(cls, user_id, user_data):
...
+
+                    if 'password' in user_data:
+                        # delete oldest password for user, as storage result in sorted order so its safe to delete its last index from pwd_history_list
+                        if len(pwd_history_list) >= USED_PASSWORD_HISTORY_COUNT:
+                            payload = PayloadBuilder().WHERE(['user_id', '=', user_id]).AND_WHERE(['pwd', '=', pwd_history_list[-1]]).payload()
+                            storage_client.delete_from_tbl("user_pwd_history", payload)
+
+                        # insert into password history table
+                        payload = PayloadBuilder().INSERT(user_id=user_id, pwd=hashed_pwd).payload()
+                        storage_client.insert_into_tbl("user_pwd_history", payload)
```
praveen-garg: please move to a separate def update_user_password for better maintenance

praveen-garg:
I believe we should get https://github.com/foglamp/FogLAMP/pull/688 into it first. so that `/admin/` based routes are for user management only;

ashish-jabble:
Everything is fixed and up now. Please review and merge!

python/foglamp/services/core/api/auth.py
```diff
@@ -257,40 +268,104 @@
...
+        raise web.HTTPBadRequest(reason=PASSWORD_ERROR_MSG)
+
+    if current_password == new_password:
+        msg = "New password should not be same as current password"
+        _logger.warning(msg)
+        raise web.HTTPBadRequest(reason=msg)
+
+    user_id = User.Objects.is_user_exists(username, current_password)
+    if not user_id:
+        msg = 'Username or Password do not match'
```
praveen-garg: Just return `Invalid current password`

PR #688, opened 2018-03-30
Admin API end point added for user reset role and password

ashish-jabble:
- [x] reset => foglamp/admin/{user_id}/reset ( Admin  route to reset Password and role for a user)

- [x] delete_user => /foglamp/admin/{user_id}/delete (Only Admin can delete the user) 

- [x] create_user => /foglamp/admin/user (Only Admin can create the user) 

- [x] update_password => foglamp/user/{username}/password (Normal user can update the password)

- [ ] update_user =>  /foglamp/user/{id} (For normal user can update the profile) - Right now not implemeted as FOGL-1226 pending

**NOTE** -- will update tests once approved :)

### DIFF
--- a/python/foglamp/services/core/api/auth.py
+++ b/python/foglamp/services/core/api/auth.py
@@ -354,7 +354,7 @@ async def reset(request):
         _logger.warning(msg)
         raise web.HTTPBadRequest(reason=msg)
 
-    if not is_valid_role(role_id):
+    if role_id and not is_valid_role(role_id):
         _logger.warning("Create user requested with bad role id")
         return web.HTTPBadRequest(reason="Invalid or bad role id")
 
@@ -380,6 +380,10 @@ async def reset(request):
         msg = "User with id:<{}> does not exist".format(int(user_id))
         _logger.warning(msg)
         raise web.HTTPNotFound(reason=msg)
+    except User.PasswordAlreadyUsed:
+        msg = "The new password should be different from previous 3 used"
+        _logger.warning(msg)
+        raise web.HTTPBadRequest(reason=msg)
     except Exception as exc:
         _logger.exception(str(exc))
         raise web.HTTPInternalServerError(reason=str(exc))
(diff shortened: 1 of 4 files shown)


python/foglamp/services/core/api/auth.py
```diff
@@ -371,6 +326,64 @@
...
+    if password and not isinstance(password, str):
+        _logger.warning(PASSWORD_ERROR_MSG)
+        raise web.HTTPBadRequest(reason=PASSWORD_ERROR_MSG)
+    if password and not re.match(PASSWORD_REGEX_PATTERN, password):
+        _logger.warning(PASSWORD_ERROR_MSG)
+        raise web.HTTPBadRequest(reason=PASSWORD_ERROR_MSG)
+
+    user_data = {}
+    if 'role_id' in data:
+        user_data.update({'role_id': data['role_id']})
```
praveen-garg: what if a bad role id (non existent) is passed

python/foglamp/services/core/api/auth.py
```diff
@@ -32,6 +32,8 @@
     
     | POST                       | /foglamp/login                                      |
     | PUT                        | /foglamp/{user_id}/logout                           |
+    
+    | PUT                        | /foglamp/admin/{user_id}/reset                      |
```
praveen-garg: please add /admin for create and delete user too

python/foglamp/services/core/api/auth.py
```diff
@@ -371,6 +327,68 @@
...
+    password = data.get('password')
+    role_id = data.get('role_id')
+
+    if not role_id and not password:
+        msg = "Nothing to update the user"
+        _logger.warning(msg)
+        raise web.HTTPBadRequest(reason=msg)
+
+    if not is_valid_role(role_id):
+        _logger.warning("Create user requested with bad role id")
```
praveen-garg: create user? its for reset password or update role here

praveen-garg: what if no role_id passed; the check above only handles AND case

ashish-jabble:
@praveen-garg made changes as per your suggestions.

PR #696, opened 2018-04-09
FOGL-1241 removed +00:00 UTC representation for user and user_login tables in sqlite and postgres will remain as with tz

ashish-jabble:
- [x] python tests -- all pass [=== 49 passed, 1 skipped in 0.61 seconds ===]
- [x] schema changes without +00.00 for sqlite storage plugins (removed only for auth specific stuff)
- [x] authentication is working with both plugins against this branch (manually verified)

**NOTE**: We need some deep testing to make sure authentication work would not be affected.

### DIFF
--- a/python/foglamp/services/core/user_model.py
+++ b/python/foglamp/services/core/user_model.py
@@ -255,8 +255,8 @@ class User:
             """
 
             storage_client = connect.get_storage()
-            token_expiration_format = '{"column": "token_expiration", "format": "YYYY-MM-DD HH24:MI:SS.MS", "alias" : "token_expiration"}'
-            payload = PayloadBuilder().SELECT(token_expiration_format).WHERE(['token', '=', token]).payload()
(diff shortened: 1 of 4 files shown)


pintomax:
I think we can also work in Python only using the rest API features:
            payload = PayloadBuilder().SELECT('{"column": "token_expiration", "format": "YYYY-MM-DD HH24:MI:SS.MS", "alias" : "token_expiration"}').WHERE(['token', '=', token]).payload()

            payload = PayloadBuilder().SELECT("pwd", "id", "role_id", '{"column": "pwd_last_changed", "format": "YYYY-MM-DD HH24:MI:SS.MS", "alias" : "pwd_last_changed"}').WHERE(['uname', '=', username]).AND_WHERE(['enabled', '=', 't']).payload()

This select a date in a proper format for both engines

pintomax:
@ashish-jabble with new change that uses payload builder, do we still need to change the schema in init.sql files?

ashish-jabble:
@pintomax bigger question is, do we support timezone at db level or not;  we need to be consistent with storage plugins DDL and client code?

And saving `+00:00` into DB  does not make any sense to me atleast.

PR #698, opened 2018-04-10
FOGL-1233 add config item REST  end-point added

ashish-jabble:
- [x] add config item REST end point added
- [x] tests added and all are passed + no regression for existing ones
- [x] 100% code coverage for the feature

### DIFF
--- a/python/foglamp/services/core/api/configuration.py
+++ b/python/foglamp/services/core/api/configuration.py
@@ -7,6 +7,8 @@
 from aiohttp import web
 from foglamp.services.core import connect
 from foglamp.common.configuration_manager import ConfigurationManager
+from foglamp.common.storage_client.payload_builder import PayloadBuilder
+from foglamp.common.audit_logger import AuditLogger
 
 __author__ = "Amarendra K. Sinha, Ashish Jabble"
 __copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
@@ -19,6 +21,7 @@ _help = """
     | GET             | /foglamp/category/{category_name}                         |
     | GET PUT         | /foglamp/category/{category_name}/{config_item}           |
     | DELETE          | /foglamp/category/{category_name}/{config_item}/value     |
+    | POST            | /foglamp/category/{category_name}/{add_config_item}       |
     -------------------------------------------------------------------------------
 """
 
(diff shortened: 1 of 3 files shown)


python/foglamp/services/core/api/configuration.py
```diff
@@ -19,6 +21,7 @@
     | GET             | /foglamp/category/{category_name}                         |
     | GET PUT         | /foglamp/category/{category_name}/{config_item}           |
     | DELETE          | /foglamp/category/{category_name}/{config_item}/value     |
+    | POST            | /foglamp/category/{category_name}/{add_config_item}       |
```
praveen-garg: add_config_item > config_item

python/foglamp/services/core/api/configuration.py
```diff
@@ -178,6 +181,75 @@
...
+
+    Returns:
+        Json response with message key
+
+    :Example:
+        curl -d '{"default": "true", "description": "Test description", "type": "boolean"}' -X POST https://localhost:1995/foglamp/category/{new_config_item} --insecure
+        curl -d '{"default": "true", "description": "Test description", "type": "boolean", "value": "false"}' -X POST https://localhost:1995/foglamp/category/test_cat/{new_config_item} --insecure
+    """
+    category_name = request.match_info.get('category_name', None)
+    new_config_item = request.match_info.get('add_config_item', None)
```
praveen-garg: per route item

python/foglamp/services/core/api/configuration.py
```diff
@@ -178,6 +181,75 @@
...
+        storage_client = connect.get_storage()
+        cf_mgr = ConfigurationManager(storage_client)
+
+        data = await request.json()
+        if not isinstance(data, dict):
+            raise ValueError('Data payload must be a dictionary')
+
+        # if value key is in data then go ahead with data payload and validate
+        # else update the data payload with value key and set its value to default value and validate
+        val = data.get('value', None)
```
praveen-garg: what if default is also missing?

python/foglamp/services/core/api/configuration.py
```diff
@@ -178,6 +181,75 @@
...
+        if val is None:
+            temp_dict = data
+            temp_dict.update({'value': temp_dict.get('default')})
+            config_item_dict = {new_config_item: temp_dict}
```
praveen-garg: can't we do `data.update({ 'value': data.get('default')})`

python/foglamp/services/core/api/configuration.py
```diff
@@ -178,6 +181,75 @@
...
+        else:
+            config_item_dict = {new_config_item: data}
+
+        # validate configuration category value
+        await cf_mgr._validate_category_val(category_val=config_item_dict, set_value_val_from_default_val=False)
```
praveen-garg: will not the `if val is None:` case will be handle via internal code itself? ` set_value_val_from_default_val=False => True

python/foglamp/services/core/api/configuration.py
```diff
@@ -178,6 +181,75 @@
...
+
+        # validate category
+        category = await cf_mgr.get_category_all_items(category_name)
+        if category is None:
+            raise NameError("No such Category found for {}".format(category_name))
```
praveen-garg: check this before `_validate_category_val`

python/foglamp/services/core/api/configuration.py
```diff
@@ -178,6 +181,75 @@
...
+        payload = PayloadBuilder().SET(value=merge_cat_val).WHERE(["key", "=", category_name]).payload()
+        result = storage_client.update_tbl("configuration", payload)
+        response = result['response']
+
+        # logged audit new config item for category
+        audit = AuditLogger(storage_client)
+        audit_details = {'category': category_name, 'item': new_config_item, 'value': config_item_dict}
+        await audit.information('CONAD', audit_details)
+
+        return web.json_response({"message": "{} config item has been saved for {} category".format(new_config_item, category_name)})
```
praveen-garg: just for better readability keep it as last line of def

python/foglamp/services/core/api/configuration.py
ashish-jabble: value set to None and thereafter validate_category_val handles the default key like missing key with 400 http status code

python/foglamp/services/core/api/configuration.py
ashish-jabble: The idea is to not to update original `data`, but yes we can do it, if no usage further

python/foglamp/services/core/api/configuration.py
ashish-jabble: Nope, it will not set as `value == default` if True as it validates only

python/foglamp/services/core/api/configuration.py
ashish-jabble: Nope, category check DB call after the validate_config_item keys

praveen-garg: gotcha! you want to avoid storage call, if supplied config item data is not valid.

PR #703, opened 2018-04-11
FOGL-1249 Use storage --plugin in scripts to find the working storage plugin

izoratti:


scripts/common/get_engine_management.sh
```diff
@@ -17,20 +17,20 @@
...
 
-get_engine_management() {
 
-    # Remove new lines and store in a variable
-    json_foglamp=`tr -d '\n' < $FOGLAMP_DATA/etc/foglamp.json`
+get_engine_management() {
 
-    # Remove tabs
-    json_foglamp=`echo $json_foglamp | tr -d '\t'`
+  storage_info=`$FOGLAMP_ROOT/scripts/services/storage --plugin`
```
MarkRiddoch: Easier way to do this is to make storage_info an array
storage_info=(`$FOGLAMP_ROOT/scripts/services/storage --plugin`)
storage=${storage_info[0]}
managed=${storage_info[1]}

saves running echo and cut twice each

scripts/storage
```diff
@@ -38,29 +38,11 @@ storage_log() {
...
-    PLUGIN_TO_USE=`get_plugin_from_storage "$FOGLAMP_DATA/etc/storage.json"`
-
-    # Check if the plugin has been found
-    if [[ "${#PLUGIN_TO_USE}" -eq 0 ]]; then
-        storage_log "err" "Missing plugin in FogLAMP storage cache file storage.json" "all" "pretty"
-    fi
-
-fi
 
+# Check if the storage cache file exists
```
MarkRiddoch: Not sure why this comment

scripts/plugins/storage/sqlite.sh
```diff
@@ -344,7 +340,7 @@ case "$engine_management" in
 
         # Unexpected value from the configuration file
         sqlite_log "err" "FogLAMP cannot start." "all" "pretty"
-        sqlite_log "err" "Missing plugin information in FogLAMP configuration file foglamp.json" "all" "pretty"
+        sqlite_log "err" "Missing plugin information the storage microservice" "all" "pretty"
```
pintomax: add "from":

Missing plugin information the storage microservice

===>

Missing plugin information from the storage microservice

izoratti: Fixed - thanks!

scripts/storage
izoratti: It was a leftover - fixed - thanks!

scripts/common/get_engine_management.sh
izoratti: that would be:

storage_info=( $($FOGLAMP_ROOT/scripts/services/storage --plugin) )

PR #705, opened 2018-04-12
FOGL-1252 no-cache-dir flag added for python-requirement target with/out user

ashish-jabble:
- [x] make
- [x] sudo make
- [x] sudo make install

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -211,11 +211,11 @@ python_build : $(PYTHON_SETUP_FILE)
 
 # install python requirements without --user 
 python_requirements : $(PYTHON_REQUIREMENTS_FILE)
-	$(PIP_INSTALL_REQUIREMENTS) $(PYTHON_REQUIREMENTS_FILE)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -211,11 +211,11 @@ python_build : $(PYTHON_SETUP_FILE)
 
 # install python requirements without --user 
 python_requirements : $(PYTHON_REQUIREMENTS_FILE)
-	$(PIP_INSTALL_REQUIREMENTS) $(PYTHON_REQUIREMENTS_FILE)
+	$(PIP_INSTALL_REQUIREMENTS) $(PYTHON_REQUIREMENTS_FILE) --no-cache-dir
```
praveen-garg: do we need this for `sudo` too?

ashish-jabble: issue is reproducible via both`sudo vs normal`, if we need no-cache dir flag then it should be for both targets. As it downloads the dependencies everytime.

PR #706, opened 2018-04-12
FOGL-1254 - Fixed smoke test to work with HTTPS

izoratti:


python/foglamp/services/core/server.py
```diff
@@ -115,10 +115,10 @@ class Server:
     rest_server_port = 0
     """ FogLAMP REST API port """
 
-    is_rest_server_http_enabled = False
+    is_rest_server_http_enabled = True
```
praveen-garg: Not needed, as when server starts it calls def `rest_api_config` (L#278) but yes for better readability we can.

python/foglamp/services/core/server.py
```diff
@@ -115,10 +115,10 @@ class Server:
...
     """ a Flag to decide to enable FogLAMP REST API on HTTP on restart """
 
-    is_auth_required = True
+    is_auth_required = False
```
praveen-garg: Not needed, as when server starts it calls def rest_api_config (L#295) but yes for better readability we can.

python/foglamp/services/core/server.py
```diff
@@ -138,7 +138,7 @@ class Server:
         'enableHttp': {
             'description': 'Enable or disable the connection via HTTP',
             'type': 'boolean',
-            'default': 'false'
+            'default': 'true'
```
praveen-garg: Yes, that is the real change required to make Admin REST API to start with http as default! 👍 

Agree! To avoid confusion, this key should be renamed to enableHttp**s**

tests/system/suites/smoke/e/002_start.expected
```diff
@@ -1 +1 @@
-FogLAMP running.
+FogLAMPvrunning
```
praveen-garg: typo?

tests/system/suites/smoke/t/002_start.test
```diff
@@ -2,6 +2,8 @@
 
 $TEST_BASEDIR/bash/exec_any_foglamp_command.bash start > /dev/null 2>&1
 $TEST_BASEDIR/bash/check_foglamp_status.bash > ${RESULT_DIR}/$TEST_NAME.temp 2>&1
-head -n1 ${RESULT_DIR}/$TEST_NAME.temp
+
+# Grab the 1st line and remove all digits (version ¢), spaces and dots
```
praveen-garg: what is  this foreign character? `¢`

tests/system/tests/bash/count_assets_https.bash
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+curl -sk https://localhost:1995/foglamp/asset | jq -S '.'
```
praveen-garg: now we should use http and hence `-k`

recommendation: when https intstead of -k, `--insecure` adds more readability

tests/system/tests/bash/read_an_asset_https.bash
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+curl -sk https://localhost:1995/foglamp/asset/$1
```
praveen-garg: as above

tests/system/suites/smoke/e/002_start.expected
izoratti: No typo, that is the expected result.

tests/system/tests/bash/count_assets_https.bash
izoratti: This is the https script, so it should use https: and -k.
-k is commonly used, --insecure has been introduced later and pretty unknown. I could say that it would add more uncertainty and developers would have to check on man what that means (as I did, because I had no idea it is the same of -k).

tests/system/tests/bash/read_an_asset_https.bash
izoratti: same as above.

PR #710, opened 2018-04-13
FOGL-1201 - Code refactored and minor fixes for foglamp server stop error

amarendra-dianomic:
Enabled all schedules and tested OK vide:
- "scripts/foglamp start" and "scripts/foglamp stop"
- "python3 -m foglamp.services.core" and "scripts/foglamp stop"
- "scripts/foglamp start" and stopping via " curl -X PUT http://localhost:8081/foglamp/shutdown"
- "scripts/foglamp start" and stopping via "curl -X POST http://localhost:<mport\>/foglamp/service/shutdown"

Test result:
`=================================================== 1136 passed, 13 skipped in 50.21 seconds ====================================================
`


amarendra-dianomic:
Latest fixed required for ubuntu 17.04.

python/foglamp/services/core/api/common.py
```diff
@@ -89,7 +89,7 @@ def get_stats(k):
 
     try:
         loop = request.loop
-        loop.call_later(2, do_shutdown, loop)
```
praveen-garg: so now we don't need  L#91 `loop = request.loop`! Yes?

python/foglamp/services/core/api/common.py
```diff
@@ -99,9 +99,10 @@ def get_stats(k):
...
-def do_shutdown(loop=None):
+def do_shutdown(request):
     _logger.info("Executing controlled shutdown")
-    if loop is None:
-        loop = asyncio.get_event_loop()
-    loop.run_until_complete(server.Server._stop())
-    loop.stop()
+    try:
+        loop = request.loop
+        loop.run_until_complete(server.Server.shutdown(request))
```
praveen-garg: we should fix `shutdown` too, To use request.loop instead of `loop = asyncio.get_event_loop()` and in for async sleep there use same loop `await asyncio.sleep(2.0, loop= <request.loop>)`

python/foglamp/services/core/api/common.py
```diff
@@ -99,9 +99,10 @@ def get_stats(k):
...
+    except RuntimeError as ex:
+        _logger.exception("Error while stopping foglamp server: {}".format(str(ex)))
```
praveen-garg: foglamp -> FogLAMP
RuntimeError as `ex` > RuntimeError as `e`

python/foglamp/services/core/server.py
```diff
@@ -617,10 +617,13 @@ def start(cls):
...
-            loop = asyncio.get_event_loop()
-        loop.run_until_complete(cls._stop())
-        loop.stop()
+        try:
+            if loop is None:
+                loop = asyncio.get_event_loop()
+            loop.run_until_complete(cls._stop())
+            loop.stop()
+        except RuntimeError as ex:
+            _logger.exception("Error while stopping foglamp server: {}".format(str(ex)))
```
praveen-garg: foglamp -> FogLAMP
RuntimeError as `ex` > RuntimeError as `e`

praveen-garg: I think we should also raise, not only just log; this is why _stop has bare exception raised!

python/foglamp/services/core/api/common.py
praveen-garg: I think we should also raise, not only just log! _stop has bare exception raise for the same purpose

amarendra-dianomic:

python/foglamp/services/core/api/common.py
praveen-garg: ah! we are using it! I was looking into diff only.

PR #711, opened 2018-04-16
FOGL-1122 Tests added for service.py

ashish-jabble:
**Code Coverage 100%** 
`73 statements   73 run 0 missing 0 excluded`

**Tests O/P**
`=== 960 passed, 12 skipped in 51.13 seconds ===`

### DIFF
--- a/python/foglamp/services/core/api/service.py
+++ b/python/foglamp/services/core/api/service.py
@@ -20,8 +20,7 @@ __version__ = "${VERSION}"
 
 _help = """
     -------------------------------------------------------------------------------
-    | POST            | /foglamp/service                                          |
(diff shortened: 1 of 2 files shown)


python/foglamp/services/core/api/service.py
```diff
@@ -72,6 +72,8 @@ def get_service_records():
 
     try:
         data = await request.json()
+        if not isinstance(data, dict):
+            raise ValueError('Data payload must be a dictionary')
```
MarkRiddoch: This is not a very meaningful error to give a user. They are using a REST API and expressing the payload in JSON so the message should really be a message about JSON.

PR #714, opened 2018-04-16
Fogl 1259 - unit test for SendingProcess

sstefano69:
Ready to be reviewed and eventually merged.


amarendra-dianomic:
@sstefano69 IMHO, it will be better if you write unit tests post incorporating proposed source code changes and refactoring such as extending the class from FoglampProcess class.

sstefano69:
@amarendra-dianomic , the current request/direction is to focus on having unit tests on the current code.

PR #719, opened 2018-04-17
Fogl 1260

ashish-jabble:
- [x] tests server file move to right place
- [x] interest registry
- [x] service registry
- [x] ping
- [ ] shutdown
- [x] change -- PASS
- [x] removed some dead code/unreachable code from delete endpoints and other code refactoring

**Tests O/P**
`=== 989 passed, 13 skipped in 56.39 seconds ==`

Code coverage upto **49%**
`549 statements   268 run 281 missing 0 excluded`


tests/unit/python/foglamp/services/core/test_server.py
```diff
@@ -0,0 +1,435 @@
...
+__author__ = "Vaibhav Singhal, Ashish Jabble"
+__copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
+__license__ = "Apache 2.0"
+__version__ = "${VERSION}"
+
+
+@pytest.allure.feature("unit")
+@pytest.allure.story("api", "server")
+class TestServer:
+    """Tests the calls to configuration manager via core management api
```
praveen-garg: Group endpoints test into sections and move this comment, under

```
############################
### Configuration Management
############################
```

tests/unit/python/foglamp/services/core/test_server.py
```diff
@@ -0,0 +1,435 @@
...
+        sname = "name"
+        stype = "Southbound"
+        sprotocol = "http"
+        saddress = "localhost"
+        sport = 1234
+        smgtport = 4321
+        data = []
+        record = ServiceRecord(service_id, sname, stype, sprotocol, saddress, sport, smgtport)
+        data.append(record)
+        Server._storage_client = MagicMock(StorageClient)
```
praveen-garg: as its used now again, can not we have a resuable def  to return service_records?

tests/unit/python/foglamp/services/core/test_server.py
```diff
@@ -0,0 +1,435 @@
...
+    async def test_ping(self, client):
+        resp = await client.get('/foglamp/service/ping')
+        assert 200 == resp.status
+        r = await resp.text()
+        json_response = json.loads(r)
+        assert 'uptime' in json_response
+        assert 0.0 < json_response["uptime"]
+
+    # FIXME: Not sure if we need this
+    async def test_change(self, client):
```
praveen-garg: I don;t think its required. as Server is not implemented with FoglampMicroservice abstract class; but we can keep it for now as it, and the test the same way.

tests/unit/python/foglamp/services/core/test_server.py
```diff
@@ -0,0 +1,435 @@
...
+        pass
+
+    # TODO: tricky one
+    async def test_shutdown(self, client):
```
praveen-garg: let me have a look, but this PR don't need to wait for this.

PR #724, opened 2018-04-19
FOGL-1265 delete configuration response fixes

ashish-jabble:
1) curl -X GET http://localhost:8081/foglamp/category/rest_api/authentication 
```
{
  "description": "To make the authentication mandatory or optional for API calls",
  "type": "string",
  "default": "optional",
  "value": "optional"
}
```

2) curl -X PUT -H "Content-Type: application/json" -d '{"value": "mandatory"}' http://localhost:8081/foglamp/category/rest_api/authentication
```
{
  "description": "To make the authentication mandatory or optional for API calls",
  "type": "string",
  "default": "optional",
  "value": "mandatory"
}
```
3) curl -X DELETE http://localhost:8081/foglamp/category/rest_api/authentication/value 
```
{
  "description": "To make the authentication mandatory or optional for API calls",
  "type": "string",
  "default": "optional",
  "value": "optional"
}
```

**Tests O/P** - No regression with this change and code coverage for this file to now **100%**

`== 1000 passed, 13 skipped in 64.15 seconds ==`

### DIFF
--- a/python/foglamp/services/core/api/configuration.py
+++ b/python/foglamp/services/core/api/configuration.py
@@ -272,7 +272,11 @@ async def delete_configuration_item_value(request):
     # TODO: make it optimized and elegant
     cf_mgr = ConfigurationManager(connect.get_storage())
     try:
-        await cf_mgr.set_category_item_value_entry(category_name, config_item, '')
+        category_item = await cf_mgr.get_category_item(category_name, config_item)
+        if category_item is None:
+            raise ValueError
+
+        await cf_mgr.set_category_item_value_entry(category_name, config_item, category_item['default'])
     except ValueError:
         raise web.HTTPNotFound(reason="No detail found for the category_name: {} and config_item: {}".format(category_name, config_item))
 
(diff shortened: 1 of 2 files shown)


ashish-jabble:
Question:
1) If value is intentionally set to `empty string` while setting a configuration item, then what will be the implications? For example, if we set `rest_api/httpPort` value to "", the current code will set it to "" but this will throw an error after foglamp server is restarted. 

curl -X GET http://localhost:8081/foglamp/category/rest_api/httpPort 
```
{
  "description": "The port to accept HTTP connections on",
   "type": "integer",
   "default": '8081",
   "value": "8081"
}
```

curl -X PUT -H "Content-Type: application/json" -d '{"value": ""}' http://localhost:8081/foglamp/category/rest_api/httpPort
```
{
  "description": "The port to accept HTTP connections on",
   "type": "integer",
   "default": '8081",
   "value": ""
}
```
Therefore in this case also we should check for empty strings while setting a config item value and set it to default if an empty string is supplied.

MarkRiddoch:
This looks good. As for the question about an empty string, if the user (of the API) requests the value to be set as an empty string then we do exactly that and set it to an empty string as we would with any other value.  If they then go on to delete the entry it will revert to the default, if a default is provided. If there is no default then delete will always set the value to be empty regardless of what the value currently is.

praveen-garg:
@MarkRiddoch 
So here we can simply say reset to default on delete?  so DELETE == RESET in true sense?

What if user had updated item value to be empty and now take delete action? will it reset to default or return `200: OK | Value is already set to empty.`
As default value always exists and so is the value on create category. the path to set empty is to update with empty string.

To align with specs:

> This will result in the value being returned to the default value if one is defined. If not the value will be blank, i.e. the value property of the JSON object will exist with an empty value.

PR #749, opened 2018-04-27
FOGL-1287 schedule repeat interval fixed for stats and schedule

ashish-jabble:
Fix works with both storage engines
- [x] postgres
- [x] sqlite

**Tests O/P**
`=== 1093 passed, 13 skipped in 56.61 seconds ===`

### DIFF
--- a/python/foglamp/services/core/api/statistics.py
+++ b/python/foglamp/services/core/api/statistics.py
@@ -9,7 +9,7 @@ from aiohttp import web
 
 from foglamp.common.storage_client.payload_builder import PayloadBuilder
 from foglamp.services.core import connect
-
+from foglamp.services.core.scheduler.scheduler import Scheduler
 
 __author__ = "Amarendra K. Sinha, Ashish Jabble"
 __copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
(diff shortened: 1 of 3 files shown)


amarendra-dianomic:
@MarkRiddoch, @izoratti, @pintomax of late, we have noticed that there has been a series of requirements that calls for adjustments in python code to suit Sqlite plugin as below:
a) [PR#672](https://github.com/foglamp/FogLAMP/pull/672) to adjust schedule_day
b) FOGL-1287 for adjusting "repeat" col resulting in this PR 
c) FOGL-1292 for adjustment of timezone

There is no surety that there will not be any more need for such tweaking. Later, if we decide to add support for more DB engines such as MongoDB etc, would we be doing all such tweaking for newer DBs, should a need arise, in Scheduler code and other Python code?

Therefore, I am of very strong opinion that all such storage plugin related adjustments MUST go into storage layer code and NOT in any client code under any circumastances.

MarkRiddoch:
@amarendra-dianomic The storage layer should hide the difference between storage systems that we use and hence the code outside of the storage layer should not change. This is the goal, but the reality is that the first implementation was based on Postgres and there are areas where it reflected too closely what postgres does. I agree that longer term changes will not be made to satisfy particular storage plugins, and certainly there should never be code that is aware of which plugin is in use outside of the plugin, but in the short term we may have situations where we discover a made choice was made and we have tweak the interface to make it more generic or to resolve an issue where the interface was too closely tied to a particular storage engine. This should become more and more rare as we introduce the third and fourth storage plugins.

PR #750, opened 2018-04-30
Sqlite specific code removed + some minor refactoring

amarendra-dianomic:
Test result on postgres:
`=================================================== 1072 passed, 13 skipped in 49.72 seconds ====================================================
`

Test result on sqlite:
`=================================================== 1072 passed, 13 skipped in 49.72 seconds ====================================================
`


praveen-garg:
could we please update the tests too? I think we need works if  x := 1 and expect fail if x := "1"

amarendra-dianomic:
@praveen-garg tests fixed. Please check.

PR #756, opened 2018-05-01
FOGL-641 - Timebucket support added to Payload Builder

amarendra-dianomic:
Test result:
`=================================================== 1110 passed, 13 skipped in 51.43 seconds ====================================================
`

### DIFF
--- a/python/foglamp/common/storage_client/payload_builder.py
+++ b/python/foglamp/common/storage_client/payload_builder.py
@@ -566,9 +566,9 @@ class PayloadBuilder(object):
         Forms a json to return a dict of timebucket col
 
         :param timestamp: timestamp col
-        :param size: size
-        :param fmt: format string
-        :param alias: alias
+        :param size: bucket size in seconds, defaults to "1"
+        :param fmt: format string, optional
+        :param alias: alias, optional
         :return:
         :example:
         PayloadBuilder().TIMEBUCKET("user_ts", "5").payload() returns
--- a/python/foglamp/common/storage_client/payload_builder.py
+++ b/python/foglamp/common/storage_client/payload_builder.py
@@ -561,7 +561,7 @@ class PayloadBuilder(object):
         return cls
 
     @classmethod
-    def TIMEBUCKET(cls, timestamp, size, fmt=None, alias=None):
+    def TIMEBUCKET(cls, timestamp, size="1", fmt=None, alias=None):
         """
         Forms a json to return a dict of timebucket col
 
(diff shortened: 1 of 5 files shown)


tests/unit/python/foglamp/common/storage_client/data/payload_timebucket1.json
```diff
@@ -0,0 +1,6 @@
+{
+  "timebucket" : {
+                    "timestamp" : "user_ts",
+                    "size"      : "5"
```
praveen-garg: what is size here?!

amarendra-dianomic: >The time bucket directive allows a timestamp column to be defined, the size of each time bucket, in seconds, an optional date format for the timestamp written in the results and an optional alias for the timestamp property that is written.

>If no size element is present then the default time bucket size is 1 second.

python/foglamp/common/storage_client/payload_builder.py
```diff
@@ -560,6 +560,50 @@ def JSON_PROPERTY(cls, *args):
                     cls.query_payload["json_properties"] = [json_property]
         return cls
 
+    @classmethod
+    def TIMEBUCKET(cls, timestamp, size="1", fmt=None, alias=None):
+        """
+        Forms a json to return a dict of timebucket col
+
+        :param timestamp: timestamp col
+        :param size: size
```
praveen-garg: add docstring here too

python/foglamp/common/storage_client/payload_builder.py
```diff
@@ -560,6 +560,50 @@ def JSON_PROPERTY(cls, *args):
                     cls.query_payload["json_properties"] = [json_property]
         return cls
 
+    @classmethod
+    def TIMEBUCKET(cls, timestamp, size="1", fmt=None, alias=None):
```
praveen-garg: will it be a string? `size="1"`

MarkRiddoch: @amarendra-dianomic as per your questions...

I don't believe the size default needs to be dealt with in the payload builder. If there is no size in the payload the storage layer gets it will use the default value of 1.

In retrospect the size would probably be best specified as a number rather than a string. Currently the storage layer only accepts a string. I suggest we leave it as a string for now and look at make it accept a number later.

I think seconds is good enough for this, the idea was to have time bucket style readings, for sub-second values to make sense I think we would need to have a very high frequency sensor. Again let's leave it expressed in seconds until we see a need for sub-second resolution.

MarkRiddoch:
Thanks Ashish, using a size greater than the list, i.e. greater then 60 for seconds and minutes and greater than 24 for hours probably doesn't not make much sense. We should probably limit this in the API and have an error returned in these case, or at the least document this limitation in the short-term. 

I suspect the differences in the payload format is an accident of history between the asset browser and the storage layer. I suggest we leave as it is for now.

PR #761, opened 2018-05-02
FOGL-1298 fix to check if foglamp pid is valid upon startup

ashwinscale:


extras/scripts/foglamp.service
```diff
@@ -74,8 +74,13 @@ case "$1" in
     if [ $PID -eq 0 ]; then
       foglamp_start
     else
-      echo "FogLAMP already running [$PID]"
-      exit 0
+      if [ -z "`ps axf | grep ${PID} | grep -v grep`" ]; then
```
izoratti: I think this is fine, but to be clear it is possible there are false positives, it would be better to capture only the PID and then decide to rm the file. 
Example:
PID in file is 12
You find a row with PID (or another string) 1234 >>> it will be identified as FogLAMP running. It happened to me many times during the tests.
I would recommend to cut or awk the result first.

ashwinscale: Perhaps I should use ps -p [pid] and base it on the return value:

amg@ubuntuserver1604:~/FogLAMP$ ps -p '29870'
  PID TTY          TIME CMD
29870 pts/1    00:00:34 python3
amg@ubuntuserver1604:~/FogLAMP$ echo $?
0
amg@ubuntuserver1604:~/FogLAMP$ ps -p '29832'
  PID TTY          TIME CMD
amg@ubuntuserver1604:~/FogLAMP$ echo $?
1
amg@ubuntuserver1604:~/FogLAMP$ ps -p '298'
  PID TTY          TIME CMD
  298 ?        00:00:00 bioset
amg@ubuntuserver1604:~/FogLAMP$ echo $?
0

PR #769, opened 2018-05-07
Fogl 1297: Sqlite does not return machine time as postgres does

pintomax:


pintomax:
The timestamp "2018-05-07 08:15:17.401555+00:00" is added by fogbench and it's in UTC (+00:00)

The data returned it's taken as it is: if data is inserted with localtime( with now() for example) we get it back as it is accordingly to current time zone.

====

sqlite> update readings set user_ts = '2018-05-07 08:39:28.960735+05:30' where id = 200;
sqlite> select id, user_ts, datetime(user_ts) from readings where id = 200;
200|2018-05-07 08:39:28.960735+05:30|2018-05-07 03:09:28

curl '127.0.0.1:8081/foglamp/asset/fogbench%2Fwall%20clock' | jq

```
  {
    "timestamp": "2018-05-07 03:09:28.960",
    "reading": {
      "tick": "tock"
    }
  }
```

Singhal-Vaibhav:
@pintomax ,
Totally agree. This needs to be fixed at python side. Lets address that in FOGL-1313

Singhal-Vaibhav:
This also fixes FOGL-1242, @ashish-jabble Can you have a quick look?

pintomax:
@ashish-jabble This branch is only for SQLite storage plugin.

We might decide to modify Postgres init.sql later.

PR #800, opened 2018-05-17
FOGL-1345: fixed.

sstefano69:


sstefano69:
@ashish-jabble, consider that regardless of id you provide the last backup is going to be restored. 
it is a known situation.

PR #802, opened 2018-05-18
FOGL-1351: sending process: position read and updated changed

sstefano69:


sstefano69:
from 1.100 rows/second to 1.428 rows/sec

sstefano69:
Obsolete in favor of #804

PR #817, opened 2018-05-30
Fogl 1366: Convert the north Http plugin to Async

sstefano69:


praveen-garg:
@sstefano69  Please fix the tests for this specific plugin too; 
/cc @ashish-jabble  please review; as this is broken on develop.

python/foglamp/plugins/north/http_north/http_north.py
```diff
@@ -75,8 +75,13 @@ def plugin_init(data):
...
 
-def plugin_send(data, payload, stream_id):
-    return http_north.send_payloads(payload, stream_id)
+async def plugin_send(data, payload, stream_id):
+
+    is_data_sent, new_last_object_id, num_sent = await http_north.send_payloads(payload, stream_id)
+
+    return is_data_sent, new_last_object_id, num_sent
+
+
```
ashish-jabble: Remove above spaces

python/foglamp/plugins/north/http_north/http_north.py
```diff
@@ -75,8 +75,13 @@ def plugin_init(data):
     return config
 
 
-def plugin_send(data, payload, stream_id):
-    return http_north.send_payloads(payload, stream_id)
+async def plugin_send(data, payload, stream_id):
+
```
ashish-jabble: Do we need `data`, as it seems to be unused? And remove space line.

python/foglamp/plugins/north/http_north/http_north.py
sstefano69: done

python/foglamp/plugins/north/http_north/http_north.py
sstefano69: it is needed, 
the sign of the plugin_send call should be the same for every plugin.

PR #821, opened 2018-05-31
FOGL-1370: Intercepts aiohttp.client_exceptions.ClientPayloadError and handles it less allerting.

sstefano69:


python/foglamp/tasks/north/sending_process.py
```diff
@@ -512,13 +513,20 @@ def _is_north_valid(self):
...
+
+        converted_data = []
         try:
             # Loads data, +1 as > is needed
             readings = await self._readings.fetch(last_object_id + 1, self._config['blockSize'])
 
             raw_data = readings['rows']
             converted_data = self._transform_in_memory_data_readings(raw_data)
 
+        except aiohttp.client_exceptions.ClientPayloadError as _ex:
```
ashwinscale: This exception is thrown by the StorageClient. The code shows that the real exception is wrapped by a general Exception.
https://github.com/foglamp/FogLAMP/blob/adbe14e1261d66c64a39bc6a12ab31d4c88e8cb4/python/foglamp/common/storage_client/storage_client.py#L919

sstefano69: The code I saw in the 1370-repro was : 
        except Exception as ex:
            raise

btw, the new/latest code in Develop is no using anymore the try/catch, so it should be fine.

praveen-garg:
@sstefano69 is this issue still exists? are we able to reproduce on `1370-repro` when rebased with develop?

python/foglamp/tasks/north/sending_process.py
sstefano69: @ashwinscale , is this PR with you ?

PR #829, opened 2018-06-04
FOGL-1392: Fix unit test common/test_statistics.py

Singhal-Vaibhav:
Test results:

`=== 18 passed in 0.72 seconds ===`

Coverage:

`FogLAMP/python/foglamp/common/statistics.py : 100%`
`72 statements   72 run 0 missing 0 excluded`

### DIFF
--- a/tests/unit/python/foglamp/common/test_statistics.py
+++ b/tests/unit/python/foglamp/common/test_statistics.py
@@ -4,14 +4,13 @@
 # See: http://foglamp.readthedocs.io/
 # FOGLAMP_END
 
-import asyncio
 import json
 
 from unittest.mock import MagicMock, patch
 import pytest
 
-from foglamp.common.statistics import Statistics, _logger
-from foglamp.common.storage_client.storage_client import StorageClient
+from foglamp.common import statistics
+from foglamp.common.storage_client.storage_client import StorageClientAsync as StorageClient
 
 
 __author__ = "Ashish Jabble, Mark Riddoch, Vaibhav Singhal"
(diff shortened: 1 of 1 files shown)


tests/unit/python/foglamp/common/test_statistics.py
```diff
@@ -24,92 +23,108 @@
...
+
+        with patch.object(stats._storage, 'insert_into_tbl', return_value=mock_coro()) as stat_update:
+            await stats.register('T1Stat', 'Test stat')
+            args, kwargs = stat_update.call_args
+            assert args[0] == 'statistics'
+            expected_storage_args = json.loads(args[1])
+            assert expected_storage_args['key'] == 'T1Stat'
+            assert expected_storage_args['value'] == 0
+            assert expected_storage_args['previous_value'] == 0
+            assert expected_storage_args['description'] == 'Test stat'
```
ashish-jabble: Fix the indentation of stat_update assertions. Can't we use here 
`stat_update.assert_called_once_with("statistics", payload)` instead of individual args key-pair ?

tests/unit/python/foglamp/common/test_statistics.py
```diff
@@ -24,92 +23,108 @@
...
+        stats = statistics.Statistics(storageMock)
+        stats._registered_keys = []
+        async def mock_coro():
+            return {"response": "updated", "rows_affected": 1}
+
+        with patch.object(stats._storage, 'insert_into_tbl', return_value=mock_coro()) as stat_update:
+            await stats.register('T2Stat', 'Test stat')
+            await stats.register('T2Stat', 'Test stat')
+            assert stat_update.called
+            assert stat_update.call_count == 1
```
ashish-jabble: Fix the indentation of stat_update assertions. Also better to assert its payload too and stat_update var should be stat_insert or patch_insert_tbl

tests/unit/python/foglamp/common/test_statistics.py
```diff
@@ -123,54 +138,59 @@ def test_load_keys(self):
...
+        async def mock_coro():
+            return {"response": "not updated", "rows_affected": 0}
+
+        with patch.object(s._storage, 'update_tbl', return_value=mock_coro()) as stat_update:
+            with patch.object(statistics._logger, 'exception') as logger_exception:
+                with pytest.raises(KeyError):
+                    await s.add_update(stat_dict)
         args, kwargs = logger_exception.call_args
         assert args[0] == 'Statistics key %s has not been registered'
         assert args[1] == 'FOGBENCH/TEMPERATURE'
```
ashish-jabble: Fix the indentation of logger_exception assertions. Also add `stat_update` assertions

tests/unit/python/foglamp/common/test_statistics.py
Singhal-Vaibhav: Tried but it fails randomly because of order of json received

tests/unit/python/foglamp/common/test_statistics.py
Singhal-Vaibhav: payload verification with register is already done in test_register. The purpose of this test is to verify that with the same key, insert_into_tbl is called just once. Other concerns fixed.

PR #863, opened 2018-06-18
Removed South (Poll) Plugin (template) example from main repository

praveen-garg:
And removed refrences from db init scripts


praveen-garg:
This code will be moved to https://github.com/foglamp/FogLAMP-Examples

PR #867, opened 2018-06-19
FOGL-1453: add utility routines for ConfigCategories and ConfigCatego…

pintomax:
…ryDescription classes

FOGL-1453: add utility routines for ConfigCategories and ConfigCategoryDescription classes


pintomax:
New unit tests added

PR #871, opened 2018-06-20
FOGL-1446 Move dht11 from main repo to foglamp-south-dht11 

praveen-garg:
https://github.com/foglamp/foglamp-south-dht11


praveen-garg:
@MarkRiddoch  Please do an initial round of review for removal and corresponding (arm only debian) repository.

PR #878, opened 2018-06-26
FOGL-1479: New URL for OCS

sstefano69:


ashish-jabble:
@sstefano69 I think we need to update reference in system tests too i.e `tests/system/tests/python/common/ocs.py`
OCS_URL = "https://qi-data.osisoft.com" with "https://dat-a.osisoft.com"

PR #889, opened 2018-06-29
FOGL-1473 Asset browser query performance

MarkRiddoch:


scripts/plugins/storage/sqlite/upgrade/2.sql
```diff
@@ -0,0 +1,2 @@
+drop index foglamp.fki_readings_fk1;
```
pintomax: Remove "foglamp."

Postgres might need schema name, SQLite doesn't need it.

PR #890, opened 2018-06-29
FOGL-1496 - add_service() method needs sanity checks and rollback feature

amarendra-dianomic:


python/foglamp/services/core/api/service.py
```diff
@@ -95,42 +96,51 @@ def get_service_records():
         is_enabled = True if ((type(enabled) is str and enabled.lower() in ['t', 'true']) or (
             (type(enabled) is bool and enabled is True))) else False
 
-        storage = connect.get_storage_async()
+        # ---------------------------------------- Sanity checks
```
praveen-garg: Follow the style-guide please

python/foglamp/services/core/api/service.py
```diff
@@ -95,42 +96,51 @@ def get_service_records():
...
+        try:
+            # "plugin_module_path" is fixed by design. It is MANDATORY to keep the plugin in the exactly similar named
+            # folder, within the plugin_module_path.
+            plugin_module_path = "foglamp.plugins.south" if service_type == 'south' else \
+                                 "foglamp.plugins.north" if service_type == 'north' else None
+            import_file_name = "{path}.{dir}.{file}".format(path=plugin_module_path, dir=plugin, file=plugin)
+            _plugin = __import__(import_file_name, fromlist=[''])
+        except ImportError as ex:
+            raise web.HTTPInternalServerError(reason='Plugin "{}" import problem from path "{}". {}'.format(plugin, plugin_module_path, str(ex)))
+        except Exception as ex:
```
praveen-garg: we should check import error ONLY, this should be the parent exception block if any;

python/foglamp/services/core/api/service.py
```diff
@@ -95,42 +96,51 @@ def get_service_records():
...
+            raise web.HTTPInternalServerError(reason='Failed to create plugin configuration. {}'.format(str(ex)))
+
+        storage = connect.get_storage()
 
         # Check that the process name is not already registered
         payload = PayloadBuilder().SELECT("name").WHERE(['name', '=', name]).payload()
-        result = await storage.query_tbl_with_payload('scheduled_processes', payload)
+        result = storage.query_tbl_with_payload('scheduled_processes', payload)
```
praveen-garg: no await?

python/foglamp/services/core/api/service.py
```diff
@@ -95,42 +96,51 @@ def get_service_records():
...
         count = result['count']
         if count != 0:
             raise web.HTTPBadRequest(reason='A service with that name already exists')
 
         # Check that the schedule name is not already registered
         payload = PayloadBuilder().SELECT("schedule_name").WHERE(['schedule_name', '=', name]).payload()
-        result = await storage.query_tbl_with_payload('schedules', payload)
```
praveen-garg: no await?

python/foglamp/services/core/api/service.py
```diff
@@ -95,42 +96,51 @@ def get_service_records():
...
+        result = storage.query_tbl_with_payload('schedules', payload)
         count = result['count']
         if count != 0:
             raise web.HTTPBadRequest(reason='A schedule with that name already exists')
+        # ----------------------------------------
```
praveen-garg: remove this  line

python/foglamp/services/core/api/service.py
```diff
@@ -95,42 +96,51 @@ def get_service_records():
...
-        if service_type == 'south':
-            script = '["services/south"]'
-            plugin_module_path = "foglamp.plugins.south"
-        if service_type == 'north':
-            script = '["services/north"]'
-            plugin_module_path = "foglamp.plugins.north"
+        # Now first create the scheduled process entry for the new service
+        script = '["services/south"]' if service_type == 'south' else \
+                 '["services/north"]' if service_type == 'north' else None
         payload = PayloadBuilder().INSERT(name=name, script=script).payload()
```
praveen-garg: can script be None? better to check `if service_type.lower() is not in  [south, north]` then raise typeerror; else script = '["services/{}"]'.format(service_type.lower())

python/foglamp/services/core/api/service.py
```diff
@@ -95,42 +96,51 @@ def get_service_records():
...
         try:
-            res = await storage.insert_into_tbl("scheduled_processes", payload)
+            res = storage.insert_into_tbl("scheduled_processes", payload)
+        except StorageServerError as ex:
+            err_response = ex.error
+            raise web.HTTPInternalServerError(reason='Failed to created scheduled process. {}'.format(err_response))
         except Exception as ins_ex:
```
praveen-garg: please move as parent as we have  this at multiple places; and actually on generic exception aiohttp should capture/ return the same way even if you don't define. but yes for logging purpose keeping it as parent may be a valid idea!

python/foglamp/services/core/api/service.py
```diff
@@ -143,21 +153,44 @@ def get_service_records():
...
                                              keep_original_items=True)
         except ImportError as ex:
+            # Revert scheduled_process entry
+            payload = PayloadBuilder().WHERE(['name', '=', name]).payload()
+            result = storage.delete_from_tbl('scheduled_processes', payload)
             raise web.HTTPInternalServerError(reason='Plugin "{}" import problem from path "{}". {}'.format(plugin, plugin_module_path, str(ex)))
         except Exception as ex:
+            # Revert scheduled_process entry
+            payload = PayloadBuilder().WHERE(['name', '=', name]).payload()
+            result = storage.delete_from_tbl('scheduled_processes', payload)
```
praveen-garg: in both the exception block we have repeated code, can we have a def please?!

python/foglamp/services/core/api/service.py
```diff
@@ -143,21 +153,44 @@ def get_service_records():
...
+            schedule = StartUpSchedule()  # TODO: For North plugin also?
+            schedule.name = name
+            schedule.process_name = name
+            schedule.repeat = datetime.timedelta(0)
+            schedule.exclusive = True
+            schedule.enabled = False  # if "enabled" is supplied, it gets activated in save_schedule() via is_enabled flag
+
+            # Save schedule
+            await server.Server.scheduler.save_schedule(schedule, is_enabled)
+            schedule = await server.Server.scheduler.get_schedule_by_name(name)
```
praveen-garg: can we have a protected def please?

python/foglamp/services/core/api/service.py
```diff
@@ -143,21 +153,44 @@ def get_service_records():
...
+            result = storage.delete_from_tbl('scheduled_processes', payload)
+            raise web.HTTPInternalServerError(reason='Failed to created schedule. {}'.format(ex.error))
+        except Exception as ins_ex:
+            # Revert configuration entry
+            payload = PayloadBuilder().WHERE(['key', '=', name]).payload()
+            result = storage.delete_from_tbl('configuration', payload)
+            # Revert scheduled_process entry
+            payload = PayloadBuilder().WHERE(['name', '=', name]).payload()
+            result = storage.delete_from_tbl('scheduled_processes', payload)
+            raise web.HTTPInternalServerError(reason='Failed to created schedule. {}'.format(str(ins_ex)))
```
praveen-garg: repeated def for ` # Revert configuration entry` can we have it as a def please?

python/foglamp/services/core/api/service.py
```diff
@@ -143,21 +153,44 @@ def get_service_records():
...
 
         return web.json_response({'name': name, 'id': str(schedule.schedule_id)})
```
praveen-garg: I'm not sure, if last action fails do we have rollback mechanism to point ZERO?

amarendra-dianomic:
To be taken up afresh in a future sprint.

PR #894, opened 2018-07-02
FOGL-1442: addition of getChildCategories and addChildCategory

pintomax:
FOGL-1442: addition of getChildCategories and addChildCategory

Exceptions refactoring


C/services/core/configuration_manager.cpp
```diff
@@ -56,53 +57,79 @@ ConfigurationManager* ConfigurationManager::getInstance(const string& host,
...
-		throw StorageOperationException();
-	}
+		// Query via Storage client
+		allCategories = m_storage->queryTable("configuration", qAllCategories);
+		if (!allCategories)
+		{
+			// Data layer error
+			throw CategoryDetailsEx();
+		}
+		if (!allCategories->rowCount())
```
MarkRiddoch: You could combine you tests if you wanted to make the code more concise

C/services/core/configuration_manager.cpp
```diff
@@ -123,50 +151,82 @@ ConfigCategory ConfigurationManager::getCategoryAllItems(const string& categoryN
...
-	// Get first row
-	ResultSet::RowIterator it = categoryItems->firstRow();
-	ResultSet::Row* row = *it;
-	if (!row)
-	{
-		delete categoryItems;
-		throw CategoryDetailsException();
-	}	
+		// Cayegory not found
+		if (!categoryItems->rowCount())
```
MarkRiddoch: Again you could combine these two tests

C/services/core/configuration_manager.cpp
pintomax: yes, done

C/services/core/configuration_manager.cpp
pintomax: The code here returns two different exceptions and we cannot combine it

PR #911, opened 2018-07-05
FOGL-1488: first implementation of ConfigurationManager entry points

pintomax:
FOGL-1488: first implementation of ConfigurationManager entry points

ConfigurationManager is set with CoreManagementApi constructor

Next commits will add remaining POST/PUT/DELETE operations and handling of childCategories


C/services/core/core_management_api.cpp
```diff
@@ -63,12 +139,58 @@ void CoreManagementApi::defaultResource(shared_ptr<HttpServer::Response> respons
...
+	ServiceRegistry* registry = ServiceRegistry::getInstance();
+	if (!registry)
+	{
+		// TODO: add exception
+		throw;
+	}
+
+	// Wait for storage layer registration
+	ServiceRecord* storageDetails = 0;
+	while (storageLayerRetries)
```
MarkRiddoch: I do not think this is a good idea. The core management API needs to be up and running for the storage layer to register. Before it can register it must be started and passed the port of the core management API. For this to happen the constructor must have returned so that we have access to a fully constructed class. This functionality needs to be outside of the constructor.

pintomax: yes, code changed: the registerService now triggers ConfigurationManager instance and URL entry points.
Next commit will add the new functionality.

The storage service registration could set a callback for ConfigurationManager operations: that can be added with future commits.

C/services/core/include/configuration_manager.h
```diff
@@ -26,25 +26,25 @@ class ConfigurationManager {
 		ConfigCategory			createCategory(const std::string& categoryName,
 							       const std::string& categoryDescription,
 							       const std::string& categoryItems) const;
-		// Called by GET /foglamp/category/{categoryName}/{configItem}
+		// Called by GET /foglamp/service/category/{categoryName}/{configItem}
```
MarkRiddoch: Not quite true, it can be called either via the microservice management API or the admin API

pintomax: Fixed in 2f7b2f07a3a0308dbee52a1781360f5455b32e92

PR #913, opened 2018-07-06
FOGL-1477 - Add API for discovery of available plugins

amarendra-dianomic:
My local machine contains a few empty plugin folder. So **curl -X GET http://localhost:8081/foglamp/plugins/installed?type=south | jq** gives:
```
{
  "plugins": [
    {
      "name": "Simple Data Simulator Async plugin",
      "description": "Simple Simulator async plugin",
      "version": "1.0",
      "type": "south"
    },
    {
      "name": "http_south",
      "description": "HTTP South Plugin",
      "version": "1.0",
      "type": "south"
    }
  ]
}
```

and **curl -X GET http://localhost:8081/foglamp/plugins/installed | jq** gives:
```
{
  "plugins": [
    {
      "name": "OMF North",
      "description": "OMF North Plugin",
      "version": "1.0.0",
      "type": "north"
    },
    {
      "name": "OCS North",
      "description": "OCS North Plugin",
      "version": "1.0.0",
      "type": "north"
    },
    {
      "name": "Simple Data Simulator Async plugin",
      "description": "Simple Simulator async plugin",
      "version": "1.0",
      "type": "south"
    },
    {
      "name": "http_south",
      "description": "HTTP South Plugin",
      "version": "1.0",
      "type": "south"
    }
  ]
}

```

However, I find a few tests failing as below:
======================================= 15 failed, 1131 passed, 33 skipped, 103 warnings in 83.21 seconds =======================================
`
All tests are failing with identical reason, after upgrading to aiohttp 3.2.1, as below:
```
/home/asinha/Development/FogLAMP/tests/unit/python/foglamp/common/web/test_middleware.py:49: RuntimeWarning: coroutine 'BaseTestServer.start_server' was never awaited
    server.start_server(loop=loop)
```

### DIFF
--- a/python/foglamp/common/plugin_discovery.py
+++ b/python/foglamp/common/plugin_discovery.py
@@ -8,6 +8,7 @@
 
 import os
 from foglamp.common import logger
+from foglamp.common.common import _FOGLAMP_ROOT
 
 __author__ = "Amarendra K Sinha"
 __copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
@@ -16,9 +17,6 @@ __version__ = "${VERSION}"
 
 
 _logger = logger.setup(__name__)
-
-_FOGLAMP_DATA = os.getenv("FOGLAMP_DATA", default=None)
-_FOGLAMP_ROOT = os.getenv("FOGLAMP_ROOT", default='/usr/local/foglamp')
 
 
 class PluginDiscovery(object):
--- a/python/foglamp/common/plugin_discovery.py
+++ b/python/foglamp/common/plugin_discovery.py
@@ -1,0 +1,73 @@
+# -*- coding: utf-8 -*-
+
+# FOGLAMP_BEGIN
+# See: http://foglamp.readthedocs.io/
+# FOGLAMP_END
+
+"""Common Plugin Discovery Class"""
+
+import os
+from foglamp.common import logger
+
+__author__ = "Amarendra K Sinha"
+__copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
+__license__ = "Apache 2.0"
+__version__ = "${VERSION}"
+
+
+_logger = logger.setup(__name__)
+
+_FOGLAMP_DATA = os.getenv("FOGLAMP_DATA", default=None)
+_FOGLAMP_ROOT = os.getenv("FOGLAMP_ROOT", default='/usr/local/foglamp')
+
+class PluginDiscoveryInstalled(object):
+    @classmethod
+    def get_plugins(cls, plugin_type):
+        directories = cls.get_plugin_folders(plugin_type)
+        configs = []
+        for d in directories:
+            configs.append(cls.get_plugin_config(d, plugin_type))
+        return configs
+
+    @classmethod
+    def get_plugin_folders(cls, plugin_type):
+        directories = []
+        try:
+            directories = [d for d in os.listdir(_FOGLAMP_ROOT + "/python/foglamp/plugins/"+plugin_type)
+                           if os.path.isdir(_FOGLAMP_ROOT + "/python/foglamp/plugins/" + plugin_type + "/" + d) and
+                           not d.startswith("__") and d != "empty" and d != "common"]
+        except FileNotFoundError:
+            pass
+        return directories
+
+    @classmethod
+    def get_plugin_config(cls, plugin_name, plugin_type):
+        plugin_module_path = "foglamp.plugins.south" if plugin_type == 'south' else "foglamp.plugins.north"
+
+        # Now load the plugin to fetch its configuration
+        try:
+            # "plugin_module_path" is fixed by design. It is MANDATORY to keep the plugin in the exactly similar named
+            # folder, within the plugin_module_path.
+            import_file_name = "{path}.{dir}.{file}".format(path=plugin_module_path, dir=plugin_name, file=plugin_name)
+            _plugin = __import__(import_file_name, fromlist=[''])
+
+            # Fetch configuration from the configuration defined in the plugin
+            plugin_info = _plugin.plugin_info()
+            plugin_config =  {
+                'name': plugin_info['name'],
+                'type': plugin_info['type'],
+                'description': plugin_info['config']['plugin']['description'],
+                'version': plugin_info['version']
+            }
+        except ImportError as ex:
+            plugin_config = {
+                'name': plugin_name,
+                'description': 'Plugin "{}" import problem from path "{}". {}'.format(plugin_name, plugin_module_path, str(ex))
+            }
+        except Exception as ex:
+            plugin_config = {
+                'name': plugin_name,
+                'description': 'Plugin "{}" raised exception "{}" while fetching config'.format(plugin_name, str(ex))
+            }
+
+        return plugin_config
--- a/python/foglamp/common/plugin_discovery.py
+++ b/python/foglamp/common/plugin_discovery.py
@@ -42,7 +42,9 @@ class PluginDiscovery(object):
         directories = cls.get_plugin_folders(plugin_type)
         configs = []
         for d in directories:
-            configs.append(cls.get_plugin_config(d, plugin_type))
+            plugin_config = cls.get_plugin_config(d, plugin_type)
+            if plugin_config is not None:
+                configs.append(plugin_config)
         return configs
 
     @classmethod
@@ -60,11 +62,10 @@ class PluginDiscovery(object):
     @classmethod
     def get_plugin_config(cls, plugin_name, plugin_type):
         plugin_module_path = "foglamp.plugins.south" if plugin_type == 'south' else "foglamp.plugins.north"
+        plugin_config = None
 
         # Now load the plugin to fetch its configuration
         try:
-            # "plugin_module_path" is fixed by design. It is MANDATORY to keep the plugin in the exactly similar named
-            # folder, within the plugin_module_path.
             import_file_name = "{path}.{dir}.{file}".format(path=plugin_module_path, dir=plugin_name, file=plugin_name)
             _plugin = __import__(import_file_name, fromlist=[''])
 
@@ -77,14 +78,8 @@ class PluginDiscovery(object):
                 'version': plugin_info['version']
             }
         except ImportError as ex:
-            plugin_config = {
-                'name': plugin_name,
-                'description': 'Plugin "{}" import problem from path "{}". {}'.format(plugin_name, plugin_module_path, str(ex))
-            }
+            _logger.error('Plugin "{}" import problem from path "{}". {}'.format(plugin_name, plugin_module_path, str(ex)))
         except Exception as ex:
-            plugin_config = {
-                'name': plugin_name,
-                'description': 'Plugin "{}" raised exception "{}" while fetching config'.format(plugin_name, str(ex))
-            }
+            _logger.exception('Plugin "{}" raised exception "{}" while fetching config'.format(plugin_name, str(ex)))
 
         return plugin_config
--- a/python/foglamp/common/plugin_discovery.py
+++ b/python/foglamp/common/plugin_discovery.py
@@ -38,7 +38,8 @@ class PluginDiscoveryInstalled(object):
                            not d.startswith("__") and d != "empty" and d != "common"]
         except FileNotFoundError:
             pass
-        return directories
+        else:
+            return directories
 
     @classmethod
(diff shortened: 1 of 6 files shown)


python/foglamp/common/plugin_discovery.py
```diff
@@ -0,0 +1,73 @@
...
+__author__ = "Amarendra K Sinha"
+__copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
+__license__ = "Apache 2.0"
+__version__ = "${VERSION}"
+
+
+_logger = logger.setup(__name__)
+
+_FOGLAMP_DATA = os.getenv("FOGLAMP_DATA", default=None)
+_FOGLAMP_ROOT = os.getenv("FOGLAMP_ROOT", default='/usr/local/foglamp')
```
praveen-garg: can we find a common place for it? We also had discussed it when schedule module was done

python/foglamp/common/plugin_discovery.py
```diff
@@ -0,0 +1,73 @@
...
+    @classmethod
+    def get_plugin_folders(cls, plugin_type):
+        directories = []
+        try:
+            directories = [d for d in os.listdir(_FOGLAMP_ROOT + "/python/foglamp/plugins/"+plugin_type)
+                           if os.path.isdir(_FOGLAMP_ROOT + "/python/foglamp/plugins/" + plugin_type + "/" + d) and
+                           not d.startswith("__") and d != "empty" and d != "common"]
+        except FileNotFoundError:
+            pass
+        return directories
```
praveen-garg: except else is better pattern

python/foglamp/common/plugin_discovery.py
```diff
@@ -0,0 +1,73 @@
...
+            # folder, within the plugin_module_path.
+            import_file_name = "{path}.{dir}.{file}".format(path=plugin_module_path, dir=plugin_name, file=plugin_name)
+            _plugin = __import__(import_file_name, fromlist=[''])
+
+            # Fetch configuration from the configuration defined in the plugin
+            plugin_info = _plugin.plugin_info()
+            plugin_config =  {
+                'name': plugin_info['name'],
+                'type': plugin_info['type'],
+                'description': plugin_info['config']['plugin']['description'],
```
praveen-garg: is the only diff b/w plugin_config and plugin_info is this value?

python/foglamp/common/plugin_discovery.py
```diff
@@ -0,0 +1,73 @@
...
+                'version': plugin_info['version']
+            }
+        except ImportError as ex:
+            plugin_config = {
+                'name': plugin_name,
+                'description': 'Plugin "{}" import problem from path "{}". {}'.format(plugin_name, plugin_module_path, str(ex))
+            }
```
praveen-garg: it should be raised

python/foglamp/common/plugin_discovery.py
```diff
@@ -0,0 +1,73 @@
...
+        except Exception as ex:
+            plugin_config = {
+                'name': plugin_name,
+                'description': 'Plugin "{}" raised exception "{}" while fetching config'.format(plugin_name, str(ex))
```
praveen-garg: its should be raised. how it can be plugin_config dict

python/foglamp/services/core/api/plugin_discovery.py
```diff
@@ -0,0 +1,45 @@
...
+
+__author__ = "Amarendra K Sinha"
+__copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
+__license__ = "Apache 2.0"
+__version__ = "${VERSION}"
+
+
+_help = """
+    -------------------------------------------------------------------------------
+    | GET             | /foglamp/plugins/installed                                |
```
praveen-garg: don't we want not installed but available too?

python/foglamp/services/core/api/plugin_discovery.py
```diff
@@ -0,0 +1,45 @@
...
+    :Example:
+        curl -X GET http://localhost:8081 /foglamp/plugins/installed
+        curl -X GET http://localhost:8081 /foglamp/plugins/installed?type=north
+    """
+
+    plugin_type = None
+    if 'type' in request.query and request.query['type'] != '':
+        plugin_type = request.query['type']
+
+    if plugin_type is not None and plugin_type not in ['north', 'south']:
```
praveen-garg: `is not None` tested? 👎

python/foglamp/services/core/api/plugin_discovery.py
```diff
@@ -0,0 +1,45 @@
...
+        raise web.HTTPNotFound(reason="Invalid plugin type. Must be 'north' or 'south'.")
+
+    if plugin_type is None:
+        plugins_list = PluginDiscoveryInstalled.get_plugins("north")
+        plugins_list_south = PluginDiscoveryInstalled.get_plugins("south")
+        plugins_list.extend(plugins_list_south)
```
praveen-garg: well! PluginDiscoveryInstalled.get_plugins should have default value as None and return south+ north by default.  if we plan to make 2 calls; we  should name   plugins_list to plugins_list_north and then merge south+north into plugins_list

python/foglamp/services/core/routes.py
```diff
@@ -119,6 +120,9 @@ def setup(app):
     # Get Syslog
     app.router.add_route('GET', '/foglamp/syslog', support.get_syslog_entries)
 
+    # Get Plugin
+    app.router.add_route('GET', '/foglamp/plugins/installed', plugin_discovery.get_plugins_installed)
```
praveen-garg: Tests please. and what about available for install list

python/foglamp/common/plugin_discovery.py
amarendra-dianomic: plugin_info contains other data also and we have picked only what has been mentioned in Architecture doc.

python/foglamp/common/plugin_discovery.py
amarendra-dianomic: This may happen in case of just one error also. Hence why should we discard all the valid entries by raising an exception just because of a single error?

python/foglamp/common/plugin_discovery.py
amarendra-dianomic: This may happen in case of just one error also. Hence why should we discard all the valid entries by raising an exception just because of a single error?

python/foglamp/services/core/api/plugin_discovery.py
amarendra-dianomic: As @MarkRiddoch mentioned, it will taken up in a future sprint.

python/foglamp/services/core/api/plugin_discovery.py
amarendra-dianomic: Please elaborate.

python/foglamp/services/core/routes.py
amarendra-dianomic: As @MarkRiddoch mentioned, it will taken up in a future sprint.

python/foglamp/services/core/api/plugin_discovery.py
amarendra-dianomic: done

python/foglamp/common/plugin_discovery.py
amarendra-dianomic: done

python/foglamp/common/plugin_discovery.py
amarendra-dianomic: Please suggest.

python/foglamp/common/plugin_discovery.py
```diff
@@ -0,0 +1,74 @@
...
+        for d in directories:
+            configs.append(cls.get_plugin_config(d, plugin_type))
+        return configs
+
+    @classmethod
+    def get_plugin_folders(cls, plugin_type):
+        directories = []
+        try:
+            directories = [d for d in os.listdir(_FOGLAMP_ROOT + "/python/foglamp/plugins/"+plugin_type)
+                           if os.path.isdir(_FOGLAMP_ROOT + "/python/foglamp/plugins/" + plugin_type + "/" + d) and
```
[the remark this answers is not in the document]
amarendra-dianomic: done

python/foglamp/common/plugin_discovery.py
```diff
@@ -0,0 +1,74 @@
...
+__license__ = "Apache 2.0"
+__version__ = "${VERSION}"
+
+
+_logger = logger.setup(__name__)
+
+_FOGLAMP_DATA = os.getenv("FOGLAMP_DATA", default=None)
+_FOGLAMP_ROOT = os.getenv("FOGLAMP_ROOT", default='/usr/local/foglamp')
+
+class PluginDiscoveryInstalled(object):
```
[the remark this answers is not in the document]
amarendra-dianomic: done

python/foglamp/common/plugin_discovery.py
praveen-garg: Can we have a common file  to import these 2? these are being used in server.py, scheduler.py and now here! and may be more...

python/foglamp/services/core/api/plugin_discovery.py
```diff
@@ -0,0 +1,40 @@
...
+        curl -X GET http://localhost:8081 /foglamp/plugins/installed
+        curl -X GET http://localhost:8081 /foglamp/plugins/installed?type=north
+    """
+
+    plugin_type = None
+    if 'type' in request.query and request.query['type'] != '':
+        plugin_type = request.query['type']
+
+    if plugin_type is not None and plugin_type not in ['north', 'south']:
+        raise web.HTTPNotFound(reason="Invalid plugin type. Must be 'north' or 'south'.")
```
praveen-garg: HTTPBadRequest

python/foglamp/common/plugin_discovery.py
praveen-garg: could you please elaborate with use case? as It looks like an error is wrapped into expected response format!

praveen-garg:
@amarendra-dianomic  Please fix / add the tests.

python/foglamp/common/plugin_discovery.py
MarkRiddoch: If we can't import a plugin then we should remove that plugin from the list of plugins we return, since it is not correctly installed in this case. We should still return the other plugins however.
The use of this entry point is to allow a user interface to offer a pick list of plugins when adding a service or a task, therefore we should limit it to those that will succeed if they are chosen.

python/foglamp/common/plugin_discovery.py
MarkRiddoch: As above remove the plugin in question from the returned list. By all means log the issue.

python/foglamp/common/plugin_discovery.py
amarendra-dianomic: Let us do it via another ticket as it will require through testing of affected files.

python/foglamp/services/core/api/plugin_discovery.py
amarendra-dianomic: done

amarendra-dianomic:

python/foglamp/common/plugin_discovery.py
praveen-garg: Pleae make a new common file, and import that here. we can fix others later.

python/foglamp/services/core/api/plugin_discovery.py
```diff
@@ -0,0 +1,40 @@
...
+    -------------------------------------------------------------------------------
+"""
+
+
+async def get_plugins_installed(request):
+    """ get list of installed plugins
+
+    :Example:
+        curl -X GET http://localhost:8081 /foglamp/plugins/installed
+        curl -X GET http://localhost:8081 /foglamp/plugins/installed?type=north
```
praveen-garg: space after host:port

python/foglamp/services/core/api/plugin_discovery.py
```diff
@@ -0,0 +1,40 @@
...
+    """
+
+    plugin_type = None
+    if 'type' in request.query and request.query['type'] != '':
+        plugin_type = request.query['type']
+
+    if plugin_type is not None and plugin_type not in ['north', 'south']:
```
praveen-garg: we should compare after converting to uppercase or lowercase

amarendra-dianomic:

PR #914, opened 2018-07-06
FOGL-1500 - Remove pending 'device | translator'  text references for plugins

amarendra-dianomic:


python/foglamp/services/south/exceptions.py
```diff
@@ -33,10 +33,10 @@ class InvalidAddressError(Exception):
...
 
 class InvalidPluginTypeError(Exception):
-    """ Invalid plugin type, only the type -device- is allowed """
+    """ Invalid plugin type, only the type -south- is allowed """
     pass
 
 
 class DataRetrievalError(Exception):
-    """ Unable to retrieve data from the device """
+    """ Unable to retrieve data from the South service """
```
praveen-garg: service should be plugin?

amarendra-dianomic: done

PR #916, opened 2018-07-06
1.3 as master

ashwinscale:


praveen-garg:
* Known issues
- FOGL-1533 (Workaround: Use (case insensitive) service name same as of the plugin name to load)
- FOGL-1527  purge process with simultaneous ingest vs DB concurrency
- systemctl stop /start does not work as expected; when auth is mandatory

praveen-garg:
@ashwinscale  I think we should just merge 1.3 branch into master; instead of develop. if we want to move with this PR itself, we shall need to revert latest commit 091fe56 on develop and wait for this PR merge for further commits.

PR #937, opened 2018-07-18
FOGL-1548: DHT11 plugin for C++ south service

AmandeepArora:
Added DHT11 C/C++ plugin support. Usage instructions are in plugins/south/dht11/README.rst.


C/plugins/south/dht11/dht11.cpp
```diff
@@ -0,0 +1,131 @@
+/*
+ * FogLAMP south service plugin
+ *
+ * Copyright (c) 2018 OSIsoft, LLC
+ *
+ * Released under the Apache 2.0 Licence
+ *
+ * Author: Mark Riddoch
```
MarkRiddoch: Please add your name as author

C/plugins/south/dht11/dht11.cpp
```diff
@@ -0,0 +1,131 @@
...
+#include <iostream>
+#include <reading.h>
+#include <wiringPi.h>
+#include <dht11.h>
+#include <logger.h>
+
+using namespace std;
+
+/**
+ * Constructor for the random "sensor"
```
MarkRiddoch: Please update comment and also add the @param following the doxygen standard.

C/plugins/south/dht11/dht11.cpp
```diff
@@ -0,0 +1,131 @@
...
+ */
+DHT11::DHT11(unsigned int pin) :
+m_pin(pin) // , m_initialDPValue(0), m_lastReading("initial", new Datapoint("dummy", m_initialDPValue))
```
MarkRiddoch: Please remove commented out code

C/plugins/south/dht11/dht11.cpp
```diff
@@ -0,0 +1,131 @@
...
+{
+	if ( wiringPiSetup() == -1 ) {
+		Logger::getLogger()->fatal("wiringPiSetup failed, can't proceed");
+		throw runtime_error("wiringPiSetup failed, can't proceed");
+	}
+	(void) takeReading(true);
+}
+
+/**
+ * Destructor for the random "sensor"
```
MarkRiddoch: Please update comment

C/plugins/south/dht11/dht11.cpp
```diff
@@ -0,0 +1,131 @@
...
+ */
+DHT11::~DHT11()
+{
+}
+
+/**
+ * Take a reading from the random "sensor"
```
MarkRiddoch: Please update comment

C/plugins/south/dht11/dht11.cpp
```diff
@@ -0,0 +1,131 @@
...
+			sensorData[k]=dht11_dat[k];
+		return true;
+        }else  {
+                //Logger::getLogger()->info( "Data not good, skip\n" );
+		return false;
+        }
+}
+
+
+Reading DHT11::takeReading(bool firstReading)
```
MarkRiddoch: Doxygen comment

C/plugins/south/dht11/dht11.cpp
```diff
@@ -0,0 +1,131 @@
...
+{
+	static uint8_t sensorData[4] = {0,0,0,0};
+	
+	bool valid = false;
+	do {
+		valid = readSensorData(sensorData);
+	} while(!valid && firstReading);
```
MarkRiddoch: I would suggest commenting on this firstReading behaviour s important here.

C/plugins/south/dht11/plugin.cpp
```diff
@@ -0,0 +1,110 @@
+/*
+ * FogLAMP south plugin.
+ *
+ * Copyright (c) 2018 OSisoft, LLC
+ *
+ * Released under the Apache 2.0 Licence
+ *
+ * Author: Mark Riddoch
```
MarkRiddoch: Change this to your name

C/plugins/south/dht11/dht11.cpp
```diff
@@ -0,0 +1,131 @@
...
+        tmp << ((unsigned int)sensorData[0]) << "." << ((unsigned int)sensorData[1]);
+	DatapointValue dpv1(stod(tmp.str()));
+        vec.push_back(new Datapoint("Humidity", dpv1));
+
+	ostringstream tmp2;
+        tmp2 << ((unsigned int)sensorData[2]) << "." <<  ((unsigned int)sensorData[3]);
+	DatapointValue dpv2(stod(tmp2.str()));
+        vec.push_back(new Datapoint ("Temperature", dpv2));
+
+	return Reading("dht11", vec);
```
MarkRiddoch: This should be using the configuration item asset and not the hardwired string.

C/plugins/south/dht11/plugin.cpp
```diff
@@ -0,0 +1,110 @@
...
+{
+	return &info;
+}
+
+/**
+ * Initialise the plugin, called to get the plugin handle
+ */
+PLUGIN_HANDLE plugin_init(ConfigCategory *config)
+{
+	unsigned int pin=7;
```
MarkRiddoch: This should come from the default of the configuration item and not be hardwired.

In fact the way the configuration works the default will become the value if no value has been set, so the code is redundant.

C/plugins/south/dht11/plugin.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+	if (config->itemExists("pin"))
+        {
+                pin = stoul(config->getValue("pin"), nullptr, 0);
+        }
+	
+	//Logger::getLogger()->info("DHT11 C++ plugin_init: config='%s' ", config->toJSON().c_str());
```
MarkRiddoch: Please take out commented out code or make #if DEBUG

C/plugins/south/dht11/plugin.cpp
```diff
@@ -0,0 +1,110 @@
...
+{
+	DHT11 *dht11 = (DHT11*)handle;
+	return dht11->takeReading();
+}
+
+/**
+ * Reconfigure the plugin
+ */
+void plugin_reconfigure(PLUGIN_HANDLE *handle, string& newConfig)
+{
```
MarkRiddoch: This entry point is to allow for the dynamic reconfiguration of the plugin. The only things that matter are probably the asset, which might or might not need attention once that is implemented. Also the pion could be changed and currently that will be ignored.

scripts/services/c-south
```diff
@@ -0,0 +1,16 @@
...
+if [ "${FOGLAMP_ROOT}" = "" ]; then
+	FOGLAMP_ROOT=/usr/local/foglamp
+fi
+
+if [ ! -d "${FOGLAMP_ROOT}" ]; then
+	logger "FogLAMP home directory missing or incorrectly set environment"
+	exit 1
+fi
+
+# We run the Python code from the python directory
```
MarkRiddoch: Please remove references to running python, it will only confuse people in the future.

C/plugins/south/dht11/plugin.cpp
AmandeepArora: To start using a new GPIO pin to get data from the sensor, it is probably better to shutdown Pi or atleast the sensor and then plug it back in using the new GPIO pin. It would be better to power reset the sensor before changing the GPIO pin it is connected to. So dynamic (runtime) configuration change is probably not required. So even if I don't add logic to the reconfigure function, I think pin configuration change would still be stored into DB and would be used the next time dht11 plugin is started.

Should I add code to handle "asset name" change?

C/plugins/south/dht11/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+cmake_minimum_required(VERSION 2.4.0)
```
praveen-garg: we have this 2.4.0 at some places and 2.8.8 at some; not sure if this should be a const at common place? so that we can use that in every (any) CMakeLists.txt.

C/plugins/south/dht11/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+
+project(dht11)
+
+set(CMAKE_CXX_FLAGS_DEBUG "-O0 -ggdb")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+set(WIRINGPILIBS -lwiringPi -lwiringPiDev)
```
praveen-garg: WIRING_PI_LIBS is more readable.

C/plugins/south/dht11/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+# Include header files
+include_directories(include ../../../services/common/include ../common/include ../../../common/include ../../../thirdparty/rapidjson/include /usr/local/include/)
+
+file(GLOB SOURCES "*.cpp" ../common/*.cpp)
+
+# Create shared library
+add_library(${PROJECT_NAME} SHARED ${SOURCES})
+target_link_libraries(${PROJECT_NAME} ${WIRINGPILIBS})
+
+set_target_properties(${PROJECT_NAME} PROPERTIES SOVERSION 1)
```
praveen-garg: better to use major.minor version convention for SOVERSION

C/plugins/south/dht11/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+
+set(CMAKE_CXX_FLAGS_DEBUG "-O0 -ggdb")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+set(WIRINGPILIBS -lwiringPi -lwiringPiDev)
+
+# Find source files
+file(GLOB SOURCES *.cpp)
+
+# Include header files
+include_directories(include ../../../services/common/include ../common/include ../../../common/include ../../../thirdparty/rapidjson/include /usr/local/include/)
```
praveen-garg: do we need all of these? e.g. rapidjson

C/plugins/south/dht11/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+
+file(GLOB SOURCES "*.cpp" ../common/*.cpp)
```
praveen-garg: could you please explain with comments w.r.t. 
```
+# Find source files
+file(GLOB SOURCES *.cpp)
```

C/plugins/south/dht11/README.rst
```diff
@@ -0,0 +1,20 @@
...
+DHT11 C/C++ South plugin for Raspberry Pi
+=========================================
+
+The DHT11 C/C++ module uses libwiringPi. To install that:
+
+git clone git://git.drogon.net/wiringPi
+
+cd wiringPi
+
+./build
```
praveen-garg: put these 3 lines in proper ReStructuredText format

```
.. code-block:: console
```

C/plugins/south/dht11/README.rst
```diff
@@ -0,0 +1,20 @@
...
+
+./build 
+
+The wiringPi library and related symbolic link would get installed in /usr/local/lib.
+
+DHT11 schedule has been added in disabled mode in sqlite init.sql.
+
+To enable the dht11 plugin schedule:
+
+curl -sX PUT http://10.2.5.15:8081/foglamp/schedule/6b25f4d9-c7f3-4fc8-bd4a-4cf79f7055ca/enable
```
praveen-garg: in ```.. code-block:: console```

C/plugins/south/dht11/dht11.cpp
```diff
@@ -0,0 +1,146 @@
+/*
+ * FogLAMP south service plugin
+ *
+ * Copyright (c) 2018 OSIsoft, LLC
+ *
+ * Released under the Apache 2.0 Licence
+ *
+ * Author: Amandeep Singh Arora
+ */
```
praveen-garg: Not very, but looks doxygen has this format for this. 

```
/** @file dht11.cpp
 *  @brief FogLAMP south service DHT11 plugin
 *
 *   ....what this file is for, if more details required then brief  ....
 *  @author Amandeep Singh Arora
 *  @version ..if any..
 *  @license  Released under the Apache 2.0 Licence or just Apache 2.0
 */
```

C/plugins/south/dht11/dht11.cpp
```diff
@@ -0,0 +1,146 @@
...
+#include <sstream>
+#include <iostream>
+#include <reading.h>
+#include <wiringPi.h>
+#include <dht11.h>
+#include <logger.h>
+
+#define MAX_SENSOR_READ_RETRIES		(100)
```
praveen-garg: any reason for too many spaces between `MAX_SENSOR_READ_RETRIES` and `(100)`

C/plugins/south/dht11/dht11.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+using namespace std;
+
+/**
+ * Constructor for the dht11 "sensor"
+ *
+ * @param pin	The RPi pin to which DHT11 sensor's data pin is attached (e.g. pin 7 = GPIO 4, this is the default)
+ */
+DHT11::DHT11(unsigned int pin) :
+m_pin(pin)
```
praveen-garg: why `m_pin(pin)` in new line?  I think we are not hitting max no of characters limit in 1 line

C/plugins/south/dht11/dht11.cpp
```diff
@@ -0,0 +1,146 @@
...
+ *
+ * @param sensorData   This buffer is filled with sensor data if read succeeds & data checksum matches
+ */
+bool DHT11::readSensorData(uint8_t sensorData[])
+{
+        uint8_t laststate       = HIGH;
+        uint8_t counter         = 0;
+        uint8_t j               = 0, i;
+        float   f; /* fahrenheit */
+	uint8_t dht11_dat[5] = {0, 0, 0, 0, 0};
```
praveen-garg: indent

C/plugins/south/dht11/dht11.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+        /* pull pin down for 18 milliseconds */
+        pinMode( m_pin, OUTPUT );
+        digitalWrite( m_pin, LOW );
+        delay( 18 );
```
praveen-garg: this is standard? or should be configurable?

C/plugins/south/dht11/dht11.cpp
```diff
@@ -0,0 +1,146 @@
...
+        /* then pull it up for 40 microseconds */
+        digitalWrite( m_pin, HIGH );
+        delayMicroseconds( 40 );
```
praveen-garg: this is standard? or should be configurable?

C/plugins/south/dht11/dht11.cpp
```diff
@@ -0,0 +1,146 @@
...
+        /* prepare to read the pin */
+        pinMode( m_pin, INPUT );
+
+        /* detect change and read data */
+        for ( i = 0; i < MAXTIMINGS; i++ )
```
praveen-garg: MAX_TIMINGS

C/plugins/south/dht11/dht11.cpp
```diff
@@ -0,0 +1,146 @@
...
+	bool valid = false;
+        unsigned int count=0;
+	do {
+		valid = readSensorData(sensorData);
+		count++;
+	} while(!valid && firstReading && count < MAX_SENSOR_READ_RETRIES);
+
+        vector<Datapoint *> vec;
+
+	stringstream tmp;
```
praveen-garg: better to name it `stringstream humidity` or `ostringstream humidity`

C/plugins/south/dht11/dht11.cpp
```diff
@@ -0,0 +1,146 @@
...
+        tmp << ((unsigned int)sensorData[0]) << "." << ((unsigned int)sensorData[1]);
+	DatapointValue dpv1(stod(tmp.str()));
+        vec.push_back(new Datapoint("Humidity", dpv1));
+
+	ostringstream tmp2;
```
praveen-garg: ostringstream temp

C/plugins/south/dht11/include/dht11.h
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * Copyright (c) 2018 OSIsoft, LLC
+ *
+ * Released under the Apache 2.0 Licence
+ *
+ * Author: Amandeep Singh Arora
+ */
+#include <reading.h>
+
+#define MAXTIMINGS      85   /* was 85 */
```
praveen-garg: MAX_TIMINGS

in comment `was` ?

C/plugins/south/dht11/include/dht11.h
```diff
@@ -0,0 +1,28 @@
...
+
+class DHT11 {
+	public:
+		DHT11(unsigned int pin=7);
```
praveen-garg: let's remove default arg value

C/plugins/south/dht11/plugin.cpp
```diff
@@ -0,0 +1,116 @@
...
+ * The DHT11 plugin interface
+ */
+extern "C" {
+
+/**
+ * The plugin information structure
+ */
+static PLUGIN_INFORMATION info = {
+	"DHT11",                  // Name
+	"1.0.0",                  // Version
```
praveen-garg: this should match the SOVERSION? it may get importance once we have debian packaging versions

C/plugins/south/dht11/plugin.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+	Logger::getLogger()->info("m_assetName set to %s", dht11->getAssetName());
+
+	return (PLUGIN_HANDLE)dht11;
+}
+
+/**
+ * Start the Async handling for the plugin
+ */
+void plugin_start(PLUGIN_HANDLE *handle)
```
praveen-garg: we should remove this!

C/plugins/south/dht11/plugin.cpp
```diff
@@ -0,0 +1,116 @@
...
+{
+}
+
+/**
+ * Poll for a plugin reading
+ */
+Reading plugin_poll(PLUGIN_HANDLE *handle)
+{
+	DHT11 *dht11 = (DHT11*)handle;
```
praveen-garg: handle is config dict, not sure why its being typecast to DHT11.

C/plugins/south/dht11/plugin.cpp
```diff
@@ -0,0 +1,116 @@
...
+void plugin_reconfigure(PLUGIN_HANDLE *handle, string& newConfig)
+{
+}
+
+/**
+ * Shutdown the plugin
+ */
+void plugin_shutdown(PLUGIN_HANDLE *handle)
+{
+	DHT11 *dht11 = (DHT11*)handle;
```
praveen-garg: handle is config dict (jdoc)

scripts/plugins/storage/sqlite/init.sql
```diff
@@ -617,6 +617,13 @@ INSERT INTO foglamp.configuration ( key, description, value )
               ' { "plugin" : { "type" : "string", "value" : "omf", "default" : "omf", "description" : "Module that OMF North Plugin will load" } } '
             );
 
+-- dht11 - South plugin for DHT11 - C
+INSERT INTO foglamp.configuration ( key, description, value )
+     VALUES ( 'dht11',
+              'DHT11 South Plugin',
```
praveen-garg: `DHT11 South C Plugin`

C/plugins/south/dht11/plugin.cpp
MarkRiddoch: No, it is not defined to always be this. The Python plugins have tended to do this for connivence, it was never intended that this was the case. The handle is just that, an opaque structure that is used by the plugin for purposes of identification, it has no meaning outside of the plugin.

praveen-garg: this makes sense to me. It would be great then to align python code the same way as well. 
We may set there to use handle(obj)->setConfig getConfig instead of just making handle the config itself; So that reconfigure callback hook can compare with newConfig.

C/plugins/south/dht11/CMakeLists.txt
AmandeepArora: Yes. RapidJson library would be required for parsing reconf request.

C/plugins/south/dht11/CMakeLists.txt
AmandeepArora: set SOURCES variable to list of files in current directory. This is being done twice in CMakeLists.txt, so  removing one of them.

C/plugins/south/dht11/plugin.cpp
AmandeepArora: Yes, will deal with it when we do so.

C/plugins/south/dht11/plugin.cpp
AmandeepArora: I think we can leave the empty signature here, just for completeness sake.

C/plugins/south/dht11/CMakeLists.txt
AmandeepArora: This is the oldest cmake version having all the features that the cmake script needs. Since the various cmake scripts in our code base use different cmake features, so I guess it is Ok to have different values in various cmake files.

C/plugins/south/dht11/CMakeLists.txt
AmandeepArora: SOVERSION 1 is being used in most other CMakeLists.txt files also in C/C++ part of the code.
@MarkRiddoch : Please provide your input here.

C/plugins/south/dht11/dht11.cpp
AmandeepArora: Doxygen format has not been used in anywhere in code base. I searched for "@author" recursively.

C/plugins/south/dht11/dht11.cpp
AmandeepArora: This is standard.

C/plugins/south/dht11/dht11.cpp
AmandeepArora: Standard.

C/plugins/south/dht11/include/dht11.h
AmandeepArora: Tried various values, during UT. Comment is not required anymore, removed it.

C/plugins/south/dht11/dht11.cpp
```diff
@@ -0,0 +1,145 @@
...
+using namespace std;
+
+/**
+ * Constructor for the dht11 "sensor"
+ *
+ * @param pin	The RPi pin to which DHT11 sensor's data pin is attached (e.g. pin 7 = GPIO 4, this is the default)
+ */
+DHT11::DHT11(unsigned int pin) : m_pin(pin)
+{
+	if ( wiringPiSetup() == -1 ) {
```
MarkRiddoch: This can only be called as root according to the documentation, is this true? We do not normally run as root.

Can we get any error details for the failure? Is errno set for example?

C/plugins/south/dht11/dht11.cpp
```diff
@@ -0,0 +1,145 @@
...
+         */
+        if ( (j >= 40) &&
+             (dht11_dat[4] == ( (dht11_dat[0] + dht11_dat[1] + dht11_dat[2] + dht11_dat[3]) & 0xFF) ) )
+        {
+		int k=0;
+		for (k=0; k<4; k++)
+			sensorData[k]=dht11_dat[k];
+		return true;
+        }else  {
+                //Logger::getLogger()->info( "Data not good, skip\n" );
```
MarkRiddoch: We should have some indication of this failure, if only on the first failed attempt after a good attempt or startup. Otherwise we get nothing and have no idea what is happening.

I note below this is called in a retry loop, so the log should be moved there. It certainly should not be commented out and left here.

C/plugins/south/dht11/dht11.cpp
```diff
@@ -0,0 +1,145 @@
...
+Reading DHT11::takeReading(bool firstReading)
+{
+	static uint8_t sensorData[4] = {0,0,0,0};
+	
+	bool valid = false;
+        unsigned int count=0;
+	do {
+		valid = readSensorData(sensorData);
+		count++;
+	} while(!valid && firstReading && count < MAX_SENSOR_READ_RETRIES);
```
MarkRiddoch: It is good to see we retry on failure, but if we exceed the retry count then we return random data.

C/plugins/south/dht11/plugin.cpp
MarkRiddoch: Sure, hardware changes will probably result in a restart, but they may not, for example if two are plugged in and you want to swap from one to the other for some reason. Also changes to the asset code could easily happen at runtime.

scripts/plugins/storage/sqlite/init.sql
```diff
@@ -680,6 +687,10 @@ INSERT INTO foglamp.scheduled_processes ( name, script ) VALUES ( 'North Statist
 --
 INSERT INTO foglamp.scheduled_processes ( name, script ) VALUES ( 'North_Readings_to_PI',   '["tasks/north_c"]' );
 
+-- South Tasks - C code
+--
+INSERT INTO foglamp.scheduled_processes ( name, script ) VALUES ( 'dht11',   '["services/c-south"]' );
```
pintomax: Better to use:

"services/south_c"
same as "tasks/north_c"

Try keeping the same naming convention

C/plugins/south/dht11/dht11.cpp
AmandeepArora: This function tries to access /dev/mem, else /dev/gpiomem and that's normally a privileged operation. But if we allow current user to access "/dev/gpiomem" beforehand, then root access is not specifically required from plugin code.

Errno is not set and this API only returns -1 or EXIT_FAILURE. So it doesn't give more specific information.

C/plugins/south/dht11/dht11.cpp
AmandeepArora: There are around 20-30% failures while reading the sensor and that is a common observation/behavior based on what I read in online discussion forums. So what is attempted here is that we just return the previous reading if current reading attempt fails. And in case of first reading, 100 tries are done to try to get good data (to fall back upon in future), if this fails, it would return 0.0% humidity and 0.0 degrees centigrade temperature. But putting an error log for initial reading attempt (and 100 retry failures) is a good idea, will add that.

scripts/plugins/storage/sqlite/init.sql
```diff
@@ -617,6 +617,13 @@ INSERT INTO foglamp.configuration ( key, description, value )
               ' { "plugin" : { "type" : "string", "value" : "omf", "default" : "omf", "description" : "Module that OMF North Plugin will load" } } '
             );
 
+-- dht11 - South plugin for DHT11 - C
```
pintomax: We don't need to add a new entry in the Database.

The caller of this plugin will get the the plugin default config and then call the configuration manager creating/updating the configuration.

Anyway, the DB addition has to be added to postgres as well.

And we need upgrade/downgrade files for both.

I suggest checking the sending_process configuration process.

Soon we will remove unnecessary DB entries.

As long as this is a new "component" we can start with no DB configuration

C/plugins/south/dht11/dht11.cpp
AmandeepArora: If it is first reading, then after 100 failed retries, "0.0% humidity and 0.0 degrees centigrade temperature" is returned. For subsequent readings, last known good data is returned.

scripts/plugins/storage/sqlite/init.sql
AmandeepArora: ok sure.

scripts/plugins/storage/sqlite/init.sql
```diff
@@ -790,6 +801,19 @@ INSERT INTO foglamp.schedules ( id, schedule_name, process_name, schedule_type,
               );
 
 
+-- DHT11 sensor south plugin - C Code
+INSERT INTO foglamp.schedules ( id, schedule_name, process_name, schedule_type,
```
pintomax: We might not need this as well.

This entry can be done via REST API.

You can check this: scripts/extras/foglamp_update

and maybe create a script or provice a copy & paste REST API URL into the README

AmandeepArora: Once changes are done for FOGL-1568, we would have a add_service REST API for C++ services also, then direct DB entries won't be required at all.

scripts/plugins/storage/sqlite/init.sql
AmandeepArora: IIRC, if I didn't provide configuration in init.sql, there was some failure during dht11 south service initialization. "add_service" Rest API is still to be implemented for C++ services in FOGL-1568. Once that is done, we won't need these changes in init.sql files. And I will need to understand upgrade/downgrade files/scripts.

pintomax: For the "configuration" you can do this:

managemnet_client:

addCategory(category, true))

example in:

develop/C/tasks/north/sending_process/sending.cpp#L522

This way you don't need to put it into foglamp.configuration

C/plugins/south/dht11/plugin.cpp
AmandeepArora: plugin_reconfigure is not being called from ManagementApi::configChange( ) at the moment. So I'll address this reconfigure part in a separate ticket, if that's ok with you.

scripts/plugins/storage/sqlite/init.sql
AmandeepArora: Will remove entries from init.sql in one go once we have the fix for FOGL-1568, the way we already have done for python plugins. Also the plugins shall be added by wizard (as per FOGL-1522), so I believe the add service and plugin discovery all shall be done by that itself.

C/plugins/south/dht11/dht11.cpp
MarkRiddoch: The problem with that is there is no indication this is being done so a faulty sensor will never get fixed.

scripts/plugins/storage/sqlite/init.sql
pintomax: ok, right now we have to keep the plugin initial  plugin configuration int the DB

PR #940, opened 2018-07-19
init.sql fixed for configuration key changes for sending process

amarendra-dianomic:
Test result:
`=================================================== 1163 passed, 31 skipped in 85.20 seconds ====================================================
`

PR #12, opened 2017-05-24
Fogl-68

ashish-ScaleDB:
[FOGL-68](https://scaledb.atlassian.net/browse/FOGL-68)

items:
1) tox
2) pylint
3) pytest 

Attached is the [pylint-report.txt](https://github.com/foglamp/FogLAMP/files/1026728/pylint-report.txt)

PR #943, opened 2018-07-20
FOGL-1560 - Drop stream Id from Python sending process arguments

amarendra-dianomic:
This PR enhances sending_process.py as below:
- "stream_id" and "debug_level" command line parameters have been removed from process definitions of Sending Process tasks. Now stream_id is calculated dynamically from "destination_type", explained next, and next available sequence. Now there is no need to hard code and remember stream_id for different processes.
- A new config item "destination_type" has been added to omf and ocs plugins. "destination_type" takes values as 1-OMF, 2-ElasticSearch, 3-OCS. This gives us freedom to add more destinations as needed.
- SendingProcess class now extends FoglampProcess class. Not only this brings uniformity in Task classes, it also makes SendingProcess class current architecture compliant.
- Configuration keys have been redefined to be based upon process names as below:
	    'SENT_PR_1' to 'North Readings to PI'
	    'SENT_PR_2' to 'North Statistics to PI'
	    'SENT_PR_3' to 'North Audit to PI'
	    'SENT_PR_4' to 'North Readings to OCS'
- Statistics keys have been redefined to be based upon process names as below:
	    'SENT_1' to North Readings to PI'
	    'SENT_2' to 'North Statistics to PI'
	    'SENT_3' to 'North Audit to PI'
	    'SENT_4' to 'North Readings to OCS'
- sqlite/init.sql and postgres/init.sql have been modified accordingly.
- 'source' configuration item has been added to all above configuration keys with values as "readings" or "statistics" or "audit" as applicable. This has facilitated to enable all three schedules i.e. 'North Readings to PI', 'North Statistics to PI' and 'North Audit to PI' simultaneously. **This was not possible earlier.**
- test_sending_process.py has been fixed accordingly.
`=================================================== 1158 passed, 36 skipped in 84.86 seconds ====================================================
`

We have tested this PR with our local PI server and everything is working fine. However, we need to investigate and fix below errors:
- 'North Statistics to PI' process is throwing below error for all statistics keys:
```
Jul 20 16:01:54 asinha-ThinkPad-E450 FogLAMP[17261] WARNING: omf: omf_north_5: an error
occurred during the request to the destination - error details |400 {"code":10,"message":"Type
redefinition.","source":"foglamp_data.0001_FOGBENCH/MAGNETOMETER_typename_measurement.1
.0.0.0","description":"OMF type with ID 
'foglamp_data.0001_FOGBENCH/MAGNETOMETER_typename_measurement.1.0.0.0' has been
 already defined differently. Redefinition of the type with the same ID is not allowed."}|
Jul 20 16:01:54 asinha-ThinkPad-E450 FogLAMP[17261] ERROR: omf: omf_north_5: cannot complete
the sending operation - error details |an error occurred during the request to the destination - error
details |400 {"code":10,"message":"Type
redefinition.","source":"foglamp_data.0001_FOGBENCH/MAGNETOMETER_typename_measurement
1.0.0.0","description":"OMF type with ID
'foglamp_data.0001_FOGBENCH/MAGNETOMETER_typename_measurement.1.0.0.0' has been
already defined differently. Redefinition of the type with the same ID is not allowed."}||#012Traceback
(most recent call last):#012  File
"/home/asinha/Development/FogLAMP/python/foglamp/plugins/north/omf/omf.py", line 414, in
plugin_send#012    await omf_north.create_omf_objects(raw_data, config_category_name, type_id)#012 
File "/home/asinha/Development/FogLAMP/python/foglamp/plugins/north/omf/omf.py", line 711, in
create_omf_objects#012    await self._create_omf_objects_automatic(item)#012  File
"/home/asinha/Development/FogLAMP/python/foglamp/plugins/north/omf/omf.py", line 550, in
_create_omf_objects_automatic#012    typename, omf_type = await
self._create_omf_type_automatic(asset_info)#012  File
"/home/asinha/Development/FogLAMP/python/foglamp/plugins/north/omf/omf.py", line 591, in
_create_omf_type_automatic#012    await self.send_in_memory_data_to_picromf("Type",
omf_type[typename])#012  File
"/home/asinha/Development/FogLAMP/python/foglamp/plugins/north/omf/omf.py", line 780, in
send_in_memory_data_to_picromf#012    raise
_error#012foglamp.plugins.north.common.exceptions.URLFetchError: an error occurred during the
request to the destination - error details |400 {"code":10,"message":"Type
redefinition.","source":"foglamp_data.0001_FOGBENCH/MAGNETOMETER_typename_measurement.
.0.0.0","description":"OMF type with ID
'foglamp_data.0001_FOGBENCH/MAGNETOMETER_typename_measurement.1.0.0.0' has been
already defined differently. Redefinition of the type with the same ID is not allowed."}|
Jul 20 16:01:54 asinha-ThinkPad-E450 FogLAMP[17261] ERROR: sending_process:
sending_process_North Statistics to PI: cannot complete the sending operation - error details |an error
occurred during the request to the destination - error details |400 {"code":10,"message":"Type
redefinition.","source":"foglamp_data.0001_FOGBENCH/MAGNETOMETER_typename_measurement.
.0.0.0","description":"OMF type with ID
'foglamp_data.0001_FOGBENCH/MAGNETOMETER_typename_measurement.1.0.0.0' has been
already defined differently. Redefinition of the type with the same ID is not allowed."}||
```
- 'North Audit to PI' related methods i.e. _load_data_into_memory_audit() is currently marked as "TO BE IMPLEMENTED"
- OCS sending process could not be tested locally though ocs.py plugin has been enhanced.
- After this PR is approved/merged, we will need to modify http_north plugin accordingly.


python/foglamp/tasks/north/sending_process.py
```diff
@@ -241,55 +178,35 @@ def handling_input_parameters():
...
-    return param_mgt_name, param_mgt_port, param_mgt_address, stream_id, log_performance, log_debug_level
-
-
-class SendingProcess:
-    """ SendingProcess """
+    return log_performance, log_debug_level
 
-    _logger = None  # type: logging.Logger
 
+class SendingProcess(FoglampProcess):
```
praveen-garg: this resolves FOGL-774

python/foglamp/tasks/north/sending_process.py
```diff
@@ -374,884 +284,654 @@ def __init__(self, loop=None):
...
             raise
+        SendingProcess._logger.debug("Stopped")
 
 
 if __name__ == "__main__":
-
     loop = asyncio.get_event_loop()
     sp = SendingProcess(loop)
-    loop.run_until_complete(sp.start())
+    loop.run_until_complete(sp.run())
```
praveen-garg: glad to see that 320 lines are gone! I hope 1 more iteration will eliminate will remove 300+ more!

python/foglamp/tasks/north/sending_process.py
```diff
@@ -374,884 +284,654 @@ def __init__(self, loop=None):
...
             self._plugin.plugin_shutdown(self._plugin_handle)
         except Exception:
-            _message = _MESSAGES_LIST["e000007"]
-            SendingProcess._logger.error(_message)
-            self._event_loop.run_until_complete(self._audit.failure(self._AUDIT_CODE, {"error - on stop": _message}))
+            SendingProcess._logger.error(_MESSAGES_LIST["e000007"])
+            self._event_loop.run_until_complete(
+                self._audit.failure(self._AUDIT_CODE, {"error - on stop": _MESSAGES_LIST["e000007"]}))
             raise
+        SendingProcess._logger.debug("Stopped")
```
praveen-garg: info? as we have ` SendingProcess._logger.info("Started")`

python/foglamp/tasks/north/sending_process.py
```diff
@@ -374,884 +284,654 @@ def __init__(self, loop=None):
...
-            # logging from different processes
             SendingProcess._logger.removeHandler(SendingProcess._logger.handle)
-            logger_name = _MODULE_NAME + "_" + str(self.input_stream_id)
-
-            SendingProcess._logger = logger.setup(logger_name, destination=_LOGGER_DESTINATION, level=_LOGGER_LEVEL)
-            
-            # TODO: Create a separate "MicroserviceManagementClient" like client for tasks
-            self._core_task_management_client = MicroserviceManagementClient(self._mgt_address, self._mgt_port)
+            logger_name = _MODULE_NAME + "_" + self._name
+            SendingProcess._logger = logger.setup(logger_name, level=logging.INFO if self._debug_level in [None, 0,
```
praveen-garg: why INFO when _debug_level is 1

python/foglamp/tasks/north/sending_process.py
```diff
@@ -303,67 +220,60 @@ class SendingProcess:
...
-            "type": "integer",
-            "default": "1"
-        },
         "source": {
-            "description": "Source of data to be sent on the stream. "
-                           "May be either readings, statistics or audit.",
+            "description": "Source of data to be sent on the stream. May be either readings, statistics or audit.",
             "type": "string",
-            "default": _DATA_SOURCE_READINGS
+            "default": "readings"
```
praveen-garg: so we can delete  `      _DATA_SOURCE_READINGS = "readings" ` ?

python/foglamp/tasks/north/sending_process.py
```diff
@@ -303,67 +220,60 @@ class SendingProcess:
...
+        },
+        'plugin': {
+            'description': 'The name of the translator to use to translate the readings into the output format and send them.',
+            'type': 'string',
+            'default': 'omf'
+        },
+        'filter': {
+            'description': 'Data filtering rules to apply to the data that has been pulled from the storage layer.',
+            'type': 'string',
+            'default': 'None'
```
praveen-garg: I think it should be applied with keys! `applyFilter`? if yes then respect `filterRule` - which has default value `.[]`

python/foglamp/tasks/north/sending_process.py
```diff
@@ -303,67 +220,60 @@ class SendingProcess:
...
+        },
+        'aggregate': {
+            'description': 'Aggregation rules to apply to the data',
+            'type': 'string',
+            'default': 'None'
```
praveen-garg: what is None here?

python/foglamp/tasks/north/sending_process.py
```diff
@@ -303,67 +220,60 @@ class SendingProcess:
...
         self._config = {
             'enable': self._CONFIG_DEFAULT['enable']['default'],
             'duration': int(self._CONFIG_DEFAULT['duration']['default']),
             'source': self._CONFIG_DEFAULT['source']['default'],
             'blockSize': int(self._CONFIG_DEFAULT['blockSize']['default']),
-            'memory_buffer_size': int(self._CONFIG_DEFAULT['memory_buffer_size']['default']),
             'sleepInterval': float(self._CONFIG_DEFAULT['sleepInterval']['default']),
-            'north': self._CONFIG_DEFAULT['north']['default'],
+            'plugin': self._CONFIG_DEFAULT['plugin']['default'],
+            'filter': self._CONFIG_DEFAULT['filter']['default'],
```
praveen-garg: applyFilter and filterRule, I think as of today its applied on each north plugin

python/foglamp/tasks/north/sending_process.py
```diff
@@ -374,884 +284,654 @@ def __init__(self, loop=None):
...
-        Raises:
-        """
-        north_ok = False
+    async def _update_statistics(self, num_sent):
+        """ Updates FogLAMP statistics"""
         try:
-            if self._plugin_info['type'] == self._PLUGIN_TYPE and \
-               self._plugin_info['name'] != "Empty North Plugin":
-                north_ok = True
+            key = self.statistics_key
```
praveen-garg: we can simple use self.statistics_key as key is used once only!

python/foglamp/tasks/north/sending_process.py
```diff
@@ -374,884 +284,654 @@ def __init__(self, loop=None):
...
 
-        except aiohttp.client_exceptions.ClientPayloadError as _ex:
+                        if data_sent:
+                            db_update = True
+                            update_last_object_id = new_last_object_id
+                            tot_num_sent = tot_num_sent + num_sent
+                            self._memory_buffer[self._memory_buffer_send_idx] = None
+                            self._memory_buffer_send_idx += 1
+                            self._task_send_data_sem.release()
+                            self.performance_track("task _task_send_data")
```
praveen-garg: should be the commented code?

python/foglamp/tasks/north/sending_process.py
```diff
@@ -374,884 +284,654 @@ def __init__(self, loop=None):
...
+                            db_update = False
+                        await self._task_fetch_data_sem.acquire()
+
+                    # Updates the Storage layer every 'self.UPDATE_POSITION_MAX' interactions
+                    if db_update:
+                        if update_position_idx >= self.TASK_SEND_UPDATE_POSITION_MAX:
+                            await self._update_position_reached(update_last_object_id, tot_num_sent)
+                            update_position_idx = 0
+                            tot_num_sent = 0
+                            db_update = False
```
praveen-garg: let's make these 4 lines an inner def of this method? and reuse

python/foglamp/tasks/north/sending_process.py
```diff
@@ -374,884 +284,654 @@ def __init__(self, loop=None):
...
+        Todo: TO BE IMPLEMENTED
         """
-        converted_data = []
+        raw_data = None
+        try:
+            # Temporary code
+            if self._module_template != "":
+                raw_data = ""
+            else:
+                raw_data = ""
```
praveen-garg: raw_data = "" on top itself? this method def is not clear even if as TODO! what is a `_module_template`

python/foglamp/tasks/north/sending_process.py
```diff
@@ -374,884 +284,654 @@ def __init__(self, loop=None):
...
-                payload = row['reading']
-                for key in list(payload.keys()):
-                    value = payload[key]
-                    payload[key] = plugin_common.convert_to_type(value)
-
-                # Adds timezone UTC
-                timestamp = apply_date_format(row['user_ts'])
-
+                timestamp = apply_date_format(row['ts'])  # Adds timezone UTC
+                asset_code = row['key'].strip()
```
praveen-garg: these 2 lines can simply be in new_row construction block, the way `reading` key got the value!

python/foglamp/tasks/north/sending_process.py
```diff
@@ -374,884 +284,654 @@ def __init__(self, loop=None):
...
 
-        await self._update_statistics(tot_num_sent, stream_id)
+        try:
+            payload = payload_builder.PayloadBuilder() \
+                .SELECT("id", "type", "description") \
+                .WHERE(['type', '=', destination_type]) \
+                .LIMIT(1) \
+                .payload()
+            destinations = await self._storage_async.query_tbl_with_payload("destinations", payload)
+            rows = destinations['rows']
```
praveen-garg:
```
payload = payload_builder.PayloadBuilder() \
+                .SELECT("id", "type", "description") \
+                .WHERE(['type', '=', destination_type]) \
+                .LIMIT(1) \
+                .payload()
+            destinations = await self._storage_async.query_tbl_with_payload("destinations", payload)
+            rows = destinations['rows']
```
use above code snippet in def ? to avoid duplicate code

python/foglamp/tasks/north/sending_process.py
```diff
@@ -374,884 +284,654 @@ def __init__(self, loop=None):
...
-            _stats = await statistics.create_statistics(self._storage_async)
-
-            await _stats.update(key, num_sent)
+            payload = payload_builder.PayloadBuilder() \
+                .SELECT("key", "description") \
+                .WHERE(['key', '=', self._name]) \
+                .LIMIT(1) \
+                .payload()
+            statistics = await self._storage_async.query_tbl_with_payload("statistics", payload)
+            rows = statistics['rows']
```
praveen-garg: looks add_statistics also calling same snippet!

python/foglamp/tasks/north/sending_process.py
```diff
@@ -374,884 +284,654 @@ def __init__(self, loop=None):
...
+        return north_ok
 
     def _plugin_load(self):
-        """ Loads the plugin
-        Args:
-        Returns:
-        Raises:
-        """
-        module_to_import = self._NORTH_PATH + self._config['north'] + "." + self._config['north']
+        module_to_import = self._NORTH_PATH + self._config['plugin'] + "." + self._config['plugin']
```
praveen-garg: it would be great to use .format style for better readablity

scripts/plugins/storage/postgres/init.sql
```diff
@@ -807,36 +807,44 @@ INSERT INTO foglamp.configuration ( key, description, value )
               ' { "plugin" : { "type" : "string", "value" : "omf", "default" : "omf", "description" : "Module that OMF North Plugin will load" } } '
             );
 
--- SEND_PR_1 - OMF Translator for readings
+-- North Readings to PI - OMF Translator for readings
 INSERT INTO foglamp.configuration ( key, description, value )
-     VALUES ( 'SEND_PR_1',
+     VALUES ( 'North Readings to PI',
```
praveen-garg: fix for FOGL-1565 should go in first to make sure, sending process is working in develop!

python/foglamp/plugins/north/empty/empty.py
```diff
@@ -19,7 +19,11 @@
 _MODULE_NAME = "Empty North Plugin"
 
 _DEFAULT_CONFIG = {
-
+    "destination_type": {
+        "description": "Destination: 1-OMF, 2-Elasticsearch, 3-OCS",
```
praveen-garg: Elasticsearch? are OMF stats are being sent to `Elasticsearch`?

python/foglamp/plugins/north/omf/omf.py
```diff
@@ -131,6 +133,11 @@
         "description": "JQ formatted filter to apply (only applicable if applyFilter is True)",
         "type": "string",
         "default": ".[]"
+    },
+    "destination_type": {
+        "description": "Destination: 1-OMF, 2-Elasticsearch, 3-OCS",
+        "type": "integer",
+        "default": "1"
```
praveen-garg: why do we need this if its OMF and we always it to be 1, same for others

python/foglamp/tasks/north/sending_process.py
amarendra-dianomic: done

python/foglamp/tasks/north/sending_process.py
amarendra-dianomic: done

python/foglamp/plugins/north/empty/empty.py
amarendra-dianomic: Schema definition for foglamp.destinations in init.sql contains:
>type          smallint                    NOT NULL DEFAULT 1,                         -- Enum : 1: OMF, 2: Elasticsearch

python/foglamp/plugins/north/omf/omf.py
amarendra-dianomic: Because, sending_process.py picks/creates destination_id from destination_type (and stream_id from destination_id) from plugin's configuration. Hence it is required to define destination_type in plugin's configuration.

scripts/plugins/storage/postgres/init.sql
amarendra-dianomic: yes, post FOGL-1565 aka #940 merger, we will need to add upgrade/downgrade scripts.

python/foglamp/tasks/north/sending_process.py
praveen-garg: what has been done :) _DATA_SOURCE_READINGS is being used as const in this file. either we should remove them all or keep that here too

python/foglamp/tasks/north/sending_process.py
```diff
@@ -374,884 +279,654 @@ def __init__(self, loop=None):
...
-            elif self._config['source'] == self._DATA_SOURCE_AUDIT:
-                data_to_send = await self._load_data_into_memory_audit(last_object_id)
-            else:
-                _message = _MESSAGES_LIST["e000008"]
-                SendingProcess._logger.error(_message)
-                raise UnknownDataSource
-        except Exception:
-            _message = _MESSAGES_LIST["e000009"]
-            SendingProcess._logger.error(_message)
+            # TODO : FOGL-623 - avoid the update of the field ts when it will be managed by the DB itself
```
praveen-garg: this is DONE as per JIRA

python/foglamp/tasks/north/sending_process.py
```diff
@@ -374,884 +279,654 @@ def __init__(self, loop=None):
...
-                    'id': row['id'],
-                    'asset_code': row['asset_code'],
-                    'read_key': row['read_key'],
-                    'reading': payload,
-                    'user_ts': timestamp
+                    'id': row['id'],  # Row id
+                    'asset_code': asset_code,  # Asset code
+                    'user_ts': timestamp,  # Timestamp
+                    'reading': {'value': row['value']}  # Converts raw data to a Dictionary
                 }
```
praveen-garg: inline comments are not redundant

python/foglamp/tasks/north/sending_process.py
praveen-garg: what is 2 here?

python/foglamp/tasks/north/sending_process.py
```diff
@@ -374,884 +279,654 @@ def __init__(self, loop=None):
...
-                is_started = await self._start(self.input_stream_id)
+                is_started = await self._start()
                 if is_started:
-                    await self.send_data(self.input_stream_id)
-
-                # Stop Sending
+                    await self.send_data()
                 self.stop()
-                SendingProcess._logger.info(_MESSAGES_LIST["i000002"])
+                SendingProcess._logger.info("Execution completed.")
```
praveen-garg: a start to get rid of _MESSAGES_LIST["i000002"] like codes?

python/foglamp/services/core/server.py
```diff
@@ -516,8 +516,16 @@ def _start_core(cls, loop=None):
...
+                # Get total count of streams
+                result = loop.run_until_complete(
+                    cls._storage_client_async.query_tbl_with_payload('streams', total_count_payload))
+                total_streams_count = result['rows'][0]['count']
+
+                # If streams table is non empty, then initialize it
+                if (total_streams_count != 0):
+                    payload = payload_builder.PayloadBuilder().SET(last_object=0, ts='now()').payload()
+                    loop.run_until_complete(cls._storage_client_async.update_tbl("streams", payload))
             else:
```
praveen-garg: Please move it do a def. and call that here

python/foglamp/tasks/north/sending_process.py
```diff
@@ -374,884 +279,652 @@ def __init__(self, loop=None):
...
+    async def send_data(self):
+        """ Handles the sending of the data to the destination using the configured plugin for a defined amount of time"""
 
-        data_sent = False
-        db_update = False
-        update_last_object_id = 0
-        tot_num_sent = 0
-        update_position_idx = 0
+        # Prepares the in memory buffer for the fetch/send operations
+        self._memory_buffer = [None for x in range(self._config['memory_buffer_size'])]
```
praveen-garg: `for _`

python/foglamp/plugins/north/ocs/ocs.py
```diff
@@ -177,7 +178,11 @@
         "type": "string",
         "default": "int32"
     },
-
+    "destination_type": {
```
MarkRiddoch: This is not required. The plugin is an OCS plugin, it does not need to be told it is OCS, and it can never be OMF or Elasticsearch

python/foglamp/plugins/north/omf/omf.py
MarkRiddoch: This is not required. The plugin is an OMF plugin, it does not need to be told it is OMF, and it can never be OCS or Elasticsearch

amarendra-dianomic:
@MarkRiddoch the sequence of operations goes like this:
1) Sending Process fetches last_object_id from streams table for a process. Therefore, each sending process plugin needs an entry in streams table (aka stream_id).
2) streams table has a foreign key destination_id dependency which maps to id of destinations table.
3) destinations table stores id for different destinations e.g. omf, ocs etc.
4) Therefore, to dynamically generate a stream_id for a plugin, we need a destination_id. For destination_id, we compulsorily need a destination_type. This destination_type can either be calculated based upon plugin name (omf, ocs, http_north etc) or should be provided by the plugin itself. I understand that this is not a configurable item as you mentioned, so we will make it "readonly" after FOGL-1432 **OR we can just declare a var in the plugin file itself which then can be fetched by the sending_process so that it stays beyond the reach of the configuration process totally.** 

Please advise.

amarendra-dianomic:
@MarkRiddoch please refer to below ping method:
```
async def ping(request):
    ...
    ...
    def get_sent_stats():
        return sum([int(s['value']) for s in stats if s['key'].startswith('SENT_')])
    ...
    ...

```
Now that we do not have any "SENT_*" statistics keys, how shall we resolve this? Shall we send all North related statistics keys separately in ping?

amarendra-dianomic:
@MarkRiddoch there are now 3 knows issues in this PR:
- sending_process for statistics to PI is throwing error as described in the PR description.
- If we enable both readings to PI and statistics to PI schedules, then I have noticed a clash amongst two processes when both try to create a configuration key "OMF_TYPES" at the same time as below:

```
Jul 24 15:29:39 asinha-ThinkPad-E450 FogLAMP[26389] ERROR: storage_client: foglamp.common.storage_client.storage_client: Error code: 400, reason: Bad Request, details: {'retryable': False, 'entryPoint': 'insert', 'message': 'UNIQUE constraint failed: configuration.key'}
Jul 24 15:29:39 asinha-ThinkPad-E450 FogLAMP[26389] ERROR: configuration_manager: foglamp.common.configuration_manager: Unable to create new category based on category_name OMF_TYPES and category_description OMF Types and category_json_schema {'type-id': {'value': '0001', 'default': '0001', 'type': 'integer', 'description': 'Identify sensor and measurement types'}}#012
Traceback (most recent call last):#012
  File "/home/asinha/Development/FogLAMP/python/foglamp/common/configuration_manager.py", line 161, in _create_new_category#012
    result = await self._storage.insert_into_tbl("configuration", payload)#012
  File "/home/asinha/Development/FogLAMP/python/foglamp/common/storage_client/storage_client.py", line 159, in insert_into_tbl#012
    raise StorageServerError(code=resp.status, reason=resp.reason, error=jdoc)#012
foglamp.common.storage_client.exceptions.StorageServerError: code: 400, reason:Bad Request, error:{'retryable': False, 'entryPoint': 'insert', 'message': 'UNIQUE constraint failed: configuration.key'}#012
#012
During handling of the above exception, another exception occurred:#012
#012
Traceback (most recent call last):#012
  File "/home/asinha/Development/FogLAMP/python/foglamp/common/configuration_manager.py", line 427, in create_category#012
    await self._create_new_category(category_name, category_val_prepared, category_description)#012
  File "/home/asinha/Development/FogLAMP/python/foglamp/common/configuration_manager.py", line 167, in _create_new_category#012
    raise ValueError(err_response)#012
ValueError: {'retryable': False, 'entryPoint': 'insert', 'message': 'UNIQUE constraint failed: configuration.key'}

```
- we need to resolve "dataSend" statistics in ping as mentioned above.

python/foglamp/plugins/north/empty/empty.py
MarkRiddoch: Yes, but the schema is a table that is applicable to all north side destinations, this is a configuration for a single plugin which is designed to work with a single destination type.

amarendra-dianomic:
@MarkRiddoch destination_type removed from plugins. Please check.

python/foglamp/tasks/north/sending_process.py
amarendra-dianomic: yes

amarendra-dianomic:
@MarkRiddoch @praveen-garg @Singhal-Vaibhav feedback changes done. Pl check.

scripts/plugins/storage/sqlite/init.sql
```diff
@@ -680,9 +687,10 @@ INSERT INTO foglamp.scheduled_processes (name, script) VALUES ('restore', '["tas
 
 -- North Tasks
 --
-INSERT INTO foglamp.scheduled_processes ( name, script ) VALUES ( 'North Readings to PI',   '["tasks/north", "--stream_id", "1", "--debug_level", "1"]' );
-INSERT INTO foglamp.scheduled_processes ( name, script ) VALUES ( 'North Readings to OCS',  '["tasks/north", "--stream_id", "4", "--debug_level", "1"]' );
-INSERT INTO foglamp.scheduled_processes ( name, script ) VALUES ( 'North Statistics to PI', '["tasks/north", "--stream_id", "2", "--debug_level", "1"]' );
+INSERT INTO foglamp.scheduled_processes ( name, script ) VALUES ( 'North Readings to PI',   '["tasks/north"]' );
+INSERT INTO foglamp.scheduled_processes ( name, script ) VALUES ( 'North Statistics to PI', '["tasks/north"]' );
+INSERT INTO foglamp.scheduled_processes ( name, script ) VALUES ( 'North Audit to PI',      '["tasks/north"]' );
```
Singhal-Vaibhav: If we have the configuration to define the source (readings, statistics or audit) then why are we creating different schedules for that? I guess the schedule should be 'Data to PI' and we can configure the source with configuration which data to send.

sstefano69:
@amarendra-dianomic 

I tried the curret code available in the PR,
I have got this error :

    Jul 31 08:27:33 ubuntu FogLAMP[11700] ERROR: omf: omf_north_2: cannot complete
    the sending operation - error details |name 'true' is not defined|#012Traceback
    (most recent call last):#012  File
    "/home/foglamp/Development/FogLAMP/python/foglamp/plugins/north/omf/omf.py",
    line 361, in plugin_send#012    await
    omf_north._create_omf_objects(asset_codes_to_evaluate,
    asset_codes_already_created, config_category_name, type_id)#012  File
    "/home/foglamp/Development/FogLAMP/python/foglamp/plugins/north/omf/omf.py",
    line 712, in _create_omf_objects#012    else await
    self._create_omf_type_automatic(item)#012  File
    "/home/foglamp/Development/FogLAMP/python/foglamp/plugins/north/omf/omf.py",
    line 521, in _create_omf_type_automatic#012    "isindex": true#012NameError:
    name 'true' is not defined

    Jul 31 08:27:33 ubuntu FogLAMP[11700] ERROR: sending_process:
    sending_process_North Statistics to PI: cannot complete the sending operation -
    error details |name 'true' is not defined|

note: the code in Develop that I tested yesterday is working fine.

sstefano69:
@amarendra-dianomic 

I think the 'destination' and 'audit' should be fixed in the PR #943  :

    - destination - the 'destination' table/functionality it is not currently used, so the current direction is :
        - now, this PR : create a new row in the stream table having a fixed value = 1, without creating rows  in the destination or any other handling as it was the previous sending process.
        - next step : we will create task/tasks to completely remove the handling for the 'destination' from the DB 
schema and code (python and C++).
    
    - Audit - it is not currently used, so we should remove the temporary code and the related process.


@MarkRiddoch , correct me if I'm wrong.

sstefano69:
stream_id handling, the C++ version of the SP implemented :

the stream_id should be retrieved from the configuration if it is properly defined,
if it is not a new row should be created in the streams table and the newly created id should be used 
and stored in the configuration for the following executions.
    
it seems to me the current implementation does not retrieve/store the stream_id in the configuration manager.

amarendra-dianomic:
@sstefano69 omf plugin accidentally got corrupted, now fixed. All tests are now passing. Please check. Secondly, please check "North Statistics to PI" and "North Readings to PI" together i.e. please enable them together and see if the error is repeated. We should not clean PI system before we enable "North Statistics to PI" because in production, both schedules/processes will/should be running together.

amarendra-dianomic:
@sstefano69 regarding your below comment:
> the stream_id should be retrieved from the configuration if it is properly defined,
> if it is not a new row should be created in the streams table and the newly created id should be used
> and stored in the configuration for the following executions.
> 
> it seems to me the current implementation does not retrieve/store the stream_id in the configuration manager.
> 
it is happening **exactly like this** in this PR. Please check below method in sending_process.py:
```
    async def _get_stream_id(self, destination_id):
        async def get_rows(description):
            payload = payload_builder.PayloadBuilder() \
                .SELECT("id", "description", "active") \
                .WHERE(['description', '=', description]) \
                .payload()
            streams = await self._storage_async.query_tbl_with_payload("streams", payload)
            return streams['rows']

        async def add_stream(destination_id, description):
            payload = payload_builder.PayloadBuilder() \
                .INSERT(destination_id=destination_id,
                        description=description) \
                .payload()
            await self._storage_async.insert_into_tbl("streams", payload)
            rows = await get_rows(description=self._name)
            return rows[0]['id']

        stream_id = None
        try:
            rows = await get_rows(description=self._name)
            if len(rows) == 0:
                stream_id = await add_stream(destination_id, self._name)
                stream_id_valid = True
            elif len(rows) > 1:
                raise ValueError(_MESSAGES_LIST["e000013"].format(stream_id))
            else:
                stream_id = rows[0]['id']
                if rows[0]['active'] == 't':
                    stream_id_valid = True
                else:
                    SendingProcess._logger.info(_MESSAGES_LIST["i000004"].format(stream_id))
                    stream_id_valid = False
        except Exception as e:
            SendingProcess._logger.error(_MESSAGES_LIST["e000013"].format(str(e)))
            raise e
        return stream_id, stream_id_valid

```
Please check.

sstefano69:
possible impact on perfromance:

I saw the transform_in_memory_data has been renamed in create_omf_data
and the check 
    if _log_debug_level == 3:

has been removed.

I don't think it is a good direction as in this way all the transformed rows
will be everytime written to the syslog and writing to the syslong could 
reduce the speed.

in general, all the points where it was used the check

    if _log_debug_level == 3:
    
to conditionally execute that part of code,
mean that code could impact performance
and so execute all the time that code I think it is not a good direction.

amarendra-dianomic:
@sstefano69 regarding your comment:

> possible impact on perfromance:
> 
> I saw the transform_in_memory_data has been renamed in create_omf_data
> 

Yes, transform_in_memory_data() has been renamed to _transform_in_memory_data_readings() just to keep the naming of the methods consistent. I do not think it should give any trouble.

> and the check
> if _log_debug_level == 3:
> 
> has been removed.
> 
> I don't think it is a good direction as in this way all the transformed rows
> will be everytime written to the syslog and writing to the syslong could
> reduce the speed.
> 
> in general, all the points where it was used the check
> 
> if _log_debug_level == 3:
> to conditionally execute that part of code,
> mean that code could impact performance
> and so execute all the time that code I think it is not a good direction.
> 
Please note that **if _log_debug_level == 3:** has been remove en block and not just this line, **so entries will never get into syslog**. Secondly, I believe that all debugging code are part of the development stage and in general, should not go into production. It makes code too heavy. After all, debugging can still be done with logging.info messages and if at all logging.debug level is required, we can always set the level accordingly and introduce logging.debug messages. But this should be kept at minimum. In develop, there are 7 places where **if _log_debug_level == 3:** block is present. If you and @MarkRiddoch feel that these blocks should be retained, it can restored. Please let me know.

Overall, I hope that you understand from above explanation that this PR will not impact performance in any possible way. But as I said, please feel free to advise for reverting the changes in part or in full by  closing this PR in totality. But in that case, you will also need to advise for an alternative way of implementing FOGL-1560.

amarendra-dianomic:
@sstefano69 carried out below modifications as discussed:
- Stream_id, Destination_id now being fetched from Configuration, if available, and new values being written back to configuration.
- Audit key/Schedules removed from source value, DB tables. Related code also removed from sending_process.py. Tests fixed.

Please check.

sstefano69:
- stream_id :
    the latest handling of stream_id is different from what expected :

    the stream_id should be retrieved from the configuration if it is properly defined,
    if it is not a new row should be created in the streams table and the newly created id should be used
    and stored in the configuration for the following executions.

    it seems to me the current implementation manages the stream_id in a different way.

sstefano69:
- destinations, there is still an handling of the table 'destinations', it was expected :

    1) No handling at all of the 'destination' table
    2) creating a new row in the stream table having 'destination-id'  fixed value = 1

sstefano69:
- a minor thing, the latest added staticis are using camel case using space, instead all the other are using Upper case and underscore.

amarendra-dianomic:
@sstefano69 both the above points regarding destination_id and stream_id incorporated in the latest commit. Please check. Regarding the statics keys with spaces, actually this follows from the FOGL-1520 and FOGL-1533.

PR #949, opened 2018-07-24
FOGL 1579

praveen-garg:


python/foglamp/common/plugin_discovery.py
```diff
@@ -58,26 +58,27 @@ def get_plugin_folders(cls, plugin_type):
...
-            import_file_name = "{path}.{dir}.{file}".format(path=plugin_module_path, dir=plugin_name, file=plugin_name)
+            plugin_module_name = plugin_dir
+            import_file_name = "{path}.{dir}.{file}".format(path=plugin_module_path, dir=plugin_dir, file=plugin_module_name)
             _plugin = __import__(import_file_name, fromlist=[''])
 
             # Fetch configuration from the configuration defined in the plugin
             plugin_info = _plugin.plugin_info()
             plugin_config =  {
-                'name': plugin_info['name'],
+                'name': plugin_info['config']['plugin']['default'],
```
praveen-garg: We can fix the design doc to say add `'plugin': plugin_info['config']['plugin']['default']` so that we have name from plugin_info and plugin module to load as  `plugin`

praveen-garg:
@amarendra-dianomic  @MarkRiddoch  Please review

PR #953, opened 2018-07-24
FOGL-1570 JSON config validation fixes

ashish-jabble:
JSON field Display problem pending

### DIFF
--- a/python/foglamp/common/configuration_manager.py
+++ b/python/foglamp/common/configuration_manager.py
@@ -157,16 +157,15 @@ class ConfigurationManager(ConfigurationManagerSingleton):
 
                 num_entries = expected_item_entries.get(entry_name)
                 if set_value_val_from_default_val and entry_name == 'value':
-                    raise ValueError(
-                        'Specifying value_name and value_val for item_name {} is not allowed if desired behavior is to use default_val as value_val'.format(
-                            item_name))
+                    raise ValueError('Specifying value_name and value_val for item_name {} is not allowed if '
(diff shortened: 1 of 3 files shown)


python/foglamp/common/configuration_manager.py
```diff
@@ -764,37 +764,17 @@ def _str_to_ipaddress(item_val):
...
-            return _str_to_json(_value)
+            return Utils.is_json(_value)
         elif _type == 'IPv4' or _type == 'IPv6':
             return _str_to_ipaddress(_value)
 
     def _clean(self, item_type, item_val):
         if item_type == 'boolean':
             return item_val.lower()
-        if item_type == 'JSON':
-            return json.dumps(json.loads(item_val))
```
praveen-garg: to solve the 404: No detail found for the category_name: x and config_item: y
```
if item_type == 'JSON':
    return json.loads(item_val)

```

praveen-garg: For test__validate_category_val_valid_type

```       
 if test_input == "JSON":
            assert c_return_value["test_item_name"]["value"] == json.loads(test_value)
        else:
            assert c_return_value["test_item_name"]["value"] == test_value
```

I'm not sure if we should start returning and accepting the values for boolean, integer value in type based format itself. Else you shall need to use json.dumps for JSON before returning

praveen-garg: https://github.com/foglamp/foglamp-gui/pull/210/files to have quick test and curl you provided!

PR #955, opened 2018-07-25
FOGL-1531 - Bad REST payload

amarendra-dianomic:
Test result:
`=================================================== 1163 passed, 31 skipped in 85.30 seconds ====================================================
`

### DIFF
--- a/python/foglamp/common/microservice_management_client/microservice_management_client.py
+++ b/python/foglamp/common/microservice_management_client/microservice_management_client.py
@@ -232,9 +232,15 @@ class MicroserviceManagementClient(object):
         :param category_data: e.g. '{"key": "TEST", "description": "description", "value": {"info": {"description": "Test", "type": "boolean", "default": "true"}}}'
         :return:
         """
-        url = '/foglamp/service/category'
+        data = json.loads(category_data)
+        if 'keep_original_items' in data:
+            keep_original_item = 'true' if data['keep_original_items'] is True else 'false'
+            url = '/foglamp/service/category?keep_original_items={}'.format(keep_original_item)
+            del data['keep_original_items']
+        else:
+            url = '/foglamp/service/category'
 
-        self._management_client_conn.request(method='POST', url=url, body=category_data)
+        self._management_client_conn.request(method='POST', url=url, body=json.dumps(data))
         r = self._management_client_conn.getresponse()
         if r.status in range(400, 500):
             _logger.error("Client error code: %d, Reason: %s", r.status, r.reason)
--- a/python/foglamp/services/core/api/configuration.py
+++ b/python/foglamp/services/core/api/configuration.py
@@ -84,6 +84,10 @@ async def create_category(request):
     :Example:
             curl -d '{"key": "TEST", "description": "description", "value": {"info": {"description": "Test", "type": "boolean", "default": "true"}}}' -X POST http://localhost:8081/foglamp/category
     """
+    keep_original_items = None
+    if 'keep_original_items' in request.query and urllib.parse.unquote(request.query['keep_original_items']) != '':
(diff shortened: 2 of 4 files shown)


python/foglamp/common/microservice_management_client/microservice_management_client.py
```diff
@@ -232,9 +232,15 @@ def create_configuration_category(self, category_data):
         :param category_data: e.g. '{"key": "TEST", "description": "description", "value": {"info": {"description": "Test", "type": "boolean", "default": "true"}}}'
         :return:
         """
-        url = '/foglamp/service/category'
-
-        self._management_client_conn.request(method='POST', url=url, body=category_data)
+        data = json.loads(category_data)
```
pintomax: Is that needed to keep the old behaviour?

python/foglamp/services/core/api/configuration.py
```diff
@@ -99,9 +103,7 @@
         category_desc = data.get('description')
         category_value = data.get('value')
 
-        should_keep_original_items = data.get('keep_original_items', False)
-        if not isinstance(should_keep_original_items, bool):
-            raise TypeError('keep_original_items should be boolean true | false')
+        should_keep_original_items = True if keep_original_items is not None and keep_original_items.lower() in ['true', '1', 1, 't'] else False
```
pintomax: It looks difficult to read.

It can be split in two or more lines

amarendra-dianomic:
@MarkRiddoch rest call to ConfigurationManager.create_category() method is being made solely by microservice management client only and this has been modified in this PR. Rest of all calls to this method are direct calls (e.g. in Scheduler, add_service(), add_task() etc). So, imho, all refereces have been updated. Please correct me if wrong.

python/foglamp/services/core/api/configuration.py
```diff
@@ -84,6 +84,10 @@
     :Example:
             curl -d '{"key": "TEST", "description": "description", "value": {"info": {"description": "Test", "type": "boolean", "default": "true"}}}' -X POST http://localhost:8081/foglamp/category
     """
+    keep_original_items = None
+    if 'keep_original_items' in request.query and urllib.parse.unquote(request.query['keep_original_items']) != '':
```
praveen-garg: @amarendra-dianomic why did we need this `urllib.parse.unquote` for `request.query['keep_original_items']`

python/foglamp/services/core/api/configuration.py
```diff
@@ -84,6 +84,10 @@
...
+        keep_original_items = urllib.parse.unquote(request.query['keep_original_items'])
```
praveen-garg: why unquote?

python/foglamp/services/core/api/configuration.py
praveen-garg: @amarendra-dianomic 
https://github.com/foglamp/FogLAMP/pull/955/files#diff-8e6a39d1dfaaf562f1c60b2f201539bfR237
only `keep_original_items.lower() in ['true']`  is needed!

python/foglamp/services/core/api/configuration.py
amarendra-dianomic: It is a query parameter and, principally, it should always be parsed and validated before using its value. We should not assume that user will always supply the correct and valid value.

python/foglamp/services/core/api/configuration.py
amarendra-dianomic: See above

python/foglamp/services/core/api/configuration.py
praveen-garg: we  allow `true` or `false`  ONLY HENCE NO! We really don't care as those are not valid requests. and should not be respected and check is not required. 
aiohttp router will send us with proper encoding as is to us.

PR #957, opened 2018-07-25
Fogl 1582

pintomax:


C/services/core/configuration_manager.cpp
```diff
@@ -482,14 +494,13 @@ void ConfigurationManager::mergeCategoryValues(Value& inputValues,
...
 					rapidjson::Writer<rapidjson::StringBuffer> writer(strbuf);
-					Value tmpObj;
 					v.Accept(writer);
 					object.SetString(strbuf.GetString(), allocator);
 					itr->value.AddMember("value", object, allocator);
 
 					break;
 				}
-				// Object & Array not supported yet
+				//  Array and numbers not supported yet
```
MarkRiddoch: We should probably support numbers

PR #961, opened 2018-07-26
Fogl 1549a

ashwinscale:


MarkRiddoch:
@ashwinscale This is now so far out of data I suspect we are best to close the pull request. Is there anything here that we have not already got in the code now?

MarkRiddoch:
This is so far out of date now I think the best thing to do is close it. Ashen you still have the branch if there is anything in that branch that is needed. It will however require a refresh from the current state of development.

PR #964, opened 2018-07-27
FOGL-1432 Merge Config change EPIC branch to develop

MarkRiddoch:


VERSION
```diff
@@ -1,2 +1,2 @@
 foglamp_version=1.3.1
-foglamp_schema=3
+foglamp_schema=4
```
ashish-jabble: Will this continue to grow in develop branch feature by feature (as required)?

MarkRiddoch: Yes, every time we make a schema change we increment the schema version, hopefully we will make less changes as time goes by.

PR #966, opened 2018-07-30
FOGL-1592 restart via admin API

praveen-garg:


praveen-garg:
There is a better version of it https://github.com/foglamp/FogLAMP/pull/968

PR #967, opened 2018-07-30
FOGL-1549 Advantech USB4704 south plugin

MarkRiddoch:


C/plugins/south/USB-4704/plugin.cpp
```diff
@@ -1,6 +1,15 @@
 /*
  * FogLAMP south plugin.
  *
+ * This plugin is for the Advantech USB-4704 Portable Data Acquisition Module.
+ * It supports a numerbof digitial and analogue inputs on that device. The mapping
```
pintomax: typo: numerbof digitial

PR #975, opened 2018-08-01
FOGL-1599

AmandeepArora:
The get_plugin_info utility would be placed in "extras/C" directory in "sudo make install" version.
And at "cmake_build/C/plugins/common/get_plugin_info" in FogLAMP code tree after running make.


C/plugins/common/get_plugin_info.cpp
```diff
@@ -0,0 +1,64 @@
+/*
```
MarkRiddoch: This is in the wrong place - it is not common to plugins, it is a utility. Files in plugins/common is meant to be shared between multiple plugins.

Please add a plugins/utility directory for this.

AmandeepArora: ok sure.

AmandeepArora:
@MarkRiddoch :  I made the required changes, please review.

MarkRiddoch:
In the make rule we should copy the binary to FOGLAMP_ROOT/plugins/south for the plugins and FOGLAMP_ROOT/utils for the utility.

PR #980, opened 2018-08-03
Fogl 1568 - add service to check plugin language

ashish-jabble:
**Fixed items**

~1) Changed default plugin value to same as directory name for C plugins; if and only If we need to use add_service REST endpoint as we have for python south plugins~

2) Utility C modified as needed (Before it was returning only config)

~3) As we have differ path for C libraries and utils file when `make` and `make install`. So did below~
```
       # make
        if os.path.isdir(_FOGLAMP_ROOT + '/cmake_build'):
            arg1 = "cmake_build/C/plugins/utils/get_plugin_info"
            arg2 = "cmake_build/C/plugins/{}/{}/lib{}.so".format(direction, name, name)
        else:
            # sudo make install
            arg1 = "extras/C/get_plugin_info"
            arg2 = "plugins/{}/{}/lib{}.so".format(direction, name, name)
```
4) Fixed plugin discovery tests for api and common one

5) TODO utils and add_service Ctype plugin unit tests

https://scaledb.atlassian.net/browse/FOGL-1643
https://scaledb.atlassian.net/browse/FOGL-1644

NOTE: When make you may see duplicate C-soutth/north plugin entries and that is known issue at python side


python/foglamp/services/core/api/utils.py
```diff
@@ -0,0 +1,30 @@
...
+from foglamp.common import logger
+from foglamp.common.common import _FOGLAMP_ROOT
+
+_logger = logger.setup(__name__)
+
+
+def get_plugin_info(direction, name):
+    try:
+        # make
+        if os.path.isdir(_FOGLAMP_ROOT + '/cmake_build'):
```
MarkRiddoch: We shouldn't really do this, rather n=run the one under FOGLAMP_ROOT. There is nothing that says the developer has to build into cmake_build, that is just a convention in some people use.

python/foglamp/common/plugin_discovery.py
```diff
@@ -57,6 +62,36 @@ def get_plugin_folders(cls, plugin_type):
         else:
             return directories
 
+    @classmethod
+    def get_c_plugin_folders(cls, plugin_type):
+        directories = []
+        if os.path.isdir(utils._FOGLAMP_ROOT + '/cmake_build'):
```
MarkRiddoch: I would not include the cmake_build directories, this is not a recognised location in any way, just somewhere where somebody might build something.

C/plugins/south/modbus/plugin.cpp
```diff
@@ -23,8 +23,8 @@ using namespace std;
 /**
  * Default configuration
  */
-#define CONFIG	"{\"plugin\" : { \"description\" : \"Modbus TCP and RTU plugin\", " \
-			"\"type\" : \"string\", \"default\" : \"foglamp-modbus\" }, " \
+#define CONFIG	"{\"plugin\" : { \"description\" : \"Modbus TCP and RTU C plugin\", " \
+			"\"type\" : \"string\", \"default\" : \"modbus\" }, " \
```
MarkRiddoch: We can't just change the name here without chaining the name we build something as, it will break things

ashish-jabble: I changed default plugin value to same as directory name for C plugins; because binaries name with "lib<directory_name>.so". 
Is there any standard follow for C-plugin default value? As we have for python plugins default value same as directory name and corresponding its py file name (to actually add the service)
I see except the `usb4704` c-south plugin, rest of the plugin starts with "foglamp-"

python/foglamp/services/core/api/utils.py
ashish-jabble: I raised this before in https://github.com/foglamp/FogLAMP/pull/975

python/foglamp/services/core/api/service.py
```diff
@@ -110,17 +111,24 @@ def get_service_records():
...
 
+            script = '["services/south"]' if service_type == 'south' else '["services/north"]'
             # Fetch configuration from the configuration defined in the plugin
             plugin_info = _plugin.plugin_info()
             plugin_config = plugin_info['config']
         except ImportError as ex:
-            raise web.HTTPNotFound(reason='Plugin "{}" import problem from path "{}". {}'.format(plugin, plugin_module_path, str(ex)))
+            # Checking for C-type plugins
+            script = '["services/south_c"]' if service_type == 'south' else '["services/north_c"]'
+            plugin_info = apiutils.get_plugin_info(plugin)
```
ashish-jabble: @MarkRiddoch See the below scenario
south_wizard

With above configuration, plugin is not loaded after enable and start the schedule.
```
Aug  6 19:19:32 nerd-034 C-dumy[6956]: Load south plugin foglamp-dummy.
Aug  6 19:19:32 nerd-034 C-dumy[6956]: PluginManager: Failed to load plugin foglamp-dummy.
Aug  6 19:19:32 nerd-034 C-dumy[6956]: Failed to load south plugin
```
To make it work, I changed the configuration value to "dummy", because C-south plugin requires same name (i.e dummy) as associated library i.e "lib<dummy>.so"

How would user know to change it like above?

ashish-jabble:
@MarkRiddoch C-south plugin works as service when `sudo make install`
But In case of `make`, I need to do following command to make it work;

1) ln -s ${FOGLAMP_ROOT}/cmake_build/C/services/south/south ${FOGLAMP_ROOT}/services/south
```
ls -l services/south
lrwxrwxrwx 1 foglamp foglamp 68 Aug  6 16:49 services/south -> /home/foglamp/Development/FogLAMP/cmake_build/C/services/south/south
```
for installed version (sudo make install), this symbolic link exists

2)  $FOGLAMP_ROOT/scripts/foglamp status is not showing C-south services
```
FogLAMP v1.3.1 running.
FogLAMP Uptime:  731 seconds.
FogLAMP records: 0 read, 0 sent, 0 purged.
FogLAMP does not require authentication.
=== FogLAMP services:
foglamp.services.core
=== FogLAMP tasks:
```
Process exists
```
ps -ef | grep ./services/south
foglamp   3204  1881  0 18:47 ?        00:00:01 ./services/south --port=41315 --address=127.0.0.1 --name=C-dummy
foglamp   7677 14185  0 19:29 pts/4    00:00:00 grep --color=auto ./services/south
```

python/foglamp/services/core/api/utils.py
```diff
@@ -0,0 +1,56 @@
...
+        for fname in files:
+            # C-utility file
+            if fname == name:
+                return os.path.join(path, fname)
+
+
+def find_c_plugin_libs(direction):
+    libraries = []
+    # FIXME: Duplicate binaries found only in case "make",
+    # follow_links=False by default in os.walk() should ignore such symbolic links but tight now its not working
```
praveen-garg: right now

python/foglamp/common/plugin_discovery.py
```diff
@@ -57,6 +62,21 @@ def get_plugin_folders(cls, plugin_type):
         else:
             return directories
 
+    @classmethod
+    def fetch_c_plugins_installed(cls, plugin_type):
+        libraries = utils.find_c_plugin_libs(plugin_type)
+        configs = []
+        for l in libraries:
```
praveen-garg: for l in libs

python/foglamp/common/plugin_discovery.py
```diff
@@ -57,6 +62,21 @@ def get_plugin_folders(cls, plugin_type):
...
+            jdoc = utils.get_plugin_info(l)
+            plugin_config = {'name': l,
+                             'type': plugin_type,
+                             'description': jdoc['config']['plugin']['description'],
+                             'version': jdoc['version']
+                             }
+            if plugin_config is not None:
```
praveen-garg: you mean jdoc not None? and it should be before `plugin_config` assignment?

python/foglamp/services/core/api/utils.py
```diff
@@ -0,0 +1,56 @@
...
+    try:
+        arg1 = _find_c_util('get_plugin_info')
+        arg2 = _find_c_lib(name)
+        cmd_with_args = [arg1, arg2, "plugin_info"]
+        p = subprocess.Popen(cmd_with_args, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        out, err = p.communicate()
+        res = out.decode("utf-8")
+        jdoc = json.loads(res)
+    except (OSError, subprocess.CalledProcessError, Exception) as ex:
+        _logger.exception("C Plugin get info failed due to {}".format(str(ex)))
```
praveen-garg: for logger better we use %s and no  need to use str( ) cast

python/foglamp/services/core/api/utils.py
```diff
@@ -0,0 +1,56 @@
...
+            if fname.endswith(name + '.so'):
+                return os.path.join(path, fname)
+
+
+def _find_c_util(name):
+    for path, subdirs, files in os.walk(_FOGLAMP_ROOT):
+        for fname in files:
+            # C-utility file
+            if fname == name:
+                return os.path.join(path, fname)
```
praveen-garg: it should return None by default?

python/foglamp/services/core/api/utils.py
```diff
@@ -0,0 +1,56 @@
...
+        res = out.decode("utf-8")
+        jdoc = json.loads(res)
+    except (OSError, subprocess.CalledProcessError, Exception) as ex:
+        _logger.exception("C Plugin get info failed due to {}".format(str(ex)))
+        return {}
+    else:
+        return jdoc
+
+
+def _find_c_lib(name):
```
praveen-garg: it should return None by default?

python/foglamp/services/core/api/utils.py
```diff
@@ -0,0 +1,56 @@
...
+    for root, dirs, files in os.walk(_FOGLAMP_ROOT):
+        for name in dirs:
+            if 'plugins' in name:
+                p = os.path.join(root, name) + "/" + direction
+                for path, subdirs, f in os.walk(p):
+                    for fname in f:
+                        # C-binary file
+                        if fname.endswith('.so'):
+                            # Replace lib and .so from fname
+                            libraries.append(fname.replace("lib", "").replace(".so", ""))
```
praveen-garg: use filter? too many for and if

python/foglamp/services/core/api/utils.py
ashish-jabble: Done

python/foglamp/services/core/api/utils.py
ashish-jabble: Done

python/foglamp/services/core/api/service.py
AmandeepArora: @MarkRiddoch 
Having this field as editable could allow the user to set plugin name different than the real lib<plugin_name>.so file present in plugins directory. And that would basically break this flow. Even this value "foglamp-dummy" is not correct to be added in the configuration DB table entry for this plugin.

Is there a reason to have prefix "foglamp-" in the plugin default value in plugin.cpp? Can we just have this value same as library name's middle part? If this is agreeable, probably we can insert this plugin's default name as ${PROJECT_NAME} (i.e. library name's middle part) from C plugin's CMakeLists.txt file.

AmandeepArora:
Barring 1-2 outstanding comments/issues (that will be handled in separate JIRA tickets), PR looks fine to me.

ashish-jabble:
Issues are linked now and will be handled separately
https://scaledb.atlassian.net/browse/FOGL-1643
https://scaledb.atlassian.net/browse/FOGL-1644
https://scaledb.atlassian.net/browse/FOGL-1648
https://scaledb.atlassian.net/browse/FOGL-1649

PR #989, opened 2018-08-08
FOGL-1525

AmandeepArora:
I was able to send coap, sinusoid and dht11 readings from this HTTP north C plugin running with a Foglamp instance running on a ubuntu VM to HTTP south python plugin running on a Foglamp instance running on Raspberry Pi. All types of readings made way to the latter instance correctly. Statistics are also updated correctly.


C/plugins/north/http-north/plugin.cpp
```diff
@@ -0,0 +1,253 @@
...
+
+/**
+ * Plugin specific default configuration
+ */
+#define PLUGIN_DEFAULT_CONFIG 	"\"plugin\": {\"description\": \"HTTP North C Plugin\", \"type\": \"string\", \"default\": \"http-north\"}, " \
+			"\"URL\": { " \
+				"\"description\": \"The URL of the HTTP Connector to send data to\", " \
+				"\"type\": \"string\", " \
+				"\"default\": \"http://localhost:6683/sensor-reading\" }, " \
+			"\"HttpTimeout\": { " \
```
praveen-garg: httpTimeout ? based on camelCase, as we are using for other keys?

C/plugins/north/http-north/plugin.cpp
```diff
@@ -0,0 +1,253 @@
...
+	size_t findProtocol = url.find_first_of(":");
+	string protocol = url.substr(0,findProtocol);
+
+	string tmpUrl = url.substr(findProtocol + 3);
+	size_t findPort = tmpUrl.find_first_of(":");
+	string hostName = tmpUrl.substr(0, findPort);
+
+	size_t findPath = tmpUrl.find_first_of("/");
+	string port = tmpUrl.substr(findPort + 1 , findPath - findPort -1);
+	string path = tmpUrl.substr(findPath);
```
praveen-garg: can this be a utility method? if http sender lib always needs us to split url and find host, port and uri path

C/plugins/north/http-north/plugin.cpp
```diff
@@ -0,0 +1,253 @@
...
+	delete connInfo;
+}
+
+const string& getReadingString(const Reading& reading)
+{
+	string *value = new string;
+	string &m_value = *value;
+
+	// Convert reading data into JSON string
+	m_value.append("{\"timestamp\" : \"" + reading.getAssetDateTime(Reading::FMT_STANDARD) + "Z" + "\"");
```
praveen-garg: if ingested readings has +05:30, then ?

C/plugins/north/http-north/plugin.cpp
```diff
@@ -0,0 +1,253 @@
...
+
+	// Create header for Readings data
+	vector<pair<string, string>> readingData = createMessageHeader();
+
+	// Build a HTTP(S) POST with readingData headers and 'allReadings' JSON payload
+	// Then get HTTPS POST ret code and return 0 to client on error
+	try
+	{
+		int res = connInfo->sender->sendRequest("POST", connInfo->path, readingData, jsonData.str());
+		if (res != 200 && res != 204 && res != 201)
```
praveen-garg: 202 == ACCEPTED, hence should be in too.

C/plugins/north/http-north/plugin.cpp
```diff
@@ -0,0 +1,253 @@
...
+			"\"HttpTimeout\": { " \
+				"\"description\": \"Timeout in seconds for the HTTP operations with the HTTP Connector Relay\", " \
+				"\"type\": \"integer\", \"default\": \"10\" }, " \
+			"\"verifySSL\": { " \
+        			"\"description\": \"Verify SSL certificate\", " \
+				"\"type\": \"boolean\", \"default\": \"False\" }, " \
+			"\"applyFilter\": { " \
+        			"\"description\": \"Whether to apply filter before processing the data\", " \
+				"\"type\": \"boolean\", \"default\": \"False\" }, " \
+			"\"filterRule\": { " \
```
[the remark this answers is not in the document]
AmandeepArora: These 2 config params were present in OMF north plugin, so I just followed that pattern. We probably don't support JQ filters with C++ bindings yet, but we might do so in future. @MarkRiddoch, can you please comment on this?

C/plugins/north/http-north/plugin.cpp
AmandeepArora: I saw the readings with same and correct UTC timestamps consistently at both http-north C (ubuntu VM) as well as http-south python (Raspberry Pi) ends. Is that correct behavior or something needs to change?

C/plugins/north/http-north/plugin.cpp
AmandeepArora: HTTP status code 202 means "The request has been accepted for processing, but the processing has not been completed. The request might or might not be eventually acted upon, and may be disallowed when processing occurs.". 

So it seems possible that eventually the readings are not ingested at the receiver end. And I think we need to be sure about this here.

praveen-garg: Its valid code that your request with data has been accepted and server wants to process it later asynchronously. if server is accepting for later execution, we just need to rely on server that it will handle it and ensure data integrity.

>  The request might or might not be eventually acted upon, and may be disallowed when processing occurs.".

It will be bad server, it should have per-processed the data for its type.

PR #990, opened 2018-08-08
FOGL-1540 URL and enumeration configuration types added with unit tests

ashish-jabble:


Singhal-Vaibhav:
I tried with the following data set and gives 400:  Unrecognized value for item_name...
```
"description": "Test", "type": "URL", "default": "https://pi-server:5460/ingress/messages"
"description": "Test", "type": "URL", "default": "www.google.com"
```
Above are valid URL's as per RFC standard. @MarkRiddoch Are we supporting these?

Singhal-Vaibhav:
Also, we are supporting only http and https protocols (which is mandatory to give), should we be supporting other protocols as well? Source: https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml

python/foglamp/common/configuration_manager.py
```diff
@@ -778,6 +805,15 @@ def _str_to_ipaddress(item_val):
...
             return _str_to_ipaddress(_value)
+        elif _type == 'URL':
+            regex = re.compile(
+                r'^(?:http|coap)s?://'  # http:// or https:// or coap:// or coaps://
+                r'(?:(?:[A-Z0-9](?:[A-Z0-9-]{0,61}[A-Z0-9]\.)?)+(?:[A-Z]{2,6}\.?|[A-Z0-9-]{2,}\.?)|'  # domain
+                r'\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}|'  # ipv4
+                r'\[?[A-F0-9]*:[A-F0-9:]+\]?)'  # ipv6
+                r'(?::\d+)?'  # optional port
+                r'(?:/?|[/?]\S+)$', re.IGNORECASE)
+            return re.match(regex, _value) is not None
```
praveen-garg: why can not we simply use `https://docs.python.org/3/library/urllib.parse.html` 


```
   from urlparse import urlparse

   def url_check(url):
      min_attr = ('scheme' , 'netloc')
       try:
          result = urlparse(url)
          if all([result.scheme, result.netloc]):
              return True
          else:
              return False
     except:
          return False
```

praveen-garg:
limiting schemes will be a bad idea, so better we just check it exists there will be many in our space:


coap/coaps/mqtt/mqtts/ then allowing username:pass in url itself for some

PR #995, opened 2018-08-09
FOGL-1588 - south/server.py enhanced to improve SensorTAG plugin

amarendra-dianomic:
Test result:
`=================================================== 1245 passed, 36 skipped in 97.22 seconds ====================================================
`
This PR must be tested with foglamp/foglamp-south-cc2650poll > FOGL-1588-poll branch, [PR#5](https://github.com/foglamp/foglamp-south-cc2650poll/pull/5), for poll plugin and with foglamp/foglamp-south-cc2650asyn > FOGL-1588-asyn branch for async plugin.

For both CC2650 poll and async plugins, now only temperature sensor is enabled by default. All other sensors must explicitly be enabled via configuration to make them functional. Of course, all sensors can be disabled as well separately via configuration.

This PR makes CC2650 device to connect automatically should it disconnect for any reason. The device can also be disconnected manually and it will reconnect automatically whenever it is switched on, should the South service is still running for this device.

Now Bluetooth address has been made part of asset_code thus enabling multiple devices at the same time. But of course, we need to create separate South processes for each device under different process names.


python/foglamp/services/south/server.py
```diff
@@ -199,6 +205,16 @@ def run(self):
         # This activates event loop and starts fetching events to the microservice server instance
         loop.run_forever()
 
+    async def _restart_plugin(self):
+        try:
+            self._plugin._plugin_stop(self._plugin_handle)
```
praveen-garg: is this standard hook of plugin?

python/foglamp/services/south/server.py
```diff
@@ -58,6 +59,8 @@ class Server(FoglampMicroservice):
 
     _task_main = None
 
+    _restart_config = None
```
praveen-garg: we already have plugin['restart'] = yes | no and return handle with new config why do we need this now again?

python/foglamp/services/south/server.py
```diff
@@ -162,7 +166,7 @@ def __init__(self):
         """
         _LOGGER.info('Started South Plugin: {}'.format(self._name))
         try_count = 1
-        while self._plugin and try_count <= _MAX_RETRY_POLL:
+        while self._plugin:
```
praveen-garg: this is causing a infinite loop for tests on CI (local?)

python/foglamp/services/south/server.py
```diff
@@ -199,6 +212,15 @@ def run(self):
         # This activates event loop and starts fetching events to the microservice server instance
         loop.run_forever()
 
+    async def _stop_and_restart(self):
+        try:
+            self._plugin._plugin_stop(self._plugin_handle)
```
praveen-garg: there is no standard `_plugin_stop`, plugin contract is never supposed to have this method!

amarendra-dianomic:
Will take up improving south/server.py separately vide another ticket.

PR #999, opened 2018-08-10
Thing speak - MATLAB IoT cloud service north plugin

MarkRiddoch:


C/plugins/north/ThingSpeak/ThingSpeak.cpp
```diff
@@ -0,0 +1,166 @@
...
+		}
+	}
+	payload << "]}";
+
+	char url[100];
+	snprintf(url, sizeof(url), "%s/%d/bulk_update.json", m_url.c_str(),
+			m_channel);
+
+	int errorCode;
+	if ((errorCode = m_https->sendRequest("POST", url, m_headers, payload.str())) == 200 || errorCode == 202)
```
AmandeepArora: During testing of http-north plugin, i was getting status code as 201 here - "The request has been fulfilled, resulting in the creation of a new resource.".

praveen-garg: :)

PR #1005, opened 2018-08-13
FOGL-1644 Improve foglamp kill script

MarkRiddoch:
Renamed C++ process as foglamp.services.south, foglamp.services.storage
Fixed issue with add_service not working for C++ south service
Updated foglamp scripts to use new process names


scripts/foglamp
```diff
@@ -399,19 +399,17 @@ foglamp_status() {
...
-                ps -ef | grep "$FOGLAMP_ROOT//services/north" | grep -v 'grep' | awk '{print "foglamp.services.north " $9 " " $10}' || true
-                ps -ef | grep -o 'python3 -m foglamp.services.south.*' | grep -o 'foglamp.services.south.*' | grep -v 'foglamp.services.south\.\*' || true
-                ps -ef | grep -o 'python3 -m foglamp.services.north.*' | grep -o 'foglamp.services.north.*' | grep -v 'foglamp.services.north\.\*' || true
+                ps -ef | grep "foglamp.services.storage" | grep -v 'grep' | grep -v awk | awk '{print "foglamp.services.storage " $9 " " $10}' || true
+                ps -ef | grep "foglamp.services.south" |grep python3| grep -v 'grep' | awk '{print "foglamp.services.south " $11 " " $12 " " $13}' || true
+                ps -ef | grep "foglamp.services.south" |grep -v python3| grep -v 'grep' | grep -v awk | awk '{print "foglamp.services.south " $9 " " $10 " " $11}' || true
+                ps -ef | grep "foglamfoglamp.services.north" | grep -v 'grep' | awk '{print "foglamp.services.north " $9 " " $10}' || true
```
ashish-jabble: grep `foglamp.services.north`

C/plugins/south/dummy/plugin.cpp
```diff
@@ -20,7 +20,7 @@ using namespace std;
 
 
 #define CONFIG	"{\"plugin\" : { \"description\" : \"Dummy C south plugin\", " \
-			"\"type\" : \"string\", \"default\" : \"foglamp-dummy\" } } "
+			"\"type\" : \"string\", \"default\" : \"dummy\" } } "
```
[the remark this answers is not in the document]
MarkRiddoch: It didn't match the actual name, so this was to correct it.

scripts/foglamp
MarkRiddoch: Thanks - fixed

C/plugins/south/dummy/plugin.cpp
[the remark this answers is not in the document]
ashish-jabble: See https://scaledb.atlassian.net/browse/FOGL-1648

python/foglamp/services/core/api/service.py
```diff
@@ -123,8 +123,7 @@ def get_service_records():
         except ImportError as ex:
             # Checking for C-type plugins
             script = '["services/south_c"]' if service_type == 'south' else '["services/north_c"]'
-            plugin_info = apiutils.get_plugin_info(plugin)
-            plugin_config = plugin_info['config']
+            plugin_config = apiutils.get_plugin_info(plugin)
```
ashish-jabble: Please revert this change

C/plugins/south/dummy/plugin.cpp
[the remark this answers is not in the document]
MarkRiddoch: Closed 1648 - the fix was already in progress

python/foglamp/services/core/api/service.py
MarkRiddoch: Without this change the adding of new C services fails because there is no 'config' element in the JSON returned from get_plugin_info

ashish-jabble: config is there,

O/P

plugins/utils/get_plugin_info plugins/south/dummy/libdummy.so plugin_info | jq
```
{
  "name": "Dummy",
  "version": "1.0.0",
  "type": "south",
  "interface": "1.0.0",
  "config": {
    "plugin": {
      "description": "Dummy C south plugin",
      "type": "string",
      "default": "dummy"
    }
  }
}
```
you may also look on https://github.com/foglamp/FogLAMP/blob/develop/C/plugins/utils/get_plugin_info.cpp#L55

MarkRiddoch: Looks like a version issue, previously in develop the config element was missing

python/foglamp/services/core/api/utils.py
```diff
@@ -16,6 +16,7 @@ def get_plugin_info(name):
         out, err = p.communicate()
         res = out.decode("utf-8")
         jdoc = json.loads(res)
+        _logger.error("C Plugin config: ", res)
```
ashish-jabble: This should not be an error, I would suggest use `_logger.debug("C Plugin config: %s", res)`

MarkRiddoch: That was debug that got left accidentally, will remove

PR #1013, opened 2018-08-14
FOGL-1653: Create C++ filter plugin with embedded Python 2.7 interpreter

pintomax:
FOGL-1653: Create C++ filter plugin with embedded Python 2.7 interpreter

 FogLAMP filters dir is now added to "sys.path".

NOTE:
- When error occurs at runtime the input data are passed untouched t next filter
- Python filters are loaded, in this first implementation, from $FOGLAMP_DATA/filters
- Python must have a method called the same as the script file.
- Script name must be specified without ".py" in filter config
- FogLAMP filters dir is added to "sys.path".

Filter, working, example:
```
import sys
import json

"""
filter_config, global variable
"""
filter_config = dict()

"""
Set the Filter configuration into filter_config (global variable)
Input daya is a dict with 'config' key and JSON string version wit data
JSON string is loaded into a dict, set to global variable filter_config
Return True
"""
def set_filter_config(configuration):
    global filter_config
    filter_config = json.loads(configuration['config'])

    return True

"""
FogLAMP Filter

Input is array oof dicts
[
    {'reading': {'random': '980'}, 'asset_code': 'dummy'},
    {'reading': {'random': '960'}, 'asset_code': 'dummy'}
]

Input data can be modified, dropped etc
Output is array of dict
"""

def readings_filter(d):
   # Get list of asset codes to filte
    asset_codes = filter_config['asset_code']

    for elem in d:
    # Use configuration   
    if (elem['asset_code'] in asset_codes):
            reading = elem['reading']

            for key in reading:
                # The following will cause a run time error
                #print '::debug_reading::' + type(reading[key])

                # Altering data: add 5000 to the current value
                newVal = reading[key] + 5000
                reading[key] = newVal

    # Return modified object
    return d
```


C/plugins/filter/python27/plugin.cpp
```diff
@@ -0,0 +1,546 @@
...
+	}
+	else
+	{
+		filtersPath = dataDir;
+	}
+	filtersPath += PYTHON_FILTERS_PATH;
+
+	// Set Python path for embedded Python 2.7
+	// Get current sys.path
+	PyObject* sysPath = PySys_GetObject((char *)string("path").c_str());
```
MarkRiddoch: Where do you decrement the reference count on these objects?

C/common/include/datapoint.h
```diff
@@ -130,6 +130,16 @@ class DatapointValue {
...
 		};
 
+		/**
+		 * Return int value
+		 */
+		int toInt() const { return m_value.i; };
+		/**
+		 * Return double  value
+		 */
+		int toDouble() const { return m_value.f; };
```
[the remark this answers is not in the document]
pintomax: Yes, oversight

C/plugins/filter/python27/plugin.cpp
pintomax: PySys_GetObject returns a borrowed reference. No need to do deal with it

C/plugins/filter/python27/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+# Include header files (first one is python2.7)
+include_directories(/usr/include/python2.7)
+include_directories(include ../../../plugins/filter/common/include ../../../services/common/include ../../../thirdparty/Simple-Web-Server ../../../thirdparty/rapidjson/include ../../../common/include ../../../plugins/common/include)
+
+# Create shared library
+add_library(${PROJECT_NAME} SHARED ${SOURCES} ${plugin_common_src})
+target_link_libraries(${PROJECT_NAME} -lssl -lcrypto)
+
+# Add Python 2.7 library
+target_link_libraries(${PROJECT_NAME} -lpython2.7)
```
MarkRiddoch: Do we need to update the README to add another dependency here?

C/plugins/filter/python27/plugin.cpp
```diff
@@ -0,0 +1,607 @@
...
+
+	/**
+	 * We now pass the filter JSON configuration to the loaded module
+	 */
+	// Set configuration object	
+	PyObject* pConfig = PyDict_New();
+	// Whole configuration as it is
+	string filterConfiguration;
+
+	// Get 'config' filter category confoguration
```
MarkRiddoch: typo

C/plugins/filter/python27/CMakeLists.txt
MarkRiddoch: Can you make this conditional on the python2.7 code/library being installed so as not to break if we don't have it.

Also please test on Raspbian and Ubuntu

C/plugins/filter/python27/plugin.cpp
```diff
@@ -0,0 +1,602 @@
...
+#include <strings.h>
+#include <string>
+#include <iostream>
+#include <filter_plugin.h>
+#include <filter.h>
+#include <reading_set.h>
+
+#include <Python.h>
+
+#define FOGLAMP_ROOT_PATH    "/usr/local/foglamp"
```
MarkRiddoch: We do this stuff to find FOGLAMP_ROOT all over the place, we really should have a common place for it.

pintomax: Yes, FOGL-1699 will take care of this

C/plugins/filter/python27/CMakeLists.txt
pintomax: Added checks in main CMakeLists.txt and plugin one

pintomax: I don't think we need to update the README file right now for Python dependency.

Python 2.7 filter is now built if libraries are found and in case of errors it's not built and errors are shown via CMake system

PR #1014, opened 2018-08-15
FOGL-1588 - Improve Sensortag plugin - Few fixes on south server side

amarendra-dianomic:
As discussed in scrum, we should remove max retry count ceiling for South Poll mode plugins and as advised by @billehunt on [PR#5 for CC2650 plugin](https://github.com/foglamp/foglamp-south-cc2650poll/pull/5), we should remove stack trace from syslog exceptions.

Test result:
`=================================================== 1245 passed, 36 skipped in 98.15 seconds ====================================================
`


python/foglamp/services/south/server.py
```diff
@@ -180,15 +180,18 @@ def __init__(self):
                 # pollInterval is expressed in milliseconds
                 sleep_seconds = int(self._plugin_handle['pollInterval']['value']) / 1000.0
                 await asyncio.sleep(sleep_seconds)
-                # If successful, then set retry count back to 1, meaning that
-                # only in case of 3 successive failures, exit.
-                try_count = 1
```
praveen-garg: why this is not set to 1 now?! and the comment

python/foglamp/services/south/server.py
```diff
@@ -180,15 +180,18 @@ def __init__(self):
...
                 _LOGGER.exception('Key error plugin {} : {}'.format(self._name, str(ex)))
+            except exceptions.QuietError:
+                try_count = 2
+                await asyncio.sleep(_TIME_TO_WAIT_BEFORE_RETRY)
             except (Exception, RuntimeError, exceptions.DataRetrievalError) as ex:
-                try_count += 1
-                _LOGGER.exception('Failed to poll for plugin {}, retry count: {}'.format(self._name, try_count))
+                try_count = 2
+                _LOGGER.error('Failed to poll for plugin {}'.format(self._name))
+                _LOGGER.debug('Exception poll plugin {}'.format(str(ex)))
```
praveen-garg: only error ?! and actually exceptions

python/foglamp/services/south/server.py
amarendra-dianomic: Because it is already set to 1 at start.

python/foglamp/services/south/server.py
amarendra-dianomic: Error hides the stack trace. Since this error may come in plenty, we don't need these verbose errors each time and crowd the syslog.

amarendra-dianomic: To avoid crowding of syslog, detailed error from plugin vide debug mode only .

python/foglamp/services/south/server.py
amarendra-dianomic: Why would we want to stop it? The process will continue to run and we can switch the sensortag off. To stop the process altogether, we will need to terminate the process, like any other South plugin.

PR #1016, opened 2018-08-17
FOGL-1618: Missing Statistics in C++ south service

AmandeepArora:
I have chosen statistics table row key as "SOUTH_READINGS_FROM_DUMMY" on the lines of NORTH_READINGS_TO_PI and NORTH_READINGS_TO_HTTP keys.


C/services/south/ingest.cpp
```diff
@@ -28,6 +28,157 @@ static void ingestThread(Ingest *ingest)
...
+/**
+ * Create a row for given assetName in statistics DB table, if not present already
+ * The key checked/created in the table is "SOUTH_READINGS_FROM_<assetName>"
+ * 
+ * @param assetName     Asset name for the plugin that is sending readings
+ */
+int Ingest::CreateStatsDbEntry(const string& assetName)
+{
+	// Prepare foglamp.statistics update
+	string statistics_key = "south_readings_from_" + assetName;
```
MarkRiddoch: Could we make this a bit shorter, maybe just the asset name to INGEST_<assetName> - the risk is with a long asset name we end up overflowing the database column.

C/services/south/ingest.cpp
```diff
@@ -28,6 +28,157 @@ static void ingestThread(Ingest *ingest)
...
+
+	string key;
+	const Condition conditionStat(Equals);
+	
+	try
+		{
+		if (m_newReadings)
+			{
+			// Prepare foglamp.statistics update
+			key = "South_readings_from_" + m_readingsAssetName;
```
MarkRiddoch: As above, a shorter name would be good.

C/services/south/ingest.cpp
```diff
@@ -28,6 +28,157 @@ static void ingestThread(Ingest *ingest)
...
+	}
+	return 0;
+}
+
+/**
+ * Thread to update statistics table in DB
+ */
+static void statsThread(Ingest *ingest)
+{
+	while (ingest->running())
```
MarkRiddoch: The thread should sleep for a while, on the Python side there is a configuration entry for how long between statistics updates.

AmandeepArora: ingest->updateStats() basically waits on a condition variable - m_statsCv. And this condition variable is notified by the ingest/other thread at the end of processQueue function. So this thread also sleeps for around same time as the ingest thread. And that time comes from config->maxSendLatency. That's 5 seconds by default. I initially implemented a separate sleeping time for this thread, but then, any snapshot of statistics table won't reflect a consistent picture (input from Praveen). So implemented the way it is presently, same sleep time for both threads.

C/services/south/ingest.cpp
AmandeepArora: Ok sure.

AmandeepArora:
Review comments incorporated.

C/services/south/ingest.cpp
MarkRiddoch: OK, thanks - missed that

PR #1018, opened 2018-08-17
FOGL-1640 - server side enhancements to add group (aka parent) info to schedules list

amarendra-dianomic:
Test result:
`=================================================== 1286 passed, 36 skipped in 95.44 seconds ====================================================
`

### DIFF
--- a/python/foglamp/services/core/api/service.py
+++ b/python/foglamp/services/core/api/service.py
@@ -161,7 +161,12 @@ async def add_service(request):
                                              category_description=category_desc,
                                              category_value=plugin_config,
                                              keep_original_items=True)
+            # Create the parent category for all South services
+            await config_mgr.create_category("South", {}, "South microservices", True)
+            await config_mgr.create_child_category("South", [name])
         except Exception as ex:
+            await revert_configuration(storage, name)  # Revert configuration entry
+            await revert_parent_child_configuration(storage, name)
             await revert_scheduled_processes(storage, plugin)  # Revert scheduled_process entry
             raise web.HTTPInternalServerError(reason='Failed to create plugin configuration. {}'.format(str(ex)))
 
@@ -180,10 +185,12 @@ async def add_service(request):
             schedule = await server.Server.scheduler.get_schedule_by_name(name)
         except StorageServerError as ex:
             await revert_configuration(storage, name)  # Revert configuration entry
+            await revert_parent_child_configuration(storage, name)
             await revert_scheduled_processes(storage, name)  # Revert scheduled_process entry
             raise web.HTTPInternalServerError(reason='Failed to created schedule. {}'.format(ex.error))
         except Exception as ins_ex:
             await revert_configuration(storage, name)  # Revert configuration entry
+            await revert_parent_child_configuration(storage, name)
             await revert_scheduled_processes(storage, name)  # Revert scheduled_process entry
             raise web.HTTPInternalServerError(reason='Failed to created schedule. {}'.format(str(ins_ex)))
 
@@ -213,3 +220,8 @@ async def revert_scheduled_processes(storage, process_name):
 async def revert_configuration(storage, key):
     payload = PayloadBuilder().WHERE(['key', '=', key]).payload()
     await storage.delete_from_tbl('configuration', payload)
+
+
+async def revert_parent_child_configuration(storage, key):
+    payload = PayloadBuilder().WHERE(['parent', '=', "South"]).AND_WHERE(['child', '=', key]).payload()
+    await storage.delete_from_tbl('category_children', payload)
--- a/python/foglamp/services/core/api/task.py
+++ b/python/foglamp/services/core/api/task.py
@@ -40,11 +40,7 @@ async def add_task(request):
         "schedule_day": 0,
         "schedule_time": 0,
         "schedule_repeat": 30,
-        "schedule_enabled": true,
-        "cmd_params": {
-            "stream_id": "1",
-            "debug_level": "1"
-        }
+        "schedule_enabled": true
      }'
     """
 
@@ -178,7 +174,12 @@ async def add_task(request):
                                              category_description=category_desc,
                                              category_value=plugin_config,
                                              keep_original_items=True)
+            # Create the parent category for all North tasks
+            await config_mgr.create_category("North", {}, 'North tasks', True)
+            await config_mgr.create_child_category("North", [name])
         except Exception as ex:
+            await revert_configuration(storage, name)  # Revert configuration entry
+            await revert_parent_child_configuration(storage, name)
             await revert_scheduled_processes(storage, plugin)  # Revert scheduled_process entry
             raise web.HTTPInternalServerError(reason='Failed to create plugin configuration. {}'.format(str(ex)))
 
@@ -202,10 +203,12 @@ async def add_task(request):
             schedule = await server.Server.scheduler.get_schedule_by_name(name)
         except StorageServerError as ex:
             await revert_configuration(storage, name)  # Revert configuration entry
+            await revert_parent_child_configuration(storage, name)
             await revert_scheduled_processes(storage, name)  # Revert scheduled_process entry
             raise web.HTTPInternalServerError(reason='Failed to created schedule. {}'.format(ex.error))
         except Exception as ins_ex:
             await revert_configuration(storage, name)  # Revert configuration entry
+            await revert_parent_child_configuration(storage, name)
             await revert_scheduled_processes(storage, name)  # Revert scheduled_process entry
             raise web.HTTPInternalServerError(reason='Failed to created schedule. {}'.format(str(ins_ex)))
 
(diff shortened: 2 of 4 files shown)


scripts/plugins/storage/sqlite/init.sql
```diff
@@ -918,3 +918,12 @@ INSERT INTO foglamp.schedules ( id, schedule_name, process_name, schedule_type,
...
               );
+
+-- Insert into category_children
+INSERT INTO foglamp.category_children (parent, child) VALUES ('South', 'dht11');
+INSERT INTO foglamp.category_children (parent, child) VALUES ('North', 'North_Readings_to_PI');
+INSERT INTO foglamp.category_children (parent, child) VALUES ('North', 'North_Statistics_to_PI');
+INSERT INTO foglamp.category_children (parent, child) VALUES ('North', 'North_Readings_to_HTTP');
+INSERT INTO foglamp.category_children (parent, child) VALUES ('North', 'North Readings to PI');
+INSERT INTO foglamp.category_children (parent, child) VALUES ('North', 'North Statistics to PI');
+INSERT INTO foglamp.category_children (parent, child) VALUES ('North', 'North Readings to OCS');
```
praveen-garg: This will allow us to see these as child if south and north in config tree too. and not in root=true!  

BTW. we also need upgrade/downgrade sql scripts for these

scripts/plugins/storage/postgres/init.sql
```diff
@@ -1117,3 +1117,12 @@ INSERT INTO foglamp.schedules ( id, schedule_name, process_name, schedule_type,
                 true,                                   -- exclusive
                 false                                   -- disabled
               );
+
+-- Insert into category_children
```
MarkRiddoch: These should not be in init.sql - we are making an effort to move away form fixed names in init.sql and thus this is a step in the wrong direction.

python/foglamp/services/core/scheduler/scheduler.py
```diff
@@ -940,10 +941,38 @@ def _schedule_row_to_schedule(cls,
...
+            psch.name = sch.name
+            psch.exclusive = sch.exclusive
+            psch.enabled = sch.enabled
+            psch.repeat = sch.repeat
+            psch.time = sch.time
+            psch.day = sch.day
+            psch.process_name = sch.process_name
+            psch.schedule_type = sch.schedule_type
+            # TODO: After scheduler bug is fixed, change process_name to name
+            psch.group = get_children[sch.process_name] if sch.process_name in get_children else ""
```
MarkRiddoch: This is mixing the concept of configuration hierarchy and schedules in a way that was never designed. The assumption that each schedule corresponds to a configuration category is probably not a good one.

scripts/plugins/storage/postgres/downgrade/8.sql
```diff
@@ -0,0 +1,8 @@
+-- Remove from category_children
```
MarkRiddoch: These fixed names should not be used, we are moving away from them with the introduction of the add service and the and task mechanism. These calls should be the way in which these entries get populated.

scripts/plugins/storage/postgres/upgrade/9.sql
```diff
@@ -0,0 +1,8 @@
+-- Insert into category_children
```
MarkRiddoch: These fixed names should not be used, we are moving away from them with the introduction of the add service and the and task mechanism. These calls should be the way in which these entries get populated.

scripts/plugins/storage/sqlite/downgrade/8.sql
```diff
@@ -0,0 +1,8 @@
+-- Remove from category_children
```
MarkRiddoch: These fixed names should not be used, we are moving away from them with the introduction of the add service and the and task mechanism. These calls should be the way in which these entries get populated.

scripts/plugins/storage/sqlite/init.sql
```diff
@@ -918,3 +918,12 @@ INSERT INTO foglamp.schedules ( id, schedule_name, process_name, schedule_type,
                 't',                                   -- exclusive
                 'f'                                   -- disabled
               );
+
+-- Insert into category_children
```
MarkRiddoch: These fixed names should not be used, we are moving away from them with the introduction of the add service and the and task mechanism. These calls should be the way in which these entries get populated.

scripts/plugins/storage/sqlite/upgrade/9.sql
```diff
@@ -0,0 +1,8 @@
+-- Insert into category_children
```
MarkRiddoch: These fixed names should not be used, we are moving away from them with the introduction of the add service and the and task mechanism. These calls should be the way in which these entries get populated.

MarkRiddoch:
I am unclear what problem you are trying to solve with this.

Its it that you want to know which schedule relates to a particular south service? Or get a list of schedules related to south services?

If it is the later then I would suggest you make an API that does what you want rather than make the schedule appear to use the configuration hierarchy. Possible enhance the /foglamp/service/south  API that @ashish-jabble is working on to include the schedule in the data returned for each south service.

amarendra-dianomic:
@MarkRiddoch please note that:
- First, the need to know whether a schedule belongs to "North" or "South" arouse from FOGL-1640 GUI requirement. Currently, schedule list api does not supply this info.
- Currently, as you are aware of, whenever a North/South schedule is added, we compulsorily make a related entry in configuration table under "process_name" of the schedule (although this is bug and will be fixed separately). 
- The reason for the previous version of this PR simply tried to use, enhance and exploit this relationship further. For example, whenever a South service is run or a North task is run, a corresponding entry is made in category_children table with either "South" or "North" as parent and schedule name as child. In the process, either "North" or "South" configuration category does get added to the configuration table. The previous version only tried to kind of preponing of recording of this relation in category_children FROM at the time when the service or task is run TO at the time when the service or task is created. This would not have changed the functioning of the service or task in any way. Since we do not add some pre-existing North/South schedules vide add_service()/add_task() call, we needed to add entries to category_children table purely for GUI requirements. Hence changes were made in init.sql and new upgrade/downgrade sqls were added. However all this has been reverted.
- An alternative to all above process has been implemented which calculates group as below::
  - Fetch plugins list with North/South info from plugin_discovery module.
  - Fetch plugin name from configuration table from category = schedule's process_name. 
  - Locate schedule's plugin name in plugins_discovery module's list to determine North/South group.
  The only issue with this approach is that it is an expensive call as compared to catergory_children table method.

 Please check.

MarkRiddoch:
I guess the real question is why is group needed at all?

Not every schedule is related to a south service or a north task, so this usage has no meaning for them. Schedules are related to things that run at a point, or many points , in time. They are not related to services or tasks other than that they provide the mechanism to get them run. It is the service or task that is related to a function, like north or south.

amarendra-dianomic:
@MarkRiddoch As I mentioned before, we need group info for fulfil GUI requirements per FOGL-1640. GUI needs to filter out all North/South schedules for the landing scheduler page.

MarkRiddoch:
The concept here is completely wrong - there is no such thing as a group of schedules - the API and data model is being broken by a desire to fit a GUI screen rather than the GUI being built to use the data model. This is wrong.

The link is that the schedule_name of a schedule is the --name parameter passed to the service or task. This name is also the name of the configuration category that the service or task loads. There is already enough information to link schedules to services and hence, because you have the names in the configuration hierarchy you can tell which ones belong to the South and which belong to the North.

There is no need for this spurious group property to be added to schedules.

praveen-garg:
We need to think how to do it in https://github.com/foglamp/FogLAMP/pull/1026 so that gui does not make multiple calls to fetch schedules then scheduled processes then filter and then find stats. Similarly for filtering schedules at schedules tab. I will think over it. We need to have some endpoint that prepare a data model for gui, instead of having round trips of calls.

amarendra-dianomic:
@MarkRiddoch we are then left with only three ways to fulfil GUI requirements:
- Either we modify api for get_scheduled_processes() to supply "script" info also so that GUI can calculate "group" info, OR
- If we do not want to expose "script" info, then we can create a new api like get_scheduled_processes_type() which will provide a list of scheduled_process, type (North/South/Other), OR
- GUI can calculate group/category/type using this time consuming procedure:
   1. Fetch schedules
   2. Fetch plugins installed list
   3. Fetch configuration for each schedule by category = process_name and item = "plugin"
   4. If "plugin" is not there then its group is "other" else find "default" value of "plugin" and match it with plugins_installed list to find the group

First two methods will not hit DB as all info is maintained by Scheduler in memory only.

Please advise.

amarendra-dianomic:
@MarkRiddoch 1) by "script"  I mean "script" column of scheduled_processes table. 2) yes, it will be possible to create the new api as proposed but there is only one issue and which is that "South" parent is created only when a microservice is run and not when just it is created. Similarly, "North" parent is created only a North task is run. GUI needs the category regardless whether the microservice or task has run or not.

MarkRiddoch:
OK, so what you really need is a list of the South and the North schedules to exclude. There are several ways to solve this in the GUI end.

Fetch three list

1. All schedules
2. Children of South configuration category
3. Children of North configuration category

Display all the schedules in list 1 unless the schedule_name appears in either list 2 or list 3

If three calls are two many then do it with two and get the full category tree and process that rather than the 2 lists.

amarendra-dianomic:
@MarkRiddoch your suggested method will work only for those schedules (children of South/North) which have been **run at least once** (as then only respective entry is made in configuration table) and not for those which remain disabled, e.g. North schedules. GUI wants schedules **regardless of the run status**. How to find this?

amarendra-dianomic:
@MarkRiddoch changes made as advised. Please check.

PR #1019, opened 2018-08-17
fixed indexing error for SQLite

praveen-garg:


praveen-garg:
@pintomax  thanks for quick review!

PR #1020, opened 2018-08-19
South services list with tracked assets and corresponding readings count [FOGL-1633]

praveen-garg:


python/foglamp/services/core/api/south.py
```diff
@@ -0,0 +1,84 @@
...
+        for service_record in svc_records:
+            sr_list.append(
+                {
+                    'name': service_record._name,
+                    'address': service_record._address,
+                    'management_port': service_record._management_port,
+                    'service_port': service_record._port,
+                    'protocol': service_record._protocol,
+                    'status': ServiceRecord.Status(int(service_record._status)).name.lower(),
+                    'assets': await _get_tracked_assets_and_readings(storage_client, service_record._name)
```
MarkRiddoch: Why limit this to those that are currently registered? We may have buffered readings for south services that are not currently running.

python/foglamp/services/core/routes.py
```diff
@@ -91,6 +92,9 @@ def setup(app):
     app.router.add_route('POST', '/foglamp/service', service.add_service)
     app.router.add_route('GET', '/foglamp/service', service.get_health)
 
+    # South
+    app.router.add_route('GET', '/foglamp/south', south.get_south_services)
```
MarkRiddoch: I would suggest if this is getting services in the  south then it would be better named as /foglamp/service/south so that it is clear it is service related.

ashish-jabble: Instead of creating new route, will be better if we could handle with an optional param `type` in foglamp.services.core.api.service.get_health.   i.e `curl -X GET http://localhost:8081/foglamp/service?type=Southbound | jq`

MarkRiddoch: Yes, sure. Please can you talk with Amarendra about the change he is proposing. I think that what he wants can also be obtained via this mechanism if we include the schedule in the health of a service.

python/foglamp/services/core/api/south.py
praveen-garg: I can drive it that way, 1) fetch all services & asset, fetch corresponding readings 2) fetch services from servicesRegistry and merge that info 1 as applicable.

In this case, we will have no info for ports, protocol, status.

python/foglamp/services/core/routes.py
praveen-garg: I kept this specifically for `/south` and `/north` needs in mind. 

Northbound tasks are not services, but that need an aggregated fetch for all north schedules, its enabled/disabled status, sent stats.

MarkRiddoch: True, but we are going to end up with a very strange API here because of the way we are mapping this articular client to the API. In order to move forward lets keep whatever works but as undocumented API until we decide what we really want.

python/foglamp/services/core/api/south.py
praveen-garg: As discussed and with FOGL-1701, we should be good with it for now. Or we want to get old services as well, which were b/w restart.

praveen-garg: @MarkRiddoch  addressed in https://github.com/foglamp/FogLAMP/pull/1029

MarkRiddoch:
Given this is merging a branch into develop for which there is another pull request to merge into this source branch I guess we should hold this until the 1029 merge is complete.

python/foglamp/services/core/api/scheduler.py
```diff
@@ -339,6 +341,96 @@ def _extract_args(data, curr_value):
...
+        if not sch_name and not sch_id:
+            raise web.HTTPBadRequest(reason='Schedule name or ID is required')
+
+        if sch_name and not sch_id:
+            storage_client = connect.get_storage_async()
+            payload = PayloadBuilder().SELECT("id").WHERE(['schedule_name', '=', sch_name]).payload()
+            result = await storage_client.query_tbl_with_payload('schedules', payload)
+
+            if int(result['count']):
+                sch_id = result['rows'][0]['id']
```
ashish-jabble: Handle the case `count == 0`; otherwise it will throw 500 internal server error (i.e **sch_id** is None for this case)
```
{
    "error": {
        "message": "[TypeError] one of the hex, bytes, bytes_le, fields, or int arguments must be given"
    }
}
```

We should `raise ValueError('No schedule found')`

python/foglamp/services/core/api/scheduler.py
```diff
@@ -339,6 +341,96 @@ def _extract_args(data, curr_value):
...
+
+        if sch_id:
+            try:
+                assert uuid.UUID(sch_id)
+            except ValueError as ex:
+                raise web.HTTPNotFound(reason="No Schedule with ID {}".format(sch_id))
```
ashish-jabble: `uuid.UUID(sch_id)` -- This will throw ValueError if invalid, no need to assert here

python/foglamp/services/core/api/scheduler.py
```diff
@@ -339,6 +341,96 @@ def _extract_args(data, curr_value):
...
+        if not sch_name and not sch_id:
+            raise web.HTTPBadRequest(reason='Schedule name or ID is required')
+
+        if sch_name and not sch_id:
+            storage_client = connect.get_storage_async()
+            payload = PayloadBuilder().SELECT("id").WHERE(['schedule_name', '=', sch_name]).payload()
+            result = await storage_client.query_tbl_with_payload('schedules', payload)
+
+            if int(result['count']):
+                sch_id = result['rows'][0]['id']
```
ashish-jabble: same as above

python/foglamp/services/core/api/scheduler.py
praveen-garg: when result['count'] value is 0, then it will be `if 0: `

python/foglamp/services/core/api/scheduler.py
praveen-garg: when result['count'] value is 0, then it will be `if 0:`

python/foglamp/services/core/api/scheduler.py
ashish-jabble: Then, when it reach to code line `assert uuid.UUID(None)`; it will throw TypeError

PR #1026, opened 2018-08-20
North schedules with stats info [FOGL-1635]

praveen-garg:
Depends on FOGL-1640. Once FOGL-1640 stuff is merged into develop, this shall need a rebase.


python/foglamp/services/core/routes.py
```diff
@@ -91,6 +92,9 @@ def setup(app):
     app.router.add_route('POST', '/foglamp/service', service.add_service)
     app.router.add_route('GET', '/foglamp/service', service.get_health)
 
+    # North
+    app.router.add_route('GET', '/foglamp/north', north.get_north_schedules)
```
praveen-garg: suggest a better name!

python/foglamp/services/core/api/north.py
```diff
@@ -0,0 +1,88 @@
...
+        stats = await _get_sent_stats()
+
+        for sch in north_schedules:
+            # FIXME: sch.name vs sch.processName?
+            stat = next((s for s in stats if s["key"] == sch["name"]), None)
+            if stat:
+                sch["sent"] = stat["value"]
+            else:
+                # FIXME: what if no key && schedule match in statistics found
+                sch["sent"] = -1
```
praveen-garg: what if a disabled schedule exists and no stats created yet; so 0 as default or -1

python/foglamp/services/core/api/north.py
```diff
@@ -0,0 +1,88 @@
...
+    :Example:
+            curl -X GET http://localhost:8081/foglamp/north
+    """
+    try:
+        north_schedules = await _get_north_schedules()
+        stats = await _get_sent_stats()
+
+        for sch in north_schedules:
+            # FIXME: sch.name vs sch.processName?
+            stat = next((s for s in stats if s["key"] == sch["name"]), None)
```
amarendra-dianomic: Safe to use sch["processName"], with a TODO, for now as this is due to a bug in the scheduler code which will be removed in a near future sprint.

python/foglamp/services/core/api/north.py
amarendra-dianomic: IMHO, 0 is more intuitive as we also provide a "enabled" col.

praveen-garg:
Agree @MarkRiddoch ! these shall really need better and thoughtful treatment, Hence just 2 seprate endpoint route and 2 module files.

python/foglamp/services/core/api/north.py
praveen-garg: gui shall only show if `>=0`

PR #1029, opened 2018-08-21
 get all exising and old services tracked assets 

praveen-garg:


Singhal-Vaibhav:
However, I am not sure what should happen in the following scenario:
1. Add one south service with name **ccpoll** (plugin cc2650)
2. Set asset prefix as st1-
3. Enable
4. Shut down the service ccpoll
5. Add another service with name **ccpoll2** (plugin cc2650)
6. Set asset prefix as st1- (same as of above)
7. Enable
8. curl -X GET http://localhost:8081/foglamp/south|jq
```
{
  "services": [
    {
      "assets": [
        {
          "asset": "st1-temperature",
          "count": 38
        }
      ],
      "service_port": 41607,
      "status": "down",
      "address": "0.0.0.0",
      "protocol": "http",
      "name": "ccpoll",
      "management_port": 41607
    },
    {
      "assets": [
        {
          "asset": "st1-temperature",
          "count": 38
        }
      ],
      "service_port": 33173,
      "status": "running",
      "address": "0.0.0.0",
      "protocol": "http",
      "name": "ccpoll2",
      "management_port": 33173
    }
  ]
}
```
You will see the same read count for asset (st-1) in different services. Since we don't have internal mapping 1-1 of service with asset name, is this the desired behaviour?

amarendra-dianomic:
I agree with @Singhal-Vaibhav 's observation as above. This seems to be a fundamental issue regardless of his PR. If we want to add two identical sensors via the same plugin, then we MUST set the assetPrefix value differently before we enable the schedules, otherwise in both readings table and in statistics table, asset keys will get mixed up and it will be impossible to distinguish readings/statistics from the two or more identical sensors.  The problem is that configuration table does not start storing the assetPrefix value prior to the start of schedule. Otherwise this PR LGTM.

praveen-garg:
As of today, the only way is to name assetPrefix (if available in plugin) correctly in config when service is enabled. So Address like BLE mac| IP | Port all such things will be controlled based on config only. /cc @MarkRiddoch

PR #1033, opened 2018-08-22
FOGL-1657: Add asset tracking to the C south service

AmandeepArora:
I am able to fetch and add Asset tracker entries for "Ingest" event in C++ south service. A cache (in the form of a hash table) of these asset tracking entries is also maintained to speed up operation since lookup needs to be done for each reading object. Cache is initially populated using relevant entries from output of "GET /foglamp/track" and then maintained as new entries are seen in ingestion process.


C/services/south/include/ingest.h
```diff
@@ -53,6 +63,9 @@ class Ingest {
 	unsigned long			m_timeout;
 	unsigned int			m_queueSizeThreshold;
 	bool				m_running;
+	std::string 		m_serviceName;
```
MarkRiddoch: minor point - formatting

C/services/south/ingest.cpp
```diff
@@ -28,6 +28,76 @@ static void ingestThread(Ingest *ingest)
...
+ *
+ * @param m_mgtClient	Management client handle
+ */
+void Ingest::populateAssetTrackingCache(ManagementClient *m_mgtClient)
+{
+	try {
+		std::vector<AssetTrackingTuple*>& vec = m_mgtClient->getAssetTrackingTuples();
+		for (AssetTrackingTuple* & rec : vec)
+			{
+			if (rec->m_serviceName != m_serviceName || rec->m_pluginName != m_pluginName || rec->m_eventName != "Ingest")
```
MarkRiddoch: The REST API allows filtering, so would be more efficient to filter on the API call rather than retrieve all the asserts and then filter here.

AmandeepArora: ok sure.

C/common/include/management_client.h
```diff
@@ -39,9 +39,9 @@ class ManagementClient {
 							   const std::vector<std::string>& children) const;
 		std::vector<AssetTrackingTuple*>&	getAssetTrackingTuples() const;
 		bool addAssetTrackingTuple(const std::string& service, 
-													const std::string& plugin, 
-													const std::string& asset, 
-													const std::string& event);
+												const std::string& plugin, 
+												const std::string& asset, 
+												const std::string& event);
```
pintomax: You can see the code is not well formatted: use tabs

AmandeepArora: I think tabs is an issue, tabs are displayed as 2, 4 or 8 spaces in different editors and viewers and websites like github. Should we expand tabs to fixed number of spaces, say 4?

AmandeepArora:
@pintomax : This is regarding your comment on Ingest constructor having too many arguments.  What you are probably suggesting is to retain the original 3 arguments and have 3 new set functions to set the 3 new values. These 3 set functions would be called on the Ingest class object immediately after instantiating it in SouthService::start(). Is my understanding correct?

In current code, after setting all the 6 values in Ingest constructor, I initialize the cache. So if I make the changes you proposed, where would I populate the cache initially?

AmandeepArora:
@pintomax: Populating AssetTrackingCache is internal business of Ingest class. So it doesn't sound correct to expose internal implementation details to external code. Also moving all these function calls outside of Ingest class, would need all these calls to be done externally and with certain loose constraints on their ordering.

Also please help me understand why 6 params to constructor is a problem? I am seeing conflicting opinion on this in various online discussion groups.

pintomax:
@AmandeepArora ok, let's keep the the constructor with 6 parameters.

Please move loadFilters and setFilterPipeline from south.cpp to ingest.cpp as we have management client in the Ingest class right now.

C/common/include/asset_tracking.h
```diff
@@ -0,0 +1,71 @@
...
+ * Released under the Apache 2.0 Licence
+ *
+ * Author: Amandeep Singh Arora
+ */
+#include <logger.h>
+#include <vector>
+#include <sstream>
+#include <unordered_set>
+
+class AssetTrackingTuple {
```
MarkRiddoch: Please add doxgen compatible comments to classes and methods.

C/services/south/include/ingest.h
```diff
@@ -30,7 +33,10 @@ class Ingest {
 public:
 	Ingest(StorageClient& storage,
 		unsigned long timeout,
-		unsigned int threshold);
+		unsigned int threshold,
+		const std::string& serviceName,
+		const std::string& pluginName,
+		ManagementClient *m_mgmtClient);
```
MarkRiddoch: Maybe we should pass it the South Service instead so that it can pick up all these items rather than having such a big set of arguments.

C/services/south/ingest.cpp
```diff
@@ -28,6 +28,76 @@ static void ingestThread(Ingest *ingest)
 	}
 }
 
+/**
+ * Fetch all asset tracking tuples from DB and populate local cache
+ *
+ * @param m_mgtClient	Management client handle
+ */
+void Ingest::populateAssetTrackingCache(ManagementClient *m_mgtClient)
```
MarkRiddoch: m_ variables are by convention member variables rather than arguments. Although we don't always remember to make all member variables m_* and we probably shoal be better at this.

C/services/south/ingest.cpp
```diff
@@ -28,6 +28,76 @@ static void ingestThread(Ingest *ingest)
...
+ * Add asset tracking tuple via microservice management API and in cache
+ *
+ * @param tuple		New tuple to add in DB and in cache
+ */
+void Ingest::addAssetTrackingTuple(AssetTrackingTuple& tuple)
+{
+	std::unordered_set<AssetTrackingTuple*>::const_iterator it = assetTrackerTuplesCache.find(&tuple);
+	if (it == assetTrackerTuplesCache.end())
+		{
+		m_logger->info("addAssetTrackingTuple(): Tuple not found in cache: '%s', adding now.", tuple.assetToString().c_str());\
```
MarkRiddoch: \ ?

C/services/south/include/ingest.h
AmandeepArora: Only 4th (service name) and 6th argument (ManagementClient ptr) arguments are members of SouthService class, other 4 are local variables. And those 2 also are private member variables, so their get function would need to be added and it would just reduce the number of arguments by 1.

C/common/include/asset_tracking.h
AmandeepArora: I added a comment at the top : "The AssetTrackingTuple class is used to represent an asset tracking tuple. Hash function and == operator are defined for this class and pointer to this class that would be required to create an unordered_set of this class." That covers the complete details.

C/services/south/ingest.cpp
AmandeepArora: Changed.

C/services/south/include/ingest.h
AmandeepArora: We can revisit this again in future and see if we can reduce the number of arguments.

PR #1036, opened 2018-08-22
FOGL-1664 Cleanup init.sql and remove predefined entries for services…

MarkRiddoch:
… and tasks


CMakeLists.txt
```diff
@@ -12,22 +14,24 @@ add_subdirectory(C/plugins/storage/postgres)
...
+		add_subdirectory(C/plugins/south/modbus)
+	endif()
+	add_subdirectory(C/plugins/north/omf)
+	add_subdirectory(C/plugins/north/ThingSpeak)
+	add_subdirectory(C/plugins/north/http-north)
+	add_subdirectory(C/plugins/filter/scale)
+endif (BUILD_C_PLUGINS)
 add_subdirectory(C/plugins/utils)
-add_subdirectory(C/plugins/filter/scale)
 # Add Python 2.7 filter
```
pintomax: C/plugins/filter/python27 is a C++ filter which embeds Python 2.7 intepreter

It should be moved into into the BUILD_C_PLUGINS conditional build as well

MarkRiddoch: This doesn't matter as filter plugins are not seen by the Python code.

pintomax: We ca move out add_subdirectory(C/plugins/filter/scale) then

scripts/plugins/storage/sqlite/downgrade/8.sql
```diff
@@ -0,0 +1,145 @@
+
+rth plugins
```
praveen-garg: @MarkRiddoch  @pintomax  this may result in error?

PR #1037, opened 2018-08-22
FOGL-1627 Add master statistics keys

MarkRiddoch:
3 New statistics keys are added to the statistics table. These are updated by all instances of the sending process

- Readings Sent

- Statistics Sent

- Audit Sent

The choice of which statistics is updated depends on the source of the data the sending process is sending. All readings data sent on all north bound tasks will cause "Readings Sent" to be updated. For Statistics Sent it is the statistics north from all stream and likewise for audit.

In particular the GUI should use the same statistic "Readings Sent" to display the number of readings sent to the north on the status bar at the top.

### DIFF
--- a/python/foglamp/tasks/north/sending_process.py
+++ b/python/foglamp/tasks/north/sending_process.py
@@ -329,6 +329,7 @@ class SendingProcess(FoglampProcess):
             key = self.statistics_key
             _stats = await statistics.create_statistics(self._storage_async)
             await _stats.update(key, num_sent)
+            await _stats.update(self.master_statistics_key, num_sent)
         except Exception:
             _message = _MESSAGES_LIST["e000010"]
             SendingProcess._logger.error(_message)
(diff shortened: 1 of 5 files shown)


python/foglamp/tasks/north/sending_process.py
```diff
@@ -819,6 +855,7 @@ def _retrieve_configuration(self, cat_name=None, cat_desc=None, cat_config=None,
             if is_stream_valid is False:
                 raise ValueError("Error in Stream Id for Sending Process {}".format(self._name))
             self.statistics_key = await self._get_statistics_key()
+            self.master_statistics_key = await self._get_master_statistics_key()
```
amarendra-dianomic: There is no need for _get_master_statistics_key() method. All new keys are fixed keys like "READINGS" etc, hence should come in init.sql.

python/foglamp/tasks/north/sending_process.py
```diff
@@ -732,6 +733,41 @@ def _transform_in_memory_data_readings(raw_data):
             raise e
         return statistics_key
 
+    async def _get_master_statistics_key(self):
```
amarendra-dianomic: This method is not required and should be removed.

python/foglamp/tasks/north/sending_process.py
MarkRiddoch: No, we are moving away from this mechanism, hence these new keys are not done this way.

amarendra-dianomic: Then in that case a bypass mechanism will be needed for missing keys in ping method, otherwise ping will continue to fail and will affect start of the foglamp server by "foglamp start" call also. Also health check call will continue to fail.

amarendra-dianomic:
Before merging, this branch should be rebased with develop.

PR #1039, opened 2018-08-23
fixed typo

praveen-garg:


scripts/plugins/storage/sqlite/downgrade/8.sql
```diff
@@ -1,6 +1,3 @@
-
-rth plugins
```
praveen-garg: https://github.com/foglamp/FogLAMP/pull/1036/files#diff-383083a967108306480f6f981dc3bd17R2

PR #1042, opened 2018-08-24
removed stats keys hard coded for plugins

praveen-garg:


VERSION
```diff
@@ -1,2 +1,2 @@
 foglamp_version=1.3.1
-foglamp_schema=9
+foglamp_schema=10
```
MarkRiddoch: Schema 10 is already taken, will need to bump this up

praveen-garg:
Done!

PR #1045, opened 2018-08-24
FOGL-1681 Performance improvement 

MarkRiddoch:
Improver performance of /foglamp/asset for large number of buffered assets.

This resolves an issue when running the GUI on the assets page


pintomax:
Looks good. Don't we need update/downgrade sql files?

MarkRiddoch:
Whoops forgot to add the new files

PR #1011, opened 2018-08-14
FOGL-1672 Check and update defaults for C++ plugins

MarkRiddoch:
Make sure plugin name matches default value for the plugin property and rationalise

PR #1046, opened 2018-08-24
FOGL-1718: change "omf" plugin to be named "pi_server"

sstefano69:
develop done
tested.

ready.

NOTE :  
downgrade/upgrade are not managed for this task.


python/foglamp/plugins/north/pi_server/pi_server.py
```diff
@@ -82,12 +82,12 @@
     "e000000": "general error.",
 }
 # Configuration related to the OMF North
-_CONFIG_CATEGORY_DESCRIPTION = 'OMF North Plugin'
+_CONFIG_CATEGORY_DESCRIPTION = 'PI-Server North Plugin'
 _CONFIG_DEFAULT_OMF = {
     'plugin': {
-        'description': 'OMF North Plugin',
+        'description': 'Server North Plugin',
```
praveen-garg: `PI Server Plugin`

python/foglamp/plugins/north/pi_server/pi_server.py
```diff
@@ -82,22 +82,25 @@
     "e000000": "general error.",
 }
 # Configuration related to the OMF North
-_CONFIG_CATEGORY_DESCRIPTION = 'OMF North Plugin'
+_CONFIG_CATEGORY_DESCRIPTION = 'PI-Server North Plugin'
```
praveen-garg: remove hyphen `PI-Server` > `PI Server`

CMakeLists.txt
```diff
@@ -26,7 +26,7 @@ if (BUILD_C_PLUGINS)
 	if (Modbus)
 		add_subdirectory(C/plugins/south/modbus)
 	endif()
-	add_subdirectory(C/plugins/north/omf)
+	add_subdirectory(C/plugins/north/PI_Server)
```
praveen-garg: I'm not sure about the convention followed here. I would sugegst to name it `PiServer` or `pi-server`

sstefano69: this was asked/defined with Mark.

Singhal-Vaibhav:
@sstefano69 Make sure system tests are considered before we merge this.

CMakeLists.txt
praveen-garg: Then its fine.

sstefano69:
@Singhal-Vaibhav , yes I updated/tested the 3 unit tests, they are now using 
```
curl -X POST http://localhost:8081/foglamp/scheduled/task...
```

praveen-garg:
We need to find a way to delete exising omf in python/foglamp/plugins/north/ when run on top of 1.3.1. or existing develop

ashish-jabble:
@sstefano69 Default value for this plugin should be same as defined in `C/plugins/north/PI_Server/CMakeLists.txt` and for this plugin ```\"default\": \"PI_Server\"```

FYI, As @MarkRiddoch has already opened a https://github.com/foglamp/FogLAMP/pull/1011/files to update defaults values for C++ plugins. Also I believe this fix should go into the same PR

PR #1051, opened 2018-08-27
Fogl 1688

ashish-jabble:
Fixed for below endpoints
```
| GET            | /foglamp/category/{category_name}                             |
| GET  PUT       | /foglamp/category/{category_name}/{config_item}               |
| DELETE         | /foglamp/category/{category_name}/{config_item}/value         |
```

### DIFF
--- a/python/foglamp/common/configuration_manager.py
+++ b/python/foglamp/common/configuration_manager.py
@@ -10,6 +10,7 @@ import copy
 import json
 import inspect
 import ipaddress
+import datetime
 
 from foglamp.common.storage_client.payload_builder import PayloadBuilder
 from foglamp.common.storage_client.storage_client import StorageClientAsync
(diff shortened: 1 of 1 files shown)


python/foglamp/common/configuration_manager.py
```diff
@@ -30,6 +31,56 @@
...
+            self.cache[category_name].update({'date_accessed': datetime.datetime.now(), 'hit': current_hit + 1})
+            return True
+        return False
+
+    def update(self, category_name, category):
+        """Update the cache dictionary and remove the oldest item"""
+        if category_name not in self.cache and len(self.cache) >= self.max_cache_size:
+            self.remove_oldest()
+
+        self.cache[category_name] = {'date_accessed': datetime.datetime.now(), 'value': category, 'hit': self.hit, 'miss': self.miss}
```
praveen-garg: not self.hit self.miss but corresponding increments to category_name only, as you did in  `__contains__` def. for class level overall stats you can use these with self, but please increment by 1  these when these are hit or miss

python/foglamp/common/configuration_manager.py
```diff
@@ -30,6 +31,56 @@
...
+        """
+        cache: value dictionary as per category_name
+        max_cache_size: Hold the 10 recently requested categories in the cache
+        hit: number of times an item is read from the cache
+        miss: number of times an item was not found in the cache and a read of the storage layer was required
+        """
+        self.cache = {}
+        self.max_cache_size = self.MAX_CACHE_SIZE
+        self.hit = 0
+        self.miss = 0
```
praveen-garg: remove hit and miss from self.

python/foglamp/common/configuration_manager.py
```diff
@@ -30,6 +31,56 @@
...
+
+    def __contains__(self, category_name):
+        """Returns True or False depending on whether or not the key is in the cache
+        and update the hit and data_accessed value only If True"""
+        if category_name in self.cache:
+            current_hit = self.cache[category_name]['hit']
+            self.cache[category_name].update({'date_accessed': datetime.datetime.now(), 'hit': current_hit + 1})
+            return True
+        return False
```
praveen-garg: update miss only if not in, handle `self.cache[category_name]['hit' | 'miss']` with try except keyError and assign current_hit|miss = 0 before update

python/foglamp/common/configuration_manager.py
praveen-garg: use update and update only date_accessed and value.

> hit and miss will be handled by ` __contains__`

ashish-jabble:
@MarkRiddoch 
> What is the issue with add_configuration_item?

This end-point is not used anymore at GUI side. During configuration revamp page work, there is no way add config item option. So, please let me know should I proceed to delete this out completely?

> Do we have any data yet as to the cache hits/misses that might be?

yeah.
With GUI is connected with foglamp instance and PING interval set to every 5 seconds, so  hit value increases +1 for rest_api category (i.e from configuration cache).

As @Singhal-Vaibhav has also ensured that no more storage calls, when category is cached.

PR #1056, opened 2018-08-29
FOGL-1755 rest api endpoint added to get the summary of all sensors values for an asset code

ashish-jabble:
```
curl -sX GET http://localhost:8081/foglamp/asset/fogbench_humidity/summary | jq
[
  {
    "humidity": {
      "max": 83,
      "average": 33.5,
      "min": 13
    }
  },
  {
    "temperature": {
      "max": 48,
      "average": 31,
      "min": 7
    }
  }
]

```

NOTE: Also fixed limit and datetimeunits are mutually exclusive for browser endpoints


python/foglamp/services/core/api/browser.py
```diff
@@ -212,6 +213,43 @@ def prepare_limit_skip_payload(request, _dict):
...
+            curl -sX GET http://localhost:8081/foglamp/asset/fogbench_humidity/summary
+    """
+    try:
+        # Get readings from asset_code
+        asset_code = request.match_info.get('asset_code', '')
+        payload = PayloadBuilder().SELECT("reading").WHERE(["asset_code", "=", asset_code]).payload()
+        _readings = connect.get_readings_async()
+        results = await _readings.query(payload)
+        # Find keys in readings
+        reading_keys = list(results['rows'][0]['reading'].keys())
```
praveen-garg: what if no reading in YET

python/foglamp/services/core/api/browser.py
```diff
@@ -212,6 +213,43 @@ def prepare_limit_skip_payload(request, _dict):
...
+                .ALIAS('aggregate', ('reading', 'min', 'min'),
+                       ('reading', 'max', 'max'),
+                       ('reading', 'avg', 'average')).chain_payload()
+            _where = PayloadBuilder(_aggregate).WHERE(["asset_code", "=", asset_code]).chain_payload()
+            _and_where = where_clause(request, _where)
+            payload = PayloadBuilder(_and_where).payload()
+            results = await _readings.query(payload)
+            response.append({reading: results['rows'][0]})
+    except KeyError:
+        raise web.HTTPBadRequest(reason=results['message'])
```
praveen-garg: you mean Not Found?

python/foglamp/services/core/api/browser.py
```diff
@@ -212,6 +213,43 @@ def prepare_limit_skip_payload(request, _dict):
...
+    except Exception as ex:
```
praveen-garg: this is not needed, it should be auto handled as HTTPException

python/foglamp/services/core/api/browser.py
```diff
@@ -212,6 +213,43 @@ def prepare_limit_skip_payload(request, _dict):
...
+    try:
+        # Get readings from asset_code
+        asset_code = request.match_info.get('asset_code', '')
+        payload = PayloadBuilder().SELECT("reading").WHERE(["asset_code", "=", asset_code]).payload()
+        _readings = connect.get_readings_async()
+        results = await _readings.query(payload)
+        # Find keys in readings
+        reading_keys = list(results['rows'][0]['reading'].keys())
+        response = []
+        for reading in reading_keys:
```
praveen-garg: could you please file a ticket for storage support need, so that we can avoid this loop, it is all from `readings` table only.

ashish-jabble: https://scaledb.atlassian.net/browse/FOGL-1768

PR #1059, opened 2018-08-31
Fogl 1771: properly change the values in the C++/python plugins to 'PI Server'

sstefano69:


sstefano69:
something wrong happened, cancelled

PR #1064, opened 2018-09-03
FOGL-1785 Improve dashboard performance

MarkRiddoch:


scripts/plugins/storage/postgres/init.sql
```diff
@@ -490,6 +490,8 @@ CREATE TABLE foglamp.statistics_history (
 CREATE INDEX statistics_history_ix2
     ON foglamp.statistics_history(key);
 
+CREATE INDEX statistics_history_ix3
+    ON statistics_history (history_ts);
```
AmandeepArora: Will that be "foglamp.statistics_history" or it is fine even without prefix "foglamp." ?

PR #1067, opened 2018-09-04
FOGL-1711/1712 - Improvements in add_task() and add_servce() methods

amarendra-dianomic:
Test result:
`=================================================== 1293 passed, 39 skipped in 99.95 seconds ====================================================
`


python/foglamp/services/core/api/service.py
```diff
@@ -71,7 +71,7 @@ def get_service_records():
     Create a new service to run a specific plugin
 
     :Example:
-             curl -X POST http://localhost:8081/foglamp/service -d '{"name": "DHT 11", "type": "south", "plugin": "dht11", "enabled": true}'
+             curl -X POST http://localhost:8081/foglamp/service -d '{"name": "DHT 11", "plugin": "dht11", "type": "south", "process_name": "south", "enabled": true}'
```
MarkRiddoch: Why have we added process_name? This seems to be the opposite of what the Jira said, "only one process is needed for Python and one for C++. This can be shared between all south service".

amarendra-dianomic: @MarkRiddoch if we do not want to pass process_name, then in this case, we will need to hard code the process names in code, one each for Python and C++, OR, we will have to deduct process names from plugin type which will be "north" or "south" for Python and "north_c" or "south_c" for C++.

Please advise.

MarkRiddoch: This is what we already do, we deduce the name from the plugin type. That was part of the original requirement for these calls and is what we want. I do not think there is any work to be done here, so we can simply close the Jira at this stage.

amarendra-dianomic: @MarkRiddoch actually we missed this original requirement of keeping process_name same as plugin type and choose schedule_name instead. So we still need to fix this. Shall I do here and remove process_name param completely?

amarendra-dianomic: We picked schedule_name as process for both add_service() and add_task() methods since beginning. For add_service(), please see https://github.com/foglamp/FogLAMP/commit/14e714a7ebf3a8b7abd7d8eefc6d16e415450e95. And for add_task(), please see https://github.com/foglamp/FogLAMP/commit/68278d67f5a88aedc65f5caf5e2f0d17ba72abad.

MarkRiddoch: Yes, the schedule name is fine been the same, they have to be unique to the service anyway so that the service can be disabled. The scheduled process can be the same for all south service in Python and likewise for C, so I think we stick with that and drop the process_name in the payload.

MarkRiddoch: We should drop process_name out of the example as we don't use it any more

amarendra-dianomic:
@MarkRiddoch documentation example fixed. Please check.

PR #1072, opened 2018-09-05
FOGL-1601 - Refactor the test for ping method of Admin API

amarendra-dianomic:
Test result:
`=================================================== 1289 passed, 39 skipped in 97.96 seconds ====================================================
`

### DIFF
--- a/tests/unit/python/foglamp/services/core/api/test_common_ping.py
+++ b/tests/unit/python/foglamp/services/core/api/test_common_ping.py
@@ -15,6 +15,8 @@ This test file assumes those 2 units are tested
 import asyncio
 import json
 import ssl
+import socket
+import subprocess
 import pathlib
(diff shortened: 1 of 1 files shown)


tests/unit/python/foglamp/services/core/api/test_common_ping.py
```diff
@@ -44,9 +46,18 @@ def ssl_ctx(certs_path):
     return ssl_ctx
 
 
+@pytest.fixture
+def get_machine_detail():
+    host_name = socket.gethostname()
+    # all addresses for the host
+    all_ip_addresses_cmd_res = subprocess.run(['hostname', '-I'], stdout=subprocess.PIPE)
```
praveen-garg: It could have been mocked, but its okay as is!

PR #1073, opened 2018-09-05
FOGL-1798 - Bad request when config item value is non-string value irrespective of the type

ashish-jabble:
```
curl -v -sX PUT http://localhost:8081/foglamp/category/North%20Readings%20to%20PI/blockSize -d '{"value":13}' | grep "HTTP/1.1"

< HTTP/1.1 400 13 should be passed in double quotes
```

### DIFF
--- a/python/foglamp/services/core/api/configuration.py
+++ b/python/foglamp/services/core/api/configuration.py
@@ -202,7 +202,7 @@ async def set_configuration_item(request):
     try:
         value = data['value']
         if not isinstance(value, str):
-            raise web.HTTPBadRequest(reason='{} should be passed in double quotes'.format(value))
+            raise web.HTTPBadRequest(reason='{} should be a string literal, in double quotes'.format(value))
     except KeyError:
         raise web.HTTPBadRequest(reason='Missing required value for {}'.format(config_item))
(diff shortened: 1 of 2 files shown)


python/foglamp/services/core/api/configuration.py
```diff
@@ -201,6 +201,8 @@
 
     try:
         value = data['value']
+        if not isinstance(value, str):
+            raise web.HTTPBadRequest(reason='{} should be passed in double quotes'.format(value))
```
praveen-garg: {} should be a string literal in double quotes.

python/foglamp/services/core/api/configuration.py
```diff
@@ -201,6 +201,8 @@
 
     try:
         value = data['value']
+        if not isinstance(value, str):
```
praveen-garg: or may be  just `value = str(value)`

MarkRiddoch: I would tend to agree, turn it into a string if it is not. so { "value" : 10 } and { "value" : "10" } with work. The most important thing is that we raise an error if the value is not stored.

PR #1077, opened 2018-09-05
FOGL-1552 - Certain special characters break payload builder

amarendra-dianomic:
Test result:
`=================================================== 1290 passed, 39 skipped in 98.59 seconds ====================================================
`

### DIFF
--- a/python/foglamp/common/storage_client/payload_builder.py
+++ b/python/foglamp/common/storage_client/payload_builder.py
@@ -18,6 +18,7 @@ import urllib.parse
 import numbers
 
 from foglamp.common import logger
+from foglamp.common.storage_client.exceptions import *
 
 
 _LOGGER = logger.setup(__name__)
@@ -36,10 +37,12 @@ class PayloadBuilder(object):
     # TODO: Add tests
 
     query_payload = None
+    action_on_identifier_error = None
 
-    def __init__(self, initial_payload=OrderedDict()):
+    def __init__(self, initial_payload=OrderedDict(), action_on_identifier_error='warn'):
         # TODO: Investigate why simple "self.__class__.query_payload = initial_payload" is not working
         self.__class__.query_payload = initial_payload if len(initial_payload) else OrderedDict()
+        self.__class__.action_on_identifier_error = action_on_identifier_error
 
     @staticmethod
     def verify_select(arg):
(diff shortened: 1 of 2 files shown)


python/foglamp/common/storage_client/payload_builder.py
```diff
@@ -604,8 +604,45 @@ def TIMEBUCKET(cls, timestamp, size="1", fmt=None, alias=None):
...
         return cls
 
+    @staticmethod
+    def filter_identifier(identifier):
+        # Characters `(backtick),
+        #            '(single quote),
+        #            "(double quote),
+        #            [(left square bracket),
+        #            ](right square bracket)
+        # are not allowed
```
praveen-garg: this should not filter but disallow! AND see where escaping is needed.

I think we are handling only backtick and single quote. We need more like `colon`, 'semi-colon'

amarendra-dianomic:
@MarkRiddoch @praveen-garg I have added a flag **action_on_identifier_error** in payload_builder constructor, valid values are "halt" and "warn" and default to "warn". "halt" will throw StorageServerError on **any** identifier error and "warn" will throw an exception but will proceed to cleaning. We will need to manually put this flag in payload_builder calls wherever it is required. This may be done via a separate ticket. Hope this will meets the expectations.

MarkRiddoch:
I would not bother with the complexity of having two options for the behaviour. If you can store what the user asked for always, if that means escaping then we escape things. If they can not be stored then we throw an exception. I do not see the value in adding this flag and the associated complexities, you still end up with the user asks for one thing to be stored and something different is stored.

python/foglamp/common/storage_client/payload_builder.py
```diff
@@ -604,8 +607,47 @@ def TIMEBUCKET(cls, timestamp, size="1", fmt=None, alias=None):
...
 
+    @classmethod
+    def filter_identifier(cls, identifier):
+        # Characters `(backtick),
+        #            '(single quote)
+        # are not allowed
+        if not isinstance(identifier, str): return identifier
+        if '`' in identifier or "'" in identifier:
+            if cls.action_on_identifier_error == 'halt':
+                raise StorageServerError(code=404, reason="Invalid characters in payload", error=json.dumps(cls.query_payload))
```
praveen-garg: its not a `StorageServerError`, it should be PayloadBuilderError or May be called `StorageClientError`

amarendra-dianomic:
@MarkRiddoch I tried inserting data with ',",` etc in both Sqlite and Postgres manually and it was accepted where as error message was returned from Storage Layer, and correctly so, when I tried to replicate the same via GUI. Hence, IMHO, if any escaping or filtering is required, it is required to be done at Storage layer level and not at Payload Builder level. Hence I propose to close the ticket + PR. Please advise.

amarendra-dianomic:
Closed as discussed in the scrum. This ticket will be handled at Storage layer level and not at Payload builder level.

PR #1082, opened 2018-09-06
[FOGL-1818] allow json value to be dict for update (PUT)

praveen-garg:
This will be purely valid, when we solve the JSON handling  for type application/json. As of today when only accept string literal it should be acceptable as interim solution


python/foglamp/services/core/api/configuration.py
```diff
@@ -220,77 +220,6 @@
     return web.json_response(result)
 
 
-# async def add_configuration_item(request):
```
ashish-jabble: We document it as something we support, so we should probably reinstate it. Just because the GUI does not use it does not mean other client do not use it. See [JIRA](https://scaledb.atlassian.net/browse/FOGL-1790). Either remove it completely (i.e from routes.py as well) and move code snippet to mentioned JIRA

praveen-garg: restored and removed comments! per FOGL-1790

python/foglamp/services/core/api/configuration.py
```diff
@@ -196,12 +196,13 @@
...
 
     data = await request.json()
-    # TODO: make it optimized and elegant
     cf_mgr = ConfigurationManager(connect.get_storage_async())
 
     try:
         value = data['value']
-        if not isinstance(value, str):
+        if isinstance(value, dict):
+            pass
```
ashish-jabble: Bad data is still accepting for string type config items.

curl -sX PUT http://localhost:8081/foglamp/category/rest_api/authentication -d '{"value": {}}' | jq
{
  "default": "optional",
  "type": "string",
  "value": {},
  "description": "API Call Authentication (mandatory or optional)"
}


I believe for full proof interim solution for "string" type, we should add below code snippet in `foglamp.common.configuration_manager.ConfigurationManager#_validate_type_value`
```
        elif _type == 'string':
            if not isinstance(_value, str):
                return False
            return True
```

praveen-garg: for JSON we already have dict and json.loads based check

MarkRiddoch:
This does not appear to be related to a particular Jira. Can you please add something that explains the motivation for doing this? I am concerned about making a change at this state int he development cycle, when we have a critical release approaching, that has the potential to break configuration already in place.

praveen-garg:
FOGL-1818 is the ticket. JSON value update had the mechanism to allow dict or a string which can be parsed to dict. 

```
  elif _type == 'JSON':
            if isinstance(_value, dict):
                return True
            return Utils.is_json(_value)
```

WITH FOGL-1798 we are only accepting string literal  for the config item value. This fix is to allow back the JSON type values in dict or parsable dict again.
 I'm sure we have plan for fixing the  stuff the same way for all support type NOT always a  string for any type in payload.

PR #1086, opened 2018-09-07
FOGL-1797 - Handle timeout error in monitoring requests

amarendra-dianomic:


python/foglamp/services/core/service_registry/monitor.py
```diff
@@ -38,7 +38,7 @@ class Monitor(object):
     _logger = None
 
     def __init__(self):
-        self._logger = logger.setup(__name__)
+        self._logger = logger.setup(__name__, level=20)
```
MarkRiddoch: Did you mean to leave this change in?

AmandeepArora:
Tried it out and it works fine for me with latest code even when I tried 5 msec polling interval.

PR #1096, opened 2018-09-11
FOGL-1841 purge error fixes

ashish-jabble:
**Fixed items:**
1)  Total count for readings with readings storage client `query` method now (As it seems to be faster than actual call)
2) For response when streams aggregated infor for min + last object like  `{'rows': [{'min_last_object': ''}], 'count': 1}`
It should be fixed from storage layer and return integer value when no data in streams table, before it was working only due to we have prerequisite entries in init.sql for stream table
3) `config['retainUnsent']['value'].lower()` - Fixed as this condition always flag value set to `retain`
4) unit tests refactored as per new change

**NOTE** - Purge by size errors are still there as no support for any of engine. See https://scaledb.atlassian.net/browse/FOGL-1301

### DIFF
--- a/python/foglamp/tasks/purge/purge.py
+++ b/python/foglamp/tasks/purge/purge.py
@@ -22,7 +22,6 @@ Statistics reported by Purge process are:
     -> Remaining readings
     All these statistics are inserted into the log table
 """
-import asyncio
 import time
 
 from foglamp.common.audit_logger import AuditLogger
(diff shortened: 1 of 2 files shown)


python/foglamp/tasks/purge/purge.py
```diff
@@ -105,17 +104,22 @@ def __init__(self):
         start_time = time.strftime('%Y-%m-%d %H:%M:%S.%s', time.localtime(time.time()))
 
         payload = PayloadBuilder().AGGREGATE(["count", "*"]).payload()
-        result = await self._storage_async.query_tbl_with_payload("readings", payload)
+        result = await self._readings_storage_async.query(payload)
```
MarkRiddoch: This is good, the previous code was wrong as if we had a separate readings plugin it woudl not have worked.

PR #1103, opened 2018-09-12
FOGL-1848 grep -a flag added for process a binary file as if it were text

ashish-jabble:


MarkRiddoch:
@ashish-jabble Please change the destination branch for this pull request to be 1.4RC as we woudl like this in the 1.4 release.

PR #1104, opened 2018-09-12
FOGL-1844: added additional checks to delete statements

pintomax:
FOGL-1844: added additional checks to delete statements in SQLite update scripts

Upgrade 9.sql and 11.sql
Detected 'sqlite' FogLAMP DB schema change from version [2] to [14], applying Upgrade/Downgrade ...
FogLAMP DB schema has been upgraded to version [14]

Downgrade 8.sql

Detected 'sqlite' FogLAMP DB schema change from version [14] to [2], applying Upgrade/Downgrade ...
FogLAMP DB schema has been downgraded to version [2]


scripts/common/get_storage_plugin.sh
```diff
@@ -23,7 +23,9 @@ __version__="1.0"
 
 # Get the storage database plugin from the Storage microservice cache file
 get_storage_plugin() {
-
+if [ "${FOGLAMP_ROOT}" ]; then
```
MarkRiddoch: In a push to make this more bullet proof I would suggest we do the following

if [ "${FOGLAMP_ROOT}" ]; then
    $FOGLAMP_ROOT/scripts/services/storage --plugin | cut -d' ' -f1
elif [ -x scripts/services/storage ]; then
    scripts/services/storage --plugin | cut -d' ' -f1
elif [ -x /usr/local/foglamp/scripts/services/storage ]; then
    /usr/local/foglamp/scripts/services/storage --plugin | cut -d' ' -f1
else
        .... LOG ERROR and force exit ....
fi

pintomax: Added log err.
/usr/local/foglamp/ case not added.
The script is called via FOGLAMP_ROOT by foglamp scripts and without it when called by make install

scripts/package/debian/upgrade/1.4.sh
```diff
@@ -15,3 +15,11 @@ if [ -d "${remove_directory}" ]; then
...
     fi
 fi
+
+# The dummy C plugin has been renamed to random, remove the old plugin
+dummy_directory="/usr/local/foglamp/plugins/south/dummy"
+
+if [ -d $dummy_directory ]; then
+	echo "FogLAMP package update: removing 'dummy' South plugin"
+	rm -rf $dummy_directory
+fi
```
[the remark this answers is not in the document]
pintomax: Removed dir /usr/local/foglamp/plugins/north/omf

scripts/plugins/storage/sqlite/upgrade/5.sql
```diff
@@ -1,8 +1,16 @@
-UPDATE foglamp.configuration SET value = '{"plugin": {"description": "OMF North Plugin", "type": "string", "default": "omf", "value": "omf"}, "source": {"description": "Source of data to be sent on the stream. May be either readings, statistics or audit.", "type": "string", "default": "audit", "value": "readings"}}'
+UPDATE foglamp.configuration SET value = json_set(value, '$.stream_id', json('{"type":"integer","description":"Stream ID","default":"1","value":"1"}'))
+       WHERE key = 'North Readings to PI';
+UPDATE foglamp.configuration SET value = json_set(value, '$.source', json('{"plugin": {"description": "OMF North Plugin", "type": "string", "default": "omf", "value": "omf"}, "source": {"description": "Source of data to be sent on the stream. May be either readings, statistics or audit.", "type": "string", "default": "audit", "value": "readings"}}'))
```
MarkRiddoch: I think this should use be

UPDATE foglamp.configuration SET value = json_set(value, '$.source', json('{"source": {"description": "Source of data to be sent on the stream. May be either readings, statistics or audit.", "type": "string", "default": "audit", "value": "readings"}}'))

The plugin should already be there and be correct.

scripts/plugins/storage/sqlite/upgrade/5.sql
```diff
@@ -1,8 +1,16 @@
...
         WHERE key = 'North Readings to PI';
-UPDATE foglamp.configuration SET value = '{"plugin": {"description": "OMF North Plugin", "type": "string", "default": "omf", "value": "omf"}, "source": {"description": "Source of data to be sent on the stream. May be either readings, statistics or audit.", "type": "string", "default": "audit", "value": "statistics"}}'
+
+UPDATE foglamp.configuration SET value = json_set(value, '$.stream_id', json('{"type":"integer","description":"Stream ID","default":"2","value":"2"}'))
+        WHERE key = 'North Statistics to PI';
```
MarkRiddoch: As above, I don't think you need the plugin part.

scripts/plugins/storage/sqlite/upgrade/5.sql
```diff
@@ -1,8 +1,16 @@
...
         WHERE key = 'North Statistics to PI';
-UPDATE foglamp.configuration SET value = '{"plugin": {"description": "OCS North Plugin", "type": "string", "default": "ocs", "value": "ocs"}, "source": {"description": "Source of data to be sent on the stream. May be either readings, statistics or audit.", "type": "string", "default": "audit", "value": "readings"}}'
+
+UPDATE foglamp.configuration SET value = json_set(value, '$.stream_id', json('{"type":"integer","description":"Stream ID","default":"4","value":"4"}'))
+        WHERE key = 'North Readings to OCS';
+UPDATE foglamp.configuration SET value = json_set(value, '$.source', json('{"plugin": {"description": "OCS North Plugin", "type": "string", "default": "ocs", "value": "ocs"}, "source": {"description": "Source of data to be sent on the stream. May be either readings, statistics or audit.", "type": "string", "default": "audit", "value": "readings"}}'))
```
MarkRiddoch: And once again I think plugin is already correct

scripts/plugins/storage/sqlite/upgrade/5.sql
pintomax: Yes, oversight. Fixed

Singhal-Vaibhav:
This requires reviews from @amarendra-dianomic and @sstefano69 as the upgrade scripts changes in sending process configuration.

scripts/plugins/storage/postgres/upgrade/5.sql
```diff
@@ -1,17 +1,25 @@
-UPDATE foglamp.configuration SET value = '{"plugin": {"description": "OMF North Plugin", "type": "string", "default": "omf", "value": "omf"}, "source": {"description": "Source of data to be sent on the stream. May be either readings, statistics or audit.", "type": "string", "default": "audit", "value": "readings"}}'
+UPDATE foglamp.configuration SET value = jsonb_set(value, '{stream_id}', '{"type":"integer","description":"Stream ID","default":"1","value":"1"}')
+       WHERE key = 'North Readings to PI';
+UPDATE foglamp.configuration SET value = jsonb_set(value, '{source}', '{"source": {"description": "Source of data to be sent on the stream. May be either readings, statistics or audit.", "type": "string", "default": "audit", "value": "readings"}}')
```
Singhal-Vaibhav: > "default": "audit", "value": "readings"

Using this if user deletes the source config, it will fall back to audit, which is wrong

scripts/plugins/storage/postgres/upgrade/5.sql
```diff
@@ -1,17 +1,25 @@
...
         WHERE key = 'North Readings to PI';
-UPDATE foglamp.configuration SET value = '{"plugin": {"description": "OMF North Plugin", "type": "string", "default": "omf", "value": "omf"}, "source": {"description": "Source of data to be sent on the stream. May be either readings, statistics or audit.", "type": "string", "default": "audit", "value": "statistics"}}'
+
+UPDATE foglamp.configuration SET value = jsonb_set(value, '{stream_id}', '{"type":"integer","description":"Stream ID","default":"2","value":"2"}')
+        WHERE key = 'North Statistics to PI';
+UPDATE foglamp.configuration SET value = jsonb_set(value, '{source}', '{"source": {"description": "Source of data to be sent on the stream. May be either readings, statistics or audit.", "type": "string", "default": "audit", "value": "statistics"}}')
```
Singhal-Vaibhav: same as above

scripts/plugins/storage/postgres/upgrade/5.sql
```diff
@@ -1,17 +1,25 @@
...
         WHERE key = 'North Statistics to PI';
-UPDATE foglamp.configuration SET value = '{"plugin": {"description": "OCS North Plugin", "type": "string", "default": "ocs", "value": "ocs"}, "source": {"description": "Source of data to be sent on the stream. May be either readings, statistics or audit.", "type": "string", "default": "audit", "value": "readings"}}'
+
+UPDATE foglamp.configuration SET value = jsonb_set(value, 'stream_id}', '{"type":"integer","description":"Stream ID","default":"4","value":"4"}')
+        WHERE key = 'North Readings to OCS';
+UPDATE foglamp.configuration SET value = jsonb_set(value, '{source}', '{"source": {"description": "Source of data to be sent on the stream. May be either readings, statistics or audit.", "type": "string", "default": "audit", "value": "readings"}}')
```
Singhal-Vaibhav: same as above

scripts/plugins/storage/postgres/upgrade/5.sql
pintomax: You can see the original implementation of 5.sql has "audit" as default.

Better to ask @amarendra-dianomic for details.

scripts/plugins/storage/postgres/upgrade/5.sql
pintomax: Fixed: 
"default": "readings", "value": "readings"
"default": "statistics", "value": "statistics"

Singhal-Vaibhav:
The upgrade looks better now. Tested with Sqlite and works well. Testing on Postgres is pending.

scripts/plugins/storage/postgres/upgrade/15.sql
```diff
@@ -1,2 +1,4 @@
 CREATE INDEX IF NOT EXISTS log_ix2 on foglamp.log(ts);
 CREATE INDEX IF NOT EXISTS tasks_ix1 on foglamp.tasks(process_name, start_time);
+
+UPDATE foglamp.schedules SET enabled = false WHERE schedule_name IN (SELECT key FROM foglamp.configuration WHERE value->'plugin'->>'value' = 'coap');
```
Singhal-Vaibhav: Is there a reason why we are explicitly disabling the coap schedule?

scripts/plugins/storage/sqlite/upgrade/15.sql
```diff
@@ -1,2 +1,4 @@
 CREATE INDEX IF NOT EXISTS log_ix2 ON log(ts);
 CREATE INDEX IF NOT EXISTS tasks_ix1 ON tasks(process_name, start_time);
+
+UPDATE foglamp.schedules SET enabled = 'f' WHERE schedule_name IN (SELECT key FROM foglamp.configuration WHERE json_extract(value, '$.plugin.value') = 'coap');
```
Singhal-Vaibhav: Again, why are we disabling the schedule for COAP?

pintomax: A configured Coap south plugin might not be availabe in the new 1.4: better to disable the schedule.

PR #1105, opened 2018-09-12
Fogl 1842: SP/C++ and north plugins: apply 'order' and 'readonly' properties applied to the python version

sstefano69:
**preview**


C/tasks/north/sending_process/sending.cpp
```diff
@@ -44,24 +44,24 @@ static map<string, string> globalConfiguration = {};
 static const string sendingDefaultConfig =
 	"\"enable\": {"
 		"\"description\": \"A switch that can be used to enable or disable execution of "
-		"the sending process.\", \"type\": \"boolean\", \"default\": \"True\" , \"readonly\": \"True\"  },"
-	"\"duration\": {"
-		"\"description\": \"How long the sending process should run (in seconds) before stopping.\", "
-		"\"type\": \"integer\", \"default\": \"60\" , \"order\": \"1\" }, "
+		"the sending process.\", \"type\": \"boolean\", \"default\": \"True\" , \"readonly\": \"true\"  },"
```
pintomax: \"readonly\": \"true\" should be \"readonly\": \"True\" for consistency

PR #1106, opened 2018-09-13
FOGL-1673 - Updated plugin should delete obsolete config items

amarendra-dianomic:


python/foglamp/common/configuration_manager.py
```diff
@@ -265,13 +278,22 @@ def get_entry_val(k):
 
     async def _create_new_category(self, category_name, category_val, category_description):
         try:
+            if isinstance(category_val, dict):
+                new_category_val = copy.deepcopy(category_val)
+                # Remove "deprecated" items from a new category configuration
+                for i, v in category_val.items():
+                    if 'deprecated' in v:
+                        new_category_val.pop(i)
```
praveen-garg: I think we need an Audit log here for the removal of deprecated

python/foglamp/common/configuration_manager.py
```diff
@@ -265,13 +278,22 @@ def get_entry_val(k):
...
             audit = AuditLogger(self._storage)
-            await audit.information('CONAD', {'name': category_name, 'category': category_val})
+            await audit.information('CONAD', {'name': category_name, 'category': new_category_val})
             payload = PayloadBuilder().INSERT(key=category_name, description=category_description,
-                                              value=category_val).payload()
+                                              value=new_category_val).payload()
             result = await self._storage.insert_into_tbl("configuration", payload)
             response = result['response']
-            self._cacheManager.update(category_name, category_val)
+            self._cacheManager.update(category_name, new_category_val)
```
praveen-garg: We should save the value as saved and returned in the DB.

python/foglamp/common/configuration_manager.py
amarendra-dianomic: First of all, this is _create_new_category() method. A check for "deprecated" attribute is required here to prevent the deprecated items from going into configuration right at the very first time. Secondly, removing "deprecated" items in _merge_category_vals() is identical to removing an  item per the regular merge process logic. Since _merge_category_vals() method never adds an Audit log entry for its operations, we also choose not to do so for removing the deprecated items.

python/foglamp/common/configuration_manager.py
amarendra-dianomic: Insert and update operations return just a string - "inserted" or "updated" as response. Hence "response" is not suitable to be added to the cache.

amarendra-dianomic:
@MarkRiddoch @praveen-garg tests added for new deprecated module. Please check.

python/foglamp/common/configuration_manager.py
praveen-garg: Not the response, but we MAY want to re-read the value from DB, immediately after insertion// update; to update cache.

amarendra-dianomic: Now cache updated from db value in _update_category() and not in _create_new_category() where it is not needed.

python/foglamp/common/configuration_manager.py
```diff
@@ -228,6 +236,9 @@ def get_entry_val(k):
                     if entry_name == 'readonly':
```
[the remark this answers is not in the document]
amarendra-dianomic: 'deprecated' taken care next

amarendra-dianomic:
@ashish-jabble correct syntax is 'deprecated': 'true'. 'deprecated': true i.e. true without quotes is not allowed.

ashish-jabble:
> @ashish-jabble correct syntax is 'deprecated': 'true'. 'deprecated': true i.e. true without quotes is not allowed.

yeah above one a typo I used `'deprecated': 'true'` and I am testing with -ve values like false/blah which results in error.

python/foglamp/common/configuration_manager.py
```diff
@@ -225,9 +233,12 @@ def get_entry_val(k):
 
                 # If Entry item exists in optional list, then update expected item entries
                 if entry_name in optional_item_entries:
-                    if entry_name == 'readonly':
+                    if entry_name == 'readonly' or entry_name == 'deprecated':
                         if self._validate_type_value('boolean', entry_val) is False:
                             raise ValueError('Unrecognized value for item_name {}'.format(entry_name))
+                        if entry_name == 'deprecated':
```
praveen-garg: why do we need this?

python/foglamp/common/configuration_manager.py
```diff
@@ -265,13 +278,22 @@ def get_entry_val(k):
 
     async def _create_new_category(self, category_name, category_val, category_description):
         try:
+            if isinstance(category_val, dict):
+                new_category_val = copy.deepcopy(category_val)
+                # Remove "deprecated" items from a new category configuration
+                for i, v in category_val.items():
+                    if 'deprecated' in v:
```
praveen-garg: if in v and true?

python/foglamp/common/configuration_manager.py
```diff
@@ -167,14 +167,22 @@ def __init__(self, storage=None):
         # keep_original_items = TRUE keep items in category_val_storage not in category_val_new
         category_val_storage_copy = copy.deepcopy(category_val_storage)
         category_val_new_copy = copy.deepcopy(category_val_new)
+        deprecated_items = []
         for item_name_new, item_val_new in category_val_new_copy.items():
             item_val_storage = category_val_storage_copy.get(item_name_new)
             if item_val_storage is not None:
                 item_val_new['value'] = item_val_storage.get('value')
                 category_val_storage_copy.pop(item_name_new)
+            if "deprecated" in item_val_new:
```
praveen-garg: and true?

python/foglamp/common/configuration_manager.py
```diff
@@ -225,9 +233,12 @@ def get_entry_val(k):
 
                 # If Entry item exists in optional list, then update expected item entries
                 if entry_name in optional_item_entries:
-                    if entry_name == 'readonly':
+                    if entry_name == 'readonly' or entry_name == 'deprecated':
                         if self._validate_type_value('boolean', entry_val) is False:
                             raise ValueError('Unrecognized value for item_name {}'.format(entry_name))
+                        if entry_name == 'deprecated':
+                            if entry_val.lower() != 'true':
+                                raise ValueError('Unrecognized value for item_name {}'.format(entry_name))
```
praveen-garg: even if we RESTRICT  (instead of moving forward  if boolean false) then give clear message? `deprecated item' value can only be true`

praveen-garg:
Yes, when its false ( and does not make sense) simply discard it and take no action.

python/foglamp/common/configuration_manager.py
amarendra-dianomic: bcoz we are restricting deprecated to true only.

python/foglamp/common/configuration_manager.py
amarendra-dianomic: it will simply be discarded

python/foglamp/common/configuration_manager.py
amarendra-dianomic: before coming to this part, validation checks have already ensured that only 'true' values are available for 'deprecated'.

amarendra-dianomic:
@ashish-jabble @praveen-garg feddback changes done. Pl check.

python/foglamp/common/configuration_manager.py
amarendra-dianomic: done

python/foglamp/common/configuration_manager.py
```diff
@@ -225,9 +233,12 @@ def get_entry_val(k):
 
                 # If Entry item exists in optional list, then update expected item entries
                 if entry_name in optional_item_entries:
-                    if entry_name == 'readonly':
+                    if entry_name == 'readonly' or entry_name == 'deprecated':
                         if self._validate_type_value('boolean', entry_val) is False:
                             raise ValueError('Unrecognized value for item_name {}'.format(entry_name))
+                        if entry_name == 'deprecated':
+                            if entry_val.lower() != 'true':
+                                raise ValueError("[deprecated] attribute can be set to [true] only")
```
ashish-jabble: As per JIRA `attribute is called “deprecated” and may be set to true or false`

python/foglamp/common/configuration_manager.py
ashish-jabble:
```
if entry_name == 'readonly' or entry_name == 'deprecated':
    if self._validate_type_value('boolean', entry_val) is False:
        raise ValueError('{} attribute can be set to [true/false] only'.format(entry_name))
```

amarendra-dianomic:
@ashish-jabble @praveen-garg Pl check.

PR #1108, opened 2018-09-14
FOGL-1745 - Handle aio cancelled error for async/poll plugins

amarendra-dianomic:


python/foglamp/services/south/server.py
```diff
@@ -232,7 +235,7 @@ def run(self):
             for task_pending in pending:
                 task_pending.cancel()
             await asyncio.sleep(2)
-        except (asyncio.CancelledError, exceptions.DataRetrievalError):
+        except asyncio.CancelledError:
```
praveen-garg: move this after `task_pending.cancel()`  inside for loop 

```
for task in pending:
        task.cancel()
        # Now, await task to execute it's cancellation.
        # Cancelled task raises asyncio.CancelledError that we can suppress:
        with suppress(asyncio.CancelledError):
            await task_pending
```


 make  await asyncio.sleep(2) to bit low. may be 1. or 0.5 and inside for loop

amarendra-dianomic: Using "with suppress(asyncio.CancelledError)" needs additional dependency contextlib.suppress. Also it provides absolutely no difference in functionality, so, IMHO, we are better off with current version only.

praveen-garg: let's remove suppress!

amarendra-dianomic: @praveen-garg note that asyncio.CancelledError is raised after the task is cancelled. Hence the correct code is as below:
```
for task_pending in pending:
        try:
             task_pending.cancel()
             # Now, await task  to execute it's cancellation.
             # Cancelled task raises asyncio.CancelledError:
        except asyncio.CancelledError:
             pass
```

python/foglamp/services/south/server.py
```diff
@@ -230,9 +233,14 @@ def run(self):
...
             done, pending = await asyncio.wait(asyncio.Task.all_tasks(), timeout=_CLEAR_PENDING_TASKS_TIMEOUT)
             for task_pending in pending:
-                task_pending.cancel()
+                try:
+                     task_pending.cancel()
+                     # Now, await task  to execute it's cancellation.
+                     # Cancelled task raises asyncio.CancelledError:
+                except asyncio.CancelledError:
+                    pass
             await asyncio.sleep(2)
```
praveen-garg: I think this is too much should we try with 0.5 / 1.0 ?

python/foglamp/services/south/server.py
```diff
@@ -230,9 +233,14 @@ def run(self):
             # Cancel all pending asyncio tasks after a timeout occurs
             done, pending = await asyncio.wait(asyncio.Task.all_tasks(), timeout=_CLEAR_PENDING_TASKS_TIMEOUT)
             for task_pending in pending:
-                task_pending.cancel()
+                try:
+                     task_pending.cancel()
+                     # Now, await task  to execute it's cancellation.
+                     # Cancelled task raises asyncio.CancelledError:
+                except asyncio.CancelledError:
+                    pass
```
praveen-garg: cancelled error can be genuine, should we await task? or we don;t care as we are going to stop it anyway, as we have already provided _CLEAR_PENDING_TASKS_TIMEOUT.

amarendra-dianomic: If you want to handle CancelledError to do some local processing or logging or whatever, you need to catch this exception within future/task itself and add the handler code in the except block locally. south/server.py->stop() is not the right place for it. Secondly, cancellation of all user defined tasks/futures have already been handled prior to this code. This code block is just a kind of safety check to ensure that cancellation of remaining pending tasks, if any at all, are properly taken care of.

PR #1120, opened 2018-09-24
[FOGL-1877] Fix typo in error response for Add Service  /Task call. Log (internal server) errors to syslog

praveen-garg:


praveen-garg:
@ashish-jabble  Thanks for feedback. logger patched.  also fixed 2 badly written tests.

PR #1124, opened 2018-09-25
FOGL-1903: Changes for supporting OMF data compression

AmandeepArora:
Tested with compression ON and OFF.


AmandeepArora:
@MarkRiddoch - Calling side changes are in C++ piserver plugin. There is a separate PR for that: https://github.com/foglamp/foglamp-north-piserver/pull/10/files. Please review that also.

PR #1126, opened 2018-09-25
FOGL-1863: Linked in libcommon-lib.so into get_plugin_info utility

AmandeepArora:
Worked for me correctly in my build dir as well as via:

curl -X GET http://localhost:8081/foglamp/plugins/installed?type=north
curl -X GET http://localhost:8081/foglamp/plugins/installed?type=south

@ashish-jabble : Please check in your environment also.


AmandeepArora:
When "get_plugin_info" is invoked on shell, then LD_LIBRARY_PATH is required to be set before that. 

But when plugin discovery is invoked thru' REST API or thru' GUI, then it works fine because LD_LIBRARY_PATH is set in scripts/foglamp and that is inherited by core process and is thus able to service the plugin discovery REST API.

Hence closing this PR.

PR #1139, opened 2018-09-27
FOGL-1862: Setting make dependencies correctly for parallel make

AmandeepArora:
Was able to create parallel build with "make -j 16" successfully 30-40 times consecutively with no failure. Having as many jobs as number of processors is probably a good idea - make -j`nproc`

Tested the build made using parallel make in home directory as well as in "/usr/local/foglamp" and with one south and one north plugin.


AmandeepArora:
@Singhal-Vaibhav - I'll look into that cmake warning on a RPi on wednesday.

AmandeepArora:
Created FOGL-1926 for cmake warning seen during parallel make seen on RPi.

PR #1143, opened 2018-09-27
FOGL-1871 add_service to allow config items to set

ashish-jabble:


praveen-garg:
Plugin file/ module is imported and being **CACHED** at run time in sys.modules, when add service / add task is being called. there we shall need to use copy.deepcopy() for plugin_info symbol. Or may be we need to use importlib.reload https://docs.python.org/3.5/library/importlib.html#importlib.reload

python/foglamp/services/core/api/service.py
```diff
@@ -171,14 +173,21 @@ def get_service_records():
...
                                              keep_original_items=True)
             # Create the parent category for all South services
             await config_mgr.create_category("South", {}, "South microservices", True)
             await config_mgr.create_child_category("South", [name])
+
+            # If config is in POST data, then update the value for each config item
+            if config is not None:
+                if not isinstance(config, dict):
+                    raise ValueError('Config must be a JSON object')
+                for k, v in config.items():
```
praveen-garg: Personally, I wanted this to be done in a single call instead of a loop. But to avoid overwriting default, this is okay!

PR #1144, opened 2018-09-28
FOGL-1843 - Discard readings when no ingest queue slot is available + other refactorings

amarendra-dianomic:


python/foglamp/services/south/ingest.py
```diff
@@ -210,6 +210,13 @@ class Ingest(object):
 
         await cls._read_config()
 
+        # Auto adjust queue size in case plugin's 'dataPointsPerSec' value exceeds default queue size
+        try:
+            cls._readings_insert_batch_size = max(cls._readings_insert_batch_size,
+                                                  int(cls._parent_service.config['dataPointsPerSec']['value']))
```
Singhal-Vaibhav: 1. dataPointsPerSec is not a standard notation across all south plugins
2. We cannot have the batch size set to dataPointsPerSec on a low end device like raspberry pi. The Host machine will starve memory in that case. I suggest to keep it configurable

python/foglamp/services/south/ingest.py
```diff
@@ -340,23 +353,15 @@ def increment_discarded_readings(cls):
             # Perform insert. Retry when fails.
             while True:
                 try:
-                    payload = dict()
-                    payload['readings'] = copy.deepcopy(readings_list)
-                    batch_size = len(payload['readings'])
+                    batch_size = len(readings_list)
```
Singhal-Vaibhav: Again, We cannot have the batch size set to len(readings_list) on a low end device like raspberry pi. The Host machine will starve memory in that case. I suggest to keep in configurable

python/foglamp/services/south/ingest.py
praveen-garg: @Singhal-Vaibhav  has a valid point here. We can not hard code this.

python/foglamp/services/south/ingest.py
amarendra-dianomic: I do not understand this comment. Since beginning, the batch_size is being calculated on reading_list count as batch_size is exactly this quantity only. What makes you think it is different now?

python/foglamp/services/south/ingest.py
amarendra-dianomic: Regarding first point, [FOGL-1780](https://scaledb.atlassian.net/browse/FOGL-1780) talks about standardisation of data points per second across all South plugins and Sinusoid plugin already implements this. So, IMHO, there is no reason to choose another name for this. 
Regarding second point, if you examine the code carefully, you will find that the batch size is there since beginning and is in play regardless of whether we are running FogLAMP on ubuntu or Raspberry. Secondly, this is a try/except loop. If a plugin does not have this item, it will simply be ignored and default/configured values would be choosen. Thirdly, it gives us a chance to auto adjust queue size which greatly benefits the readings buffer size sent to DB layer and helps in minimizing DB write lock frequencies and thus in increasing the throughput. And lastly, nothing stops us in choosing the queue size manually above the dps rate because choosing lower rate would choke the Ingest process that we have experienced so many times for high dps rates. Also, please note that currently queue size is not configurable per se and can be set directly in code only.

amarendra-dianomic:
@MarkRiddoch a minor correction is required here. Per ticket and per current Ingest code, if a queue slot is not available, the Ingest process just waits there and does not discard the readings. That is why there is no discarded readings count. This in turn, slows down the ingest process because the queue is emptied only upon DB write and not before. The ticket wants this to be changed and wants all such waiting readings to be discarded. This PR attempts to counter this issue on two fronts:
a) Increase the queue size automatically so that more readings can be held in memory as well as can use larger DB write buffer and thus minimize DB lock occasions.
b) Discard the waiting readings altogether. Because for high speed devices, limited Raspberry memory simply cannot possibly buffer all incoming readings awaiting DB write because DB write, as we all know, is inherently slower.

amarendra-dianomic:
@praveen-garg regarding your comment "the scope of this ticket is too add to discard when ingest is busy. NOT handling a specific plugin and play with profile tuning.", I have explained above that this is not one plugin specific. Though dps is being used in Sinusoid plugin only but that does not stop other plugins to use the same. And since we would like to standardize upon the name for dps, like we did for pollInterval and sleepInterval, we can choose any other name, if we so desire,

python/foglamp/services/south/ingest.py
```diff
@@ -22,7 +22,7 @@
 __license__ = "Apache 2.0"
 __version__ = "${VERSION}"
 
-_LOGGER = logger.setup(__name__)  # type: logging.Logger
+_LOGGER = logger.setup(__name__, level=20)  # type: logging.Logger
```
praveen-garg: remove the log level.

python/foglamp/services/south/ingest.py
```diff
@@ -242,6 +250,8 @@ class Ingest(object):
         cls._insert_readings_task = asyncio.ensure_future(cls._insert_readings())
         cls._readings_lists_not_full = asyncio.Event()
 
+        cls._payload_events = cls._parent_service._core_microservice_management_client.get_asset_tracker_events()['track']
```
praveen-garg: avoid the extra new lines around

python/foglamp/services/south/ingest.py
```diff
@@ -291,7 +301,11 @@ def increment_discarded_readings(cls):
 
         while list_index <= cls._max_concurrent_readings_inserts-1:
             if cls._stop:
-                break  # Terminate this method
+                readings = 0
+                for i, v in cls._sensor_stats.items():
+                    readings += v
+                if cls._readings_stats+cls._discarded_readings_stats+readings == 0:
```
praveen-garg: add space around `+` operator and add a comment why this SUM should be 0 to break

python/foglamp/services/south/ingest.py
```diff
@@ -529,12 +538,10 @@ def is_available(cls) -> bool:
...
 
-        # Wait for an empty slot in the list
-        while not cls.is_available():
-            cls._readings_lists_not_full.clear()
-            await cls._readings_lists_not_full.wait()
-            if cls._stop:
-                raise RuntimeError('The South Service is stopping')
+        # If an empty slot is not available, discard the reading
+        if not cls.is_available():
+            cls.increment_discarded_readings()
```
praveen-garg: this `increment_discarded_readings` is okay, but why removed the clear and wait event. and  if stop why not the RuntimeError  `RuntimeWarning` actually

python/foglamp/services/south/ingest.py
```diff
@@ -231,6 +236,9 @@ class Ingest(object):
         cls._insert_readings_wait_tasks = []
         cls._readings_list_batch_size_reached = []
         cls._readings_list_not_empty = []
+
+
+
```
praveen-garg: why these 3 new lines?

python/foglamp/services/south/ingest.py
amarendra-dianomic: clear and wait event handling has become now redundant  as we are not waiting for the empty slot now and simply discard the reading if queue is full. Secondly, RuntimeWarning, if raised, will require to handled somewhere. This method is fired by two callers - a) South server poll task, and b) async plugin. Both these callers do not have anything useful to do if this RuntimeWarning is raised. Besides, syslog error message is always generated when all ingest queues become full. Hence not provided.

amarendra-dianomic:

praveen-garg:
@amarendra-dianomic  please resolve the conflicts

python/foglamp/services/south/ingest.py
```diff
@@ -242,13 +246,14 @@ class Ingest(object):
         cls._insert_readings_task = asyncio.ensure_future(cls._insert_readings())
         cls._readings_lists_not_full = asyncio.Event()
 
+        cls._payload_events = cls._parent_service._core_microservice_management_client.get_asset_tracker_events()['track']
```
MarkRiddoch: What is this for? I don't see any uses of _payload_events being added in this pull request and do not see how it relates to the discarding of readings.

amarendra-dianomic: @MarkRiddoch yes, cls._payload_events is not related to discarded readings at all but this is a kind of cache that we maintain locally to save network trip when we check and add an asset to asset_tracker table. We just need to create an initial cache from DB though which is the purpose of the above code and which we missed it earlier.

PR #1145, opened 2018-09-28
FOGL-1754 : updated stat handling.

sstefano69:


ashish-jabble:
@sstefano69 Sending process update stats is still not working, as we have used `toupper(c)` for every key in C-code.
readings sent key : `Readings Sent` Vs `READINGS SENT`

PR #1149, opened 2018-10-03
FOGL-1872 - South service to add advanced configuration as child category

amarendra-dianomic:
Test result:
`================================================================================== 1390 passed, 39 skipped in 107.85 seconds ==================================================================================
`

### DIFF
--- a/python/foglamp/services/south/ingest.py
+++ b/python/foglamp/services/south/ingest.py
@@ -123,7 +123,7 @@ class Ingest(object):
         """Creates default values for the South configuration category and then reads all
         values for this category
         """
-        category = 'South'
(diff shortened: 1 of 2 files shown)


python/foglamp/services/south/ingest.py
```diff
@@ -173,13 +173,16 @@ class Ingest(object):
             "key": category,
             "description": 'South Service configuration',
             "value": default_config,
-            "keep_original_items": False
+            "keep_original_items": True
```
MarkRiddoch: Having this keep_original_items in the JSON is now obsolete, it is a parameter. Why has this changed from False to True, I would have expected the same behaviour here?

PR #1150, opened 2018-10-03
FOGL-1341 Fix prompt and make mandatory/optional an enumeration

MarkRiddoch:


python/foglamp/services/core/server.py
```diff
@@ -156,8 +156,9 @@ class Server:
             'default': 'foglamp'
         },
         'authentication': {
-            'description': 'API Call Authentication (mandatory or optional)',
-            'type': 'string',
+            'description': 'API Call Authentication',
+            'type': 'enumeration',
```
praveen-garg: is this change upgrade safe?

PR #1156, opened 2018-10-04
FOGL-1874 display name  added for create cat and optional attribute for config item; upgrade/downgrade scripts added

ashish-jabble:
**Fixed items:**

- [x] Allow an optional property of a configuration item, in the JSON document, called displayName. 

- [x] Add a new database column to the configuration table called display_name. The default value of this should be a copy of the key column, write an upgrade script that will copy the key value to the display_name value

- [x] Update the API response to include the property displayName in the response for /foglamp/category and /foglamp/category/xxx.

- [x] Also add displayName to the request payload for the POST that creates categories.

- [x] Update the management client to accept displayName when creating a category.


ashish-jabble:
@MarkRiddoch What about renaming of displayName for the existing category? as there is only possibility of merge category values as per design? Below is the payload
`{"key": "rest_api", "description": "description", "display_name": "Admin API", "value": {....} }`

python/foglamp/common/configuration_manager.py
```diff
@@ -298,10 +302,11 @@ def get_entry_val(k):
...
 
+            display_name = category_name if display_name is None else display_name
             audit = AuditLogger(self._storage)
             await audit.information('CONAD', {'name': category_name, 'category': new_category_val})
             payload = PayloadBuilder().INSERT(key=category_name, description=category_description,
-                                              value=new_category_val).payload()
+                                              value=new_category_val, display_name=display_name).payload()
             result = await self._storage.insert_into_tbl("configuration", payload)
             response = result['response']
             self._cacheManager.update(category_name, new_category_val)
```
praveen-garg: I believe, We need to cache this display name too!

MarkRiddoch: Yes, we should probably keep the display name in the cache

MarkRiddoch:
I'm not sure I understand what the issue is here with adding a display name for the category itself. The original Jira ticket says...

> Also add displayName to the request payload for the POST that creates categories. Again if it is not in the payload then use the value of the key. Also update underlying API to allow the displayName parameter to be passed through to the configuration manager.
> 
> Update the management client to accept displayName when creating a category.

Which I believe covers all the issues were we want to be able to add categories with a display name. There is no PUT of /foglamp/category/{categoryName}, so we have no update interface for the category name and description and therefore should not have one for the display name either.

We will have other changes that will update the default configuration for certain components of the system, but this is outside of the scope of this change.

ashish-jabble:
> Also add displayName to the request payload for the POST that creates categories. Again if it is not in the payload then use the value of the key. Also update underlying API to allow the displayName parameter to be passed through to the configuration manager.
Update the management client to accept displayName when creating a category.

It has been already implemented in the PR itself, See [curl example](https://github.com/foglamp/FogLAMP/pull/1156/files#diff-d6e2f8f5081014e10aa81280a3474109R101) and it works only with FRESH/NEWLY category and well tested.


The Problem is with **existing category**. See the configuration manager code line [here](https://github.com/foglamp/FogLAMP/pull/1156/files#diff-28b9ac168197801fa73fdc04687409d5R688), we only MERGE category vals only.

```
 # if validating category from storage succeeds, merge new and storage
else:
   category_val_prepared = await self._merge_category_vals(category_val_prepared, category_val_storage, keep_original_items, category_name)
  if json.dumps(category_val_prepared, sort_keys=True) == json.dumps(category_val_storage, sort_keys=True):
      return 
await self._update_category(category_name, category_val_prepared, category_description)
```
To achieve above displayName updation for existing category we need to update this logic.

Should I go ahead to change it? As it needs a lot testing thats why I haven't touched the code. 

**NOTE:** Also we need to update/logic for configuration cache as well, as we only cached values like:
`{"category_name": {"value": {....}}}`. 
And *keep_original_items* flag only works with `{"value": {....}}` i.e _merge_category_vals?


**Expected: to change existing category be like as below**
```
await cls._configuration_manager.create_category(category, config, 'FogLAMP Admin and User REST API', True, "Admin API")

where, "Admin API" is a display_name
```

ashish-jabble:
@MarkRiddoch I have fixed display name reference for **rest_api** category.
Also rest of the internal category references for display name shall be modified later, in other tickets?

ashish-jabble:
JIRA has been created [FOGL-1956](https://scaledb.atlassian.net/browse/FOGL-1956) for the pending displayName references for internal categories

PR #1159, opened 2018-10-04
Fogl 1909 - Revised sleep seconds calculation

amarendra-dianomic:


amarendra-dianomic:
@MarkRiddoch I had to revert earlier PR for FOGL-1909 as CC2650 Poll plugin gave some error on shorter sleep seconds (when set from 1000 ms to 500 ms) when delta became > sleep seconds somehow. Hence I have revised sleep seconds calculation to ignore delta in such cases.

PR #1161, opened 2018-10-05
FOGL-1938: Fixed compilation warnings on RPi

AmandeepArora:


C/services/south/south.cpp
```diff
@@ -219,7 +219,7 @@ void SouthService::start(string& coreAddress, unsigned short corePort)
 		// Get and ingest data
 		if (! southPlugin->isAsync())
 		{
-			int fd = createTimerFd(m_pollInterval * 1000); // interval to be passed is in usecs
+			int fd = createTimerFd((int) m_pollInterval * 1000); // interval to be passed is in usecs
```
MarkRiddoch: You will need to update this with FOGL-1801 changes.

AmandeepArora: yeah sure.

PR #1162, opened 2018-10-05
FOGL-1801: FogLAMP Benchmarking activities

AmandeepArora:
Let me know if I should retain the omf.h and omf.cpp changes or not.

### DIFF
--- a/C/plugins/common/omf.cpp
+++ b/C/plugins/common/omf.cpp
@@ -72,13 +72,11 @@ OMF::OMF(HttpSender& sender,
 	m_lastError = false;
         m_readings = 0;
         m_usecs = 0;
-        m_loopUsecs = 0;
 }
 
 // Destructor
 OMF::~OMF()
 {
-	Logger::getLogger()->info("Total %lld readings sent to PI server in %lld usecs, amortized readings' loop time = %lld usec/reading", m_readings, m_loopUsecs, m_loopUsecs/m_readings);
 	Logger::getLogger()->info("Total %lld readings sent to PI server in %lld usecs, amortized request->response time = %lld usec/reading", m_readings, m_usecs, m_usecs/m_readings);
 }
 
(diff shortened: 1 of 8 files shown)


C/plugins/common/omf.cpp
```diff
@@ -319,8 +323,13 @@ uint32_t OMF::sendToServer(const vector<Reading *>& readings,
 	// Then get HTTPS POST ret code and return 0 to client on error
 	try
 	{
+		std::chrono::high_resolution_clock::time_point t1 = std::chrono::high_resolution_clock::now();
```
MarkRiddoch: I wonder how expensive it is to track the time?

C/services/south/south.cpp
```diff
@@ -202,8 +202,11 @@ void SouthService::start(string& coreAddress, unsigned short corePort)
 
 		try {
 			m_pollInterval = 500;
+			m_readingsPerSec = 1;
 			if (m_config.itemExists("pollInterval"))
```
MarkRiddoch: We can probably ditch the pollInterval stuff completely and just have the readingsPerSec for today

C/plugins/common/omf.cpp
AmandeepArora: I don't think much, but I can remove it if you wish. Or I can find out the overhead.

AmandeepArora: Actually, when we have sent out 5000 readings to Pi server, it takes around 700-1000 msec to get POST request serviced, so in that sense overhead of time calculation is insignificant.

C/services/south/south.cpp
AmandeepArora: Done.

C/plugins/common/omf.cpp
AmandeepArora: Tried it on my ubuntu VM, I got "1000 times duration calc took 213 usecs" and "100000 times duration calc took 17590 usecs". So it's around 175-213 nano-secs.

        std::chrono::high_resolution_clock::time_point t1 = std::chrono::high_resolution_clock::now();
	std::chrono::high_resolution_clock::time_point t2;
	int i;
	for (i=0; i<1000; i++)
		t2 = std::chrono::high_resolution_clock::now();
 	auto usecs = std::chrono::duration_cast<std::chrono::microseconds>( t2 - t1 ).count();
	logger->info("%d times duration calc took %lld usecs", i, usecs);

MarkRiddoch: OK, fine thanks.

PR #1166, opened 2018-10-08
Fogl 1861: fix statistics_history loading and fields used by SP-C++/Python 

sstefano69:
Ready, tested : 
- Vbox  : 
**SP/C++
**SP/Python
**SQLite
**PG
- Raspibian:
**Python / SQLite


C/common/reading.cpp
```diff
@@ -167,3 +167,48 @@ ostringstream assetTime;
...
+ * @return              The formatted datetime string
+ */
+const string Reading::getAssetDateUserTime(readingTimeFormat dateFormat, bool addMS) const
+{
+char date_time[DATE_TIME_BUFFER_LEN];
+char micro_s[10];
+ostringstream assetTime;
+
+        // Populate tm structure
+        const struct tm *timeinfo = std::localtime(&(m_userTimestamp.tv_sec));
```
MarkRiddoch: localtime? Doesn't this return UTC?

sstefano69:
@Mark, please take a look at the following 2 files to be sure the change are fine  : 

C/plugins/storage/postgres/connection.cpp
C/plugins/storage/sqlite/connection.cpp

C/plugins/storage/sqlite/connection.cpp
```diff
@@ -711,8 +711,6 @@ SQLBuffer	jsonConstraints;
 									sql.append("strftime('%Y-%m-%d %H:%M:%f', ");
 									sql.append((*itr)["column"].GetString());
 									sql.append(", 'utc')");
-									sql.append(" AS ");
```
MarkRiddoch: Don't we need this to get the right column names back?

sstefano69: I had problems at the first run 
adding a log, I found there was  .. AS field... twice in the created SQL,
as (at my understanding) it was already added by the code : 

{code}
if (itr->HasMember("alias")) {
    sql.append(" AS \"");
    sql.append((*itr)["alias"].GetString());
{code}

the current code works fine for the tests I have done (related to this topic) 
but I would like to be sure if it is in general fine or 
if there are considerations to do that I don't see right now because they require a broader view.

PR #1167, opened 2018-10-08
FOGL-1942 utc timestamp fixes for statistics history

ashish-jabble:


mshariq-nerd:
Looks good on GUI timestamp for asset readings graph vs dashboard readings graph matching now.

PR #1169, opened 2018-10-09
[FOGL-1943] Handle storage connection timeout/ cancelled error gracefully 

praveen-garg:
if connection is being closed by upstream (storage server here) for a request with timeout cancellation, We need to handle it gracefully and avoid the noise in syslog. 

Also a out of scope change is added in this PR: to move scheduler::info logs to debug which were slightly an overhead. 

> We MAY still want to investigate why stats query with  payload is taking 5+ seconds; and may also think to have a configurable timeout in storage client request and return `504 | Gateway Timeout server error response `.


python/foglamp/services/south/ingest.py
```diff
@@ -310,7 +313,6 @@ def increment_discarded_readings(cls):
 
             readings_list = cls._readings_lists[list_index]
             min_readings_reached = cls._readings_list_batch_size_reached[list_index]
-            list_not_empty = cls._readings_list_not_empty[list_index]
```
praveen-garg: This was an unused variable.

PR #1171, opened 2018-10-10
FOGL-1943 Removed unused config and Avoid object instantiation on each queue ingest 

praveen-garg:
- [x] Avoid object instantiation with call to load (get) stats key on each queue ingest but only when Ingest is started.

- [x] Need to fix failing tests

### DIFF
--- a/python/foglamp/services/south/ingest.py
+++ b/python/foglamp/services/south/ingest.py
@@ -52,12 +52,6 @@ class Ingest(object):
     _sensor_stats = {}  # type: dict
     """Number of sensor readings accepted before statistics were written to storage"""
 
-    _write_statistics_task = None  # type: asyncio.Task
-    """asyncio task for :meth:`_write_statistics`"""
-
-    _write_statistics_sleep_task = None  # type: asyncio.Task
-    """asyncio task for asyncio.sleep"""
-
     _stop = False
     """True when the server needs to stop"""
 
@@ -91,10 +85,6 @@ class Ingest(object):
     _readings_list_size = 0  # type: int
     """Maximum number of readings items in each buffer"""
 
-    # Configuration (begin)
-    _write_statistics_frequency_seconds = 5
-    """The number of seconds to wait before writing readings-related statistics to storage"""
-
     _readings_buffer_size = 4096
     """Maximum number of readings to buffer in memory(_max_concurrent_readings_inserts x _readings_insert_batch_size)"""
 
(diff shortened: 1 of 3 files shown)


amarendra-dianomic:
A comparision of develop branch and this PR:

Activity | develop | FOGL-1943-stats
-- | -- | --
Fogb running for 60 seconds. | Total Sent: 24977 | Total Sent: 24836
grep ">>>>> staistics->update(READINGS"  /var/log/syslog \| wc -l | 28 | 28
grep ">>>>> staistics->update(DISCARDED"  /var/log/syslog \| wc -l | 28 | 28
grep ">>>>> staistics->add_update"  /var/log/syslog \| wc -l | 28 | 28
grep ">>>>> staistics->register"  /var/log/syslog \| wc -l(Actual DB write hit) per asset | 1 | 1
Total DB write hits: | 85 | 85
No. of times Statisitics->_load_keys() is being hit | 28 | 1
Total DB read hits: | 28 | 1
select * from statistics, READINGS | 24977 | 24836
select * from statistics, TI SENSORTAG/LUXOMETER | 24977 | 24836

praveen-garg:
should have been @amarendra-dianomic !

python/foglamp/services/south/ingest.py
```diff
@@ -126,12 +119,6 @@ class Ingest(object):
         category = "{}Advanced".format(cls._parent_service._name)
 
         default_config = {
-            "write_statistics_frequency_seconds": {
```
MarkRiddoch: How is this actually controlled? Maybe the problem is not that we should remove the unused config, but rather we soul be making use of it.

praveen-garg: This is removed via PR https://github.com/foglamp/FogLAMP/pull/1094/files 

More detailed info is in PR comments e.g. https://github.com/foglamp/FogLAMP/pull/1173#issuecomment-428528544 /cc @amarendra-dianomic

amarendra-dianomic: @MarkRiddoch write_statistics_frequency_seconds was in use earlier when we used to execute _write_statistics() via a task scheduled to run every **write_statistics_frequency_seconds** seconds, independent of _insert_readings() task. Now in FOGL-1824, this behaviour has been changed and now _write_statistics() is being run at the end of processing of each readings queue in _insert_readings() task itself. Since _write_statistics() is no longer being run independently,  write_statistics_frequency_seconds has become redundant and hence removed in this PR.

PR #1172, opened 2018-10-10
FOGL-1947: added missing type.id in OMF messages

pintomax:
FOGL-1947: added missing type-id in OMF messages

This allows to change type-id.


tests/unit/C/plugins/common/test_omf_translation.cpp
```diff
@@ -17,6 +17,8 @@
 using namespace std;
 using namespace rapidjson;
 
+#define TYPE_ID "1234"
```
MarkRiddoch: Eventually we want to be able to have type id changed, but I am assuming this is a temporary mechanism

pintomax: Only for tests

PR #1173, opened 2018-10-10
Fogl 1943 - statistics writer restored to be run as periodic task instead of being run with each readings queue

amarendra-dianomic:
Earlier, we used to write readings statistics as an independent task but later we changed it to be run at the end of processing of each readings queue vide [FOGL-1824](https://scaledb.atlassian.net/browse/FOGL-1824). **This affected the throughput adversely**. Therefore, the earlier behaviour has been restored and the concerns of FOGL-1824 have also been addressed.

A comparison of the statistics handling between current develop branch and this PR and another branch (FOGL-1943-stat) is given below:

Activity | develop | FOGL-1943-stat-future* | FOGL-1943-stats
-- | -- | -- | --
Fogb running for 60 seconds. Total Sent:|24977|24887|24836
grep ">>>>> staistics->update(READINGS"  /var/log/syslog \| wc -l | 28 | 14 | 28
grep ">>>>> staistics->update(DISCARDED"  /var/log/syslog \| wc -l | 28 | 14 | 28
grep ">>>>> staistics->add_update"  /var/log/syslog \| wc -l | 28 | 14 | 28
grep ">>>>> staistics->register"  /var/log/syslog \| wc -l(Actual DB write hit) per asset | 1 | 1 | 1
Total DB write hits: | 85 | 43 | 85
No. of times Statisitics->_load_keys() is being hit | 28 | 1 | 1
Total DB read hits: | 28 | 1 | 1
select * from statistics, READINGS | 24977 | 24887 | 24836
select * from statistics, TI SENSORTAG/LUXOMETER | 24977 | 24887 | 24836

*For branch FOGL-1943-stat-future, Ingest-> _write_statistics_frequency_seconds = 5.

### DIFF
--- a/python/foglamp/services/south/ingest.py
+++ b/python/foglamp/services/south/ingest.py
@@ -12,7 +12,6 @@ import time
 import uuid
 from typing import List, Union
 import json
-import copy
 from foglamp.common import logger
 from foglamp.common import statistics
 from foglamp.common.storage_client.exceptions import StorageServerError
@@ -116,6 +115,7 @@ class Ingest(object):
     _payload_events = []
     """The list of unique reading payload for asset tracker"""
 
+    stats = None
     # Configuration (end)
 
     @classmethod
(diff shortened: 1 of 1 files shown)


praveen-garg:
@amarendra-dianomic  I'm okay with this approach, as it restore the old behavior and still manage to handle FOGL-1824. But we are always 5 sec behind in stats the actual ingest.

@MarkRiddoch  Could you please advise on this? I think bulk update in statistics (if planned) will allow us to NOT hogging the statistics writes. the PR #1171 manages to avoid unwanted stats READ calls only.

praveen-garg:
closing in the favor of #1171 . As FOGL-1824 is re-appearing with this.

PR #1175, opened 2018-10-11
Fogl 1874 (Part-B) - Fixed display_name update logic for existing category

ashish-jabble:


ashish-jabble:
@MarkRiddoch displayName fixes for existing category has been fixed. Please have a look.

PR #1176, opened 2018-10-11
FOGL-1813 - script upload API via configuration categories

ashish-jabble:
- [x] script data type supported for config items in configuration manager
- [x] `POST |   /foglamp/category/{category_name}/{config_item}/upload` - Upload script endpoint added and it accepts only **.py** extension
- [x] GET endpoints fixes with **file** attribute if any config item has **script type**

**Response format:**

```
{
    "test": {
        "file": "/home/foglamp/Development/FogLAMP/data/scripts/test_test4_filter.py",
        "type": "script",
        "value": "import sys\nimport json\n\nfilter_config = dict()\n\ndef set_filter_config(configuration):\n\n    global filter_config\n    filter_config = json.loads(configuration['config'])\n    return True\n\ndef readings_filter(readings):\n\n    if ('asset_code' in filter_config):\n        asset_codes = filter_config['asset_code']\n    else:\n        asset_codes = [ ]\n\n    for elem in readings:\n        if (elem['asset_code'] and elem['asset_code'] in asset_codes):\n            reading = elem['reading']\n            # Apply some changes: add 5000 to all datapoints value\n            for key in reading:\n                newVal = reading[key] + 5000\n                reading[key] = newVal\n\n                # Apply max_value check\n                if ('max_value' in filter_config):\n                    maximum_value = int(filter_config['fixed'])\n                    if (reading[key] > maximum_value):\n                        reading[key] = maximum_value\n    return readings\n",
        "description": "d",
        "default": ""
    }
}
```


python/foglamp/services/core/api/configuration.py
```diff
@@ -85,6 +92,14 @@
     if category is None:
         raise web.HTTPNotFound(reason="No such Category found for {}".format(category_name))
 
+    for k, v in category.items():
+        if v['type'] == 'script':
+            prefix_file_name = category_name.lower() + "_" + k.lower() + "_"
```
praveen-garg: right now, if we have a file `dp.py` for category name `CAT #1` and config item `filterScript` then we are saving it 
as `cat #1_filterscript_dp.py` I'm not sure if we want it this way? 

May be we could have saved it in `cat #1_filterscript` directory to keep file name as is. 

In case filename does not matter, and we want the client to expect the name same way. Its okay!

python/foglamp/services/core/api/configuration.py
```diff
@@ -439,3 +460,63 @@
...
+    data = await request.post()
+
+    # contains the name of the file in string format
+    script_file = data.get('script')
+    if not script_file:
+        raise web.HTTPBadRequest(reason="Script file is missing")
+
+    # TODO: For the time being accepted extension is '.py'
+    script_filename = script_file.filename
+    if not script_filename.endswith('.py'):
```
praveen-garg: should we have a list of acceptable extensions on top of this module? Not sure if we want  to allow  more extensions?

python/foglamp/services/core/api/configuration.py
```diff
@@ -439,3 +460,63 @@
...
+            if name.startswith(prefix_file_name):
+                if name != file_name:
+                    os.remove(script_dir + name)
+
+    except Exception as ex:
+        os.remove(script_file_path)
+        raise web.HTTPBadRequest(reason=ex)
+    else:
+        result = await cf_mgr.get_category_item(category_name, config_item)
+        result['file'] = script_file_path
```
praveen-garg: Okay! We are not saving this to database.

ashish-jabble:
@mshariq-nerd I believe you haven't taken the latest pull, #1 is not reproducible on latest commit and for #2 What are those special characters, seems like issue at storage-layer?

praveen-garg:
@ashish-jabble https://github.com/foglamp/foglamp-gui/pull/328#issuecomment-429087436

python/foglamp/services/core/api/configuration.py
```diff
@@ -22,12 +25,16 @@
     | GET            | /foglamp/category/{category_name}                           |
     | GET POST PUT   | /foglamp/category/{category_name}/{config_item}             |
     | DELETE         | /foglamp/category/{category_name}/{config_item}/value       |
+    | POST           | /foglamp/category/{category_name}/{config_item}/upload      |
```
ashish-jabble: @MarkRiddoch I have added an additional endpoint to upload script type and works only if it has category name and config item exists in database. Is it okay? Or should we need to handle in `PUT   | /foglamp/category/{category_name}/{config_item}` itself.

And for add service and add task we have an optional `config` data parameter during this if any of config item has type `script` then it should not work at the moment.

Also for fresh category and merge category if any of config item has type `script` then it also not work at this moment.
Do we really need to handle these situation?
Please advise.

MarkRiddoch: I need to think about how best to handle the merge and add_service/task should work with this. My suggestion is let's get what we have in place as this let's us move forward and solve these other issues, if we need to, in a future Jira ticket.

ashish-jabble: @MarkRiddoch See https://github.com/foglamp/FogLAMP/pull/1186 if it looks good to you.

ashish-jabble:
Also file a JIRA [FOGL-1967](https://scaledb.atlassian.net/browse/FOGL-1967)

python/foglamp/services/core/api/configuration.py
MarkRiddoch: We will want to support more extensions, although at this level the extension is not relevant, it is just a file we are uploading. If the file has an extension on the machine we upload from it should have the same extension when save in the FogLAMP server.

PR #1182, opened 2018-10-15
Ingest->Statistics writer modified to use new bulk update feature

amarendra-dianomic:
Result of running fogb for 60 seconds with coap plugin:

Total recs generated in 60 seconds: 24898
No. of assets: 1
No. of times statistics table hit for updating statistics: 11
No. of times statistics table hit for registering a new key: 1
**Total no. of DB hits: 12**

This is a vast improvement over develop where the it takes **85 hits** to statistics table for the same no. of recs.

### DIFF
--- a/python/foglamp/common/statistics.py
+++ b/python/foglamp/common/statistics.py
@@ -4,6 +4,7 @@
 # See: http://foglamp.readthedocs.io/
 # FOGLAMP_END
 
+import json
 from foglamp.common import logger
 from foglamp.common.storage_client.payload_builder import PayloadBuilder
 from foglamp.common.storage_client.storage_client import StorageClientAsync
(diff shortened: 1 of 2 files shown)


amarendra-dianomic:
@MarkRiddoch regret typo in update count hits in description. Now fixed. This _write_statistics() method is hit at the end of processing of each queue and hence hit count is incremented by one per queue and is actually dependent upon queue size and no. of queues.

python/foglamp/common/statistics.py
```diff
@@ -48,6 +49,31 @@ def __init__(self, storage=None):
         if self._registered_keys is None:
             await self._load_keys()
 
+    async def update_bulk(self, stat_list):
+        """ UPDATE the value column only of a statistics row based on key
```
praveen-garg: Please fix the docstring

PR #1185, opened 2018-10-15
FOGL-1963: PI_Server_V2 is part of FogLAMP repository

pintomax:
FOGL-1963: PI_Server_V2 is part of FogLAMP repository


C/plugins/north/PI_Server_V2/plugin.cpp
```diff
@@ -0,0 +1,269 @@
...
+	 * Allocate the HTTPS handler for "Hostname : port"
+	 * connect_timeout and request_timeout.
+	 * Default is no timeout at all
+	 */
+	connInfo->sender = new SimpleHttps(connInfo->hostAndPort,
+					   connInfo->timeout,
+					   connInfo->timeout);
+  
+	// Allocate the PI Server data protocol
+	connInfo->omf = new OMF(*connInfo->sender,
```
AmandeepArora: Can you please clarify what was the reason to move SimpleHttps and OMF class instantiation to plugin_send?

C/plugins/north/PI_Server_V2/plugin.cpp
```diff
@@ -0,0 +1,269 @@
...
+				connInfo->producerToken);
+
+	// Set OMF FormatTypes  
+	connInfo->omf->setFormatType(OMF_TYPE_FLOAT,
+				     connInfo->formatNumber);
+	connInfo->omf->setFormatType(OMF_TYPE_INTEGER,
+				     connInfo->formatInteger);
+
+	// Send data
+	uint32_t ret = connInfo->omf->sendToServer(readings,
```
AmandeepArora: Need to handle error return values?

C/plugins/north/PI_Server_V2/plugin.cpp
pintomax: This is not a new code, just copying the code form foglamp-north-piserver repo into FogLAMP repo.

The change was made some days ago in order to resolve issues with httpClient class as plugin_init call is made by thread A and plugin_send is called by thread B

C/plugins/north/PI_Server_V2/plugin.cpp
pintomax: ret value is passed to the caller: it's not handled here.

PR #1186, opened 2018-10-16
FOGL-1813 add service with filter separate endpoint added

ashish-jabble:
I have added a separate endpoint for add_service when 'script type'

`POST /foglamp/service/filter` and it accepts Header Content-Type: multipart/form-data

**NOTE:** - 
1) Same apply fix to add_task endpoint, only if get approved
2) The other scenario is PUT config item whose type is script,
`curl -X PUT -H "Content-Type: application/json" -d '{"value": "<some value>" }' http://localhost:8081/foglamp/category/{category_name}/{config_item}` - It wil fail, either we restrict on API side if type is script, then returns BAD request and use ` POST           | /foglamp/category/{category_name}/{config_item}/upload` 
3) For new and merge category, we will update config item whose type is script with `| POST           | /foglamp/category/{category_name}/{config_item}/upload` after new/merge cat operation done


praveen-garg:
@MarkRiddoch this PR is a **proposed** solution to add service for a south plugin with filter config item. https://github.com/foglamp/foglamp-gui/pull/332/files. and solved the problem that we had https://github.com/foglamp/foglamp-gui/pull/328#issuecomment-429087436.

so GUI client shall need to decide at run time to call add service (json payload) vs add service with filter ( form data). Most of this endpoint def code will be moved to def and common to  add_service, except the file handling.

If we have better solution or don't want to solve this add filter script while adding a service, This solution can be discarded.

PR #1198, opened 2018-10-24
[FOGL-1928] delete south service & north instance api support

praveen-garg:
Test scenarios:

1) enabled service
2) disabled service
3) a service with shutdown status [add, enable and (disable or shutdown via curl)]

4) enabled north task (instance)
5) disabled north task (instance)

### DIFF
--- a/python/foglamp/services/core/api/service.py
+++ b/python/foglamp/services/core/api/service.py
@@ -100,7 +100,6 @@ async def delete_service(request):
             raise web.HTTPBadRequest(reason='A service with this name does not exist.')
 
         svc_schedule = result['rows'][0]
-
         sch_id = uuid.UUID(svc_schedule['id'])
         if svc_schedule['enabled'].lower() == 't':
             # disable it
@@ -310,4 +309,3 @@ async def delete_parent_child_configuration(storage, key):
 async def delete_advance_child_configuration(storage, key):
     payload = PayloadBuilder().WHERE(['parent', '=', key]).payload()
     await storage.delete_from_tbl('category_children', payload)
-
--- a/python/foglamp/services/core/api/service.py
+++ b/python/foglamp/services/core/api/service.py
@@ -288,6 +288,7 @@ async def get_schedule(storage, schedule_name):
 
 async def delete_configuration(storage, key):
     await delete_configuration_category(storage, key)
+    await delete_configuration_category(storage, "{}Advanced".format(key))
     await delete_parent_child_configuration(storage, key)
     await delete_advance_child_configuration(storage, key)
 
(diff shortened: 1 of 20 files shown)


python/foglamp/services/core/service_registry/service_registry.py
```diff
@@ -103,6 +103,7 @@ def unregister(cls, service_id):
         """
         expunged_service = cls._expunge(service_id, ServiceRecord.Status.Down)
         cls._logger.info("Stopped {}".format(str(expunged_service)))
+        cls.remove_from_registry(service_id)
```
praveen-garg: May be, We should move this as first action to delete service method, if service's status is DOWN then remove from service registry.

python/foglamp/services/core/api/service.py
```diff
@@ -69,6 +69,50 @@ def get_service_records():
...
+        import uuid
+        sch_id = uuid.UUID(svc_schedule['id'])
+        if svc_schedule['enabled'].lower() == 't':
+            # disable it
+            await server.Server.scheduler.disable_schedule(sch_id)
+        # delete it
+        await server.Server.scheduler.delete_schedule(sch_id)
+
+        # delete config for the service name
+        await revert_configuration(storage, svc)  # Delete configuration entry
```
MarkRiddoch: We should also remove any child categories of the category whose name matches the service.

praveen-garg:
@MarkRiddoch  feedback changes done.  Also fixed for add service, as it was missing `for <name>Advanced`

praveen-garg:
#1201 is provided on top of this for north task instance deletion;

python/foglamp/services/core/api/service.py
```diff
@@ -236,14 +280,34 @@ def get_service_records():
...
-async def revert_configuration(storage, key):
+async def get_schedule(storage, schedule_name):
+    payload = PayloadBuilder().SELECT(["id", "enabled"]).WHERE(['schedule_name', '=', schedule_name]).payload()
+    result = await storage.query_tbl_with_payload('schedules', payload)
+    return result
+
+
+async def delete_configuration(storage, key):
+    await delete_configuration_category(storage, key)
+    await delete_configuration_category(storage, "{}Advanced".format(key))
```
MarkRiddoch: It should not just be a fixed name. What I am saying is that you need to find all the children of this category and delete them all. So go to the category_children table, get all children of this category and delete all of those categories. For each of those categories repeat this until you find no new children. Then remove the entries from the category_children table.

This code should not have any hardwired name, like Advanced in it all all.

praveen-garg: Well, {}Advanced is being deleted from `configuration` table itself. which is required as we have that hardwired insertion via south->ingest.py `        category = "{}Advanced".format(cls._parent_service._name)`

> get all children of this category and delete all of those categories

next 2 def calls are doing the same: delete_parent_child_configuration, delete_advance_child_configuration

I did not find any config thing (including) left in `configuration` or `category_children` table after these calls.

praveen-garg: @MarkRiddoch  FOGL-2014 created.

PR #1201, opened 2018-10-25
[FOGL-1928] delete north instance task api support

praveen-garg:
Quick test output:
     
     curl -X DELETE http://localhost:8081/foglamp/scheduled/task/http%20north
     {"result": "North instance http north deleted successfully."}

### DIFF
--- a/python/foglamp/services/core/api/task.py
+++ b/python/foglamp/services/core/api/task.py
@@ -27,7 +27,7 @@ __version__ = "${VERSION}"
 
 _help = """
     -------------------------------------------------------------------------------
-    | GET POST             | /foglamp/scheduled/task                              |
+    | POST                 | /foglamp/scheduled/task                              |
     | DELETE               | /foglamp/scheduled/task/{task_name}                  |
     -------------------------------------------------------------------------------
 """
@@ -276,7 +276,6 @@ async def delete_task(request):
             raise web.HTTPBadRequest(reason='A north instance task with this name does not exist.')
 
         north_instance_schedule = result['rows'][0]
-
         sch_id = uuid.UUID(north_instance_schedule['id'])
         if north_instance_schedule['enabled'].lower() == 't':
             # disable it
--- a/python/foglamp/services/core/api/task.py
+++ b/python/foglamp/services/core/api/task.py
@@ -5,6 +5,7 @@
 # FOGLAMP_END
 
 import datetime
+import uuid
 
 from aiohttp import web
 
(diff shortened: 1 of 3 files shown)


python/foglamp/services/core/api/task.py
```diff
@@ -215,8 +217,8 @@
                 for k, v in config.items():
                     await config_mgr.set_category_item_value_entry(name, k, v['value'])
         except Exception as ex:
-            await revert_configuration(storage, name)  # Revert configuration entry
-            await revert_parent_child_configuration(storage, name)
+            await delete_configuration(storage, name)  # Revert configuration entry
+            await delete_parent_child_configuration(storage, name)
```
MarkRiddoch: We should also recursively delete child categories of name. E.g. the Advanced children of a south service.

praveen-garg: This is for north instance. the feedback on south endpoint is handled in https://github.com/foglamp/FogLAMP/pull/1198/files#diff-e21909c861b284ca7159d051a8d325d1R289

praveen-garg:
@MarkRiddoch  Please review again, I have left my comment on feedback.

praveen-garg:
the missing relationship delete with`North` category  is fixed.

MarkRiddoch:
> the missing relationship delete with`North` category is fixed.

It is not just the relationship from the North category to our task category that we need to remove. The task category may itself have child categories, so we need to remove those child categories and the relationship between the task category and those children. This could go down many levels, hence the comment about recursively removing categories and relationships.

E.g. If our task is called DeleteMe then the categories could be something like....

North
    |
    +--- DeleteMe
                  |
                  +---- Buffering
                  |
                  +---- Advanced
                                    |
                                     +---- Logging

In this case we would need to delete the categories DeleteMe, Buffering, Advanced and Logging and also remove the relationships that linked them together, e.g.

Parent.      |.   Child
-----------------------
North.       | DeleteMe
DeleteMe | Buffering
DeleteMe | Advanced
Advanced | Logging

It is exactly the same thing we need to do when deleting any category, the descendants must be deleted otherwise there will be configuration categories left "dangling" that are not connected in the hierarchy. It is like deleting a directory (of folder), you need to make sure it is empty first.

praveen-garg:
@MarkRiddoch  would it fine to have a  jira for this recursive deletion and close 1928 with gui integration with whatever we have? 

With the current develop, I don't think this pull request code leave any config for south and north; and above use case may be planned and applied for next features?!

MarkRiddoch:
> @MarkRiddoch would it fine to have a jira for this recursive deletion and close 1928 with gui integration with whatever we have?
> 
> With the current develop, I don't think this pull request code leave any config for south and north; and above use case may be planned and applied for next features?!

OK, please raise a new Jira for the recursive delete.

praveen-garg:
FOGL-2014 recursively delete all child categories and the relationships for delete south service and north task instance created.

PR #1202, opened 2018-10-26
FOGL-1922: Asset filter

AmandeepArora:
Added new repo for asset filter: https://github.com/foglamp/foglamp-filter-asset, please review.

Also removed some timing logs.


C/services/south/ingest.cpp
```diff
@@ -304,7 +299,7 @@ Ingest::~Ingest()
 	delete m_queue;
 	delete m_thread;
 	delete m_statsThread;
-	delete m_data;
+	//delete m_data;
```
MarkRiddoch: Why is this commented out?

If it is no longer required, because something else is deleting this or it is no longer created, then please remove the commented out code. I would also suggest a run with Valgrind to make sure no memory leaks have been introduced.

AmandeepArora: Earlier, sometimes south service didn't shutdown cleanly. Now it takes the required time to insert all pending readings and corresponding stats into DB. m_data is deleted either at end of Ingest::processQueue() or if filtering is used, then in Ingest::useFilteredData(). I was getting double free error during my testing in Ingest destructor, hence I removed this second delete. Anyway, I'll check once again with valgrind. Uptil now, Ingest class is instantiated only once for a south service instance (so all memory is freed at process exit anyway), I guess it is possible to instantiate Ingest class multiple times during the lifetime of a south service for certain use-cases?

AmandeepArora: @MarkRiddoch : It would take a long time to check all possible cases and get all of them correct - with no filters, with empty filtering pipeline, with a filter which doesn't create a new readingSet (e.g. scale filter), a filter which creates a new readingSet (e.g. asset filter) and a mix of such filters in a pipeline. I tried few cases withe valgrind and it is not straight-forward to get all cases correct easily. Also valgrind reports leaks in other classes also. So I would like to spawn a separate activity for that. Since we are not instantiating Ingest class multiple times in the lifetime of a south service, so we can live with shutdown-time minor leak that doesn't grow with number of readings processed. Would you agree? These are leaks of 40 bytes or so.

PR #1207, opened 2018-10-30
FOGL-1986 Perform purge in blocks to allow access from other threads

MarkRiddoch:
Perform the merge process in blocks, with a delay between each block, in order to allow other threads to get access to the SQLite database between blocks.

### DIFF
--- a/C/plugins/storage/sqlite/connection.cpp
+++ b/C/plugins/storage/sqlite/connection.cpp
@@ -26,6 +26,17 @@
 #include <time.h>
 #include <unistd.h>
 #include <utils.h>
+#include <chrono>
+#include <thread>
+
+/*
+ * Control the way purge deletes readings. The block size sets a limit as to how many rows
+ * get deleted in each call, whilst the sleep interval controls how long the thread sleeps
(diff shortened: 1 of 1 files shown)


AmandeepArora:
Should we calculate the number of rows to delete in one query/block based on DB operations' performance? Normally, one iteration of DELETE of 1000 rows would be quite fast on a decently spec'ed system, but it might be too long on low end device like Raspberry Pi. On high end (SSD) system, it might be more optimum to use a larger batch/block size.

And I guess this thread sleep only impacts the purge request and has no impact on any other DB operations in the system? All other parallel DB operations are in separate/different threads?

MarkRiddoch:
Hi @AmandeepArora, yes, other db operations will be on different threads, hence sleeping this thread will allow them in. The times and size of the block are fairly arbitrary, at some point in the future we might add a mechanism, to allow them to be tuned, or better still automatically adjusted based on average durations.

AmandeepArora:
Perfect, thanks.

MarkRiddoch:
> Postgres has better threading/locking mechanisms and hence doesn't suffer from this issue?

Yes, exactly. This is only really an issue in SQLite as far as we know at the moment.

PR #1216, opened 2018-11-07
FOGL-1859: SP-c++/plugins - modify hardcoded configuration.

sstefano69:


C/plugins/north/PI_Server_V2/plugin.cpp
```diff
@@ -41,6 +41,11 @@ using namespace rapidjson;
 				"\"description\": \"The producer token that represents this FogLAMP stream\", " \
 				"\"type\": \"string\", \"default\": \"omf_north_0001\", " \
 				"\"order\": \"2\" }, " \
+			"\"source\": {" \
```
MarkRiddoch: This should be an enumeration rather than a string

sstefano69: you are right, fixing.

sstefano69:
@MarkRiddoch -
the default  for the 'compression' :
* is true in the Python
* is false in the c++

is the c++ right ?

MarkRiddoch:
I suspect having compression on by default is probably a good thing. The ability to turn it off if there is a problem is probably wise to have however.

C/plugins/north/PI_Server_V2/plugin.cpp
sstefano69: fixed

PR #1227, opened 2018-11-12
FOGL-2021: Added handling for OPTIONS pre-flight request (CORS)

AmandeepArora:
Initial fix. Need to confirm from Mark whether this approach seems fine.


MarkRiddoch:
The GUI should not be calling this API - all calls to it should be via the Core and not direct - the fact not supporting OPTIONS blocks this is actually a positive outcome.

AmandeepArora:
@MarkRiddoch @praveen-garg   This issue will be taken care of, in FOGL-2051 by Praveen.

PR #1241, opened 2018-11-19
FOGL-1739: Duplicate readings insertion issue

AmandeepArora:


C/services/storage/storage_api.cpp
```diff
@@ -361,6 +361,37 @@ string  tableName;
...
 string	responsePayload;
 
+	auto header_seq = request->header.find("SeqNum");
+	if(header_seq != request->header.end())
+	{
+		string threadId = header_seq->second.substr(0, header_seq->second.find("_"));
+		int seqNum = stoi(header_seq->second.substr(header_seq->second.find("_")+1));
+
+		{
+			std::unique_lock<std::mutex> lock(mtx_seqnum_map);
```
MarkRiddoch: I would be tempted to limited the mutex to the minimum here. Do the check, set a flag and update the sequence number if required and then release the mutex before doing any processing based on the flag that was set. This mutex will effectively serialise all the south service appending readings and could hence have a big impact on performance.

I.e. don't do all this string manipulation whilst holding the mutex.

C/services/storage/storage_api.cpp
```diff
@@ -518,6 +549,37 @@ void StorageApi::readingAppend(shared_ptr<HttpServer::Response> response, shared
...
 string  responsePayload;
+	
+	auto header_seq = request->header.find("SeqNum");
+	if(header_seq != request->header.end())
+	{
+		string threadId = header_seq->second.substr(0, header_seq->second.find("_"));
+		int seqNum = stoi(header_seq->second.substr(header_seq->second.find("_")+1));
+
+		{
+			std::unique_lock<std::mutex> lock(mtx_seqnum_map);
```
MarkRiddoch: Same as above regarding doing the minimum whilst holding the mutex.

C/services/storage/storage_api.cpp
AmandeepArora: The work done while holding the mutex is a search in a small ordered map (BST), condition  "(seqNum <= it->second)" is false most of the times, and respond function is also just string append, and then there is an update to a BST element's value with new sequence number. The actual DB operation is down below in this function.

On a 3-4 years old i5 laptop running ubuntu, acquiring the mutex and doing the subsequent work while holding the mutex takes 1-2 usec(s). In comparison, DB operations take 5000-30000 usecs, typically.

C/services/storage/storage_api.cpp
```diff
@@ -537,7 +599,7 @@ string  responsePayload;
 			respond(response, SimpleWeb::StatusCode::client_error_bad_request, responsePayload);
 		}
 
-		respond(response, responsePayload);
+		//respond(response, responsePayload);
```
AmandeepArora: In the original code, respond was getting called twice in this function, hence HTTP response was being appended twice alongwith "\r\n\r\n" end marker. I tried just fixing this, but it was not sufficient and I was still getting EoF's in HTTP client code.

PR #1249, opened 2018-11-21
FOGL-2065 Dynamic reconfiguration

MarkRiddoch:


C/services/south/south_plugin.cpp
```diff
@@ -68,7 +68,7 @@ Reading SouthPlugin::poll()
 /**
  * Call the reconfigure method in the plugin
  */
-void SouthPlugin::reconfigure(string& newConfig)
+void SouthPlugin::reconfigure(const string& newConfig)
 {
 	return this->pluginReconfigurePtr(instance, newConfig);
```
AmandeepArora: The south plugins don't have 2nd param as const in:

plugin_reconfigure(PLUGIN_HANDLE *handle, string& newConfig)

But I guess it is not possible to check const-correctness (or even param type) in this case anyway, since we are using dlopen/dlsym mechanism to call plugin_* functions.

C/services/south/south.cpp
```diff
@@ -422,11 +422,19 @@ void SouthService::configChange(const string& categoryName, const string& catego
...
+	if (categoryName.compare(m_name) == 0)
+	{
+		m_config = ConfigCategory(m_name, category);
+		southPlugin->reconfigure(category);
+	}
+	if (categoryName.compare(m_name+"Advanced") == 0)
+	{
+		m_configAdvanced = ConfigCategory(m_name+"Advanced", category);
+		try {
+			m_readingsPerSec = (unsigned long)atoi(m_configAdvanced.getValue("readingsPerSec").c_str());
```
AmandeepArora: There is a separate issue here that m_readingsPerSec, bufferThreshold and maxSendLatency in advanced config don't take effect until next plugin restart. Later 2 params also need to be read from m_configAdvanced here. Later 2 params need to be passed to Ingest class instance. And readingsPerSec would need re-creating timerFd with correct interval.

AmandeepArora: Created FOGL-2097.

PR #1254, opened 2018-11-26
FOGL-1927: Python/PiServer - Add support in SP/Python datatype for arrays of floating point numbers

sstefano69:
working stage

### DIFF
--- a/python/foglamp/plugins/north/common/common.py
+++ b/python/foglamp/plugins/north/common/common.py
@@ -78,30 +78,33 @@ def evaluate_type(value):
      Raises:
      """
 
-    try:
-        float(value)
-
+    if type(value) is list:
+        evaluated_type = "array"
+    else:
         try:
-            # Evaluates if it is a int or a number
-            if str(int(float(value))) == str(value):
+            float(value)
 
-                # Checks the case having .0 as 967.0
-                int_str = str(int(float(value)))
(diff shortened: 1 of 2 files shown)


python/foglamp/plugins/north/common/common.py
```diff
@@ -78,31 +78,34 @@ def evaluate_type(value):
...
 
-                # Checks the case having .0 as 967.0
-                int_str = str(int(float(value)))
-                value_str = str(value)
+                    # Checks the case having .0 as 967.0
+                    int_str = str(int(float(value)))
+                    value_str = str(value)
+
+                    if int_str == value_str:
+                        evaluated_type = "integer"
```
MarkRiddoch: I suspect we want number always - the risk sampling an int value amongst numbers is something we have seen before.

sstefano69: I propose to leave the code as it is, 
as the function evaluate_type is a general one and then the consumer of it decide what to do, 
in the case of the pi_server plugin : 
```
            if item_type == "integer":

                # Forces the creation of integer as number
                omf_type[typename][1]["properties"][item] = {"type":  "number",
                                                             "format": self._config['formatNumber']}

```

sstefano69:
closed - for  #1257

PR #1261, opened 2018-11-30
FOGL-2014 - recursively delete all child categories and the relationships for delete south service and north task instance

amarendra-dianomic:


python/foglamp/common/configuration_manager.py
```diff
@@ -971,6 +971,67 @@ def diff(lst1, lst2):
...
+                raise ValueError(ex)
+            except StorageServerError as ex:
+                err_response = ex.error
+                raise ValueError(err_response)
+            else:
+                return result
+
+        try:
+            result = await delete_recursively(category_name)
+        except Exception as ex:
```
praveen-garg: I think we should expect `ValueError` here also, as its being raised in delete_recursively. 

Are we not writing error to syslog?

python/foglamp/common/configuration_manager.py
```diff
@@ -1026,7 +1015,16 @@ def diff(lst1, lst2):
                 return result
 
         try:
+            deleted_ones = list()
             result = await delete_recursively(category_name)
+            # All children and relationships removed. Remove remaining entries for this cat as child.
+            payload = PayloadBuilder().WHERE(["child", "=", category_name]).payload()
+            result = await self._storage.delete_from_tbl("category_children", payload)
+            if result['response'] == 'deleted':
+                _logger.warning('Deleted parent and/or child entries in catgory_children for %s', category_name)
```
praveen-garg: its an INFO

praveen-garg: Better message will be: `Deleted parent child relationship entries for %s', category_name

python/foglamp/common/configuration_manager.py
```diff
@@ -971,6 +971,65 @@ def diff(lst1, lst2):
 
         return result
 
+    async def delete_recursively_parent_category(self, category_name):
+        """Delete a parent-child relationship for a parent recursively
```
praveen-garg: Delete a category and its child recursively including their parent-child relationship

python/foglamp/common/configuration_manager.py
amarendra-dianomic: The default log level set is warning and it is mandatory to write to syslog for such events, hence level kept as warning.

tests/unit/python/foglamp/common/test_configuration_manager.py
```diff
@@ -2025,6 +2025,10 @@ def mock_coro():
                 assert str(msg) == str(excinfo.value)
         patch_read_cat_val.assert_called_once_with("south")
 
+    @pytest.mark.skip(reason="TODO")
+    async def test_delete_recursively_parent_category(self, reset_singleton):
```
praveen-garg: please add the unit test

python/foglamp/services/core/api/configuration.py
```diff
@@ -505,7 +505,7 @@
 
     cf_mgr = ConfigurationManager(connect.get_storage_async())
     try:
-        await cf_mgr.delete_parent_category(category_name)
+        await cf_mgr.delete_recursively_parent_category(category_name)
```
praveen-garg: I think its main use case is for delete service. We need to call this in api/service.py>delete_service and remove redundant calls from there.

python/foglamp/common/configuration_manager.py
```diff
@@ -971,6 +971,65 @@ def diff(lst1, lst2):
...
+        Keyword Arguments:
+        category_name -- name of the category (required)
+
+        Return Values:
+        JSON
+        """
+        if not isinstance(category_name, str):
+            raise TypeError('category_name must be a string')
+
+        category = await self._read_category_val(category_name)
```
ashish-jabble: Avoid `DB` call if `category_name` exists in cache.

python/foglamp/common/configuration_manager.py
```diff
@@ -971,6 +971,65 @@ def diff(lst1, lst2):
...
+            # All children and relationships removed. Remove remaining entries for this cat as child.
+            payload = PayloadBuilder().WHERE(["child", "=", category_name]).payload()
+            result = await self._storage.delete_from_tbl("category_children", payload)
+            if result['response'] == 'deleted':
+                _logger.info('Deleted parent child relationship entries for %s', category_name)
+            audit = AuditLogger(self._storage)
+            audit_details = {'categoriesDeleted': deleted_ones}
+            await audit.information('CONCH', audit_details)
+        except ValueError as ex:
+            raise ValueError(ex)
```
ashish-jabble: Add `except StorageServerError as ex:`

python/foglamp/common/configuration_manager.py
```diff
@@ -971,6 +971,65 @@ def diff(lst1, lst2):
...
+        try:
+            deleted_ones = list()
+            result = await delete_recursively(category_name)
+            # All children and relationships removed. Remove remaining entries for this cat as child.
+            payload = PayloadBuilder().WHERE(["child", "=", category_name]).payload()
+            result = await self._storage.delete_from_tbl("category_children", payload)
+            if result['response'] == 'deleted':
+                _logger.info('Deleted parent child relationship entries for %s', category_name)
+            audit = AuditLogger(self._storage)
+            audit_details = {'categoriesDeleted': deleted_ones}
```
ashish-jabble: Remove it from `configuration cache` if exists

python/foglamp/common/configuration_manager.py
```diff
@@ -971,6 +971,65 @@ def diff(lst1, lst2):
...
+            await audit.information('CONCH', audit_details)
```
ashish-jabble: Add FIXME `FOGL-2140` to use right code, as we are deleting category.

python/foglamp/common/configuration_manager.py
```diff
@@ -971,6 +971,65 @@ def diff(lst1, lst2):
...
+                payload = PayloadBuilder().WHERE(["key", "=", cat]).payload()
+                result = await self._storage.delete_from_tbl("configuration", payload)
+                if result['response'] == 'deleted':
+                    _logger.info('Deleted parent category from configuration: %s', cat)
+                    deleted_ones.append(cat)
+            except KeyError as ex:
+                raise ValueError(ex)
+            except StorageServerError as ex:
+                err_response = ex.error
+                raise ValueError(err_response)
```
ashish-jabble: Rollback `parent-child relation`, if exception occurred in between?

amarendra-dianomic:
@praveen-garg @ashish-jabble feedback changes done. Pl check.

python/foglamp/common/configuration_manager.py
```diff
@@ -971,6 +975,75 @@ def diff(lst1, lst2):
 
         return result
 
+    async def delete_recursively_parent_category(self, category_name):
```
MarkRiddoch: The name of this is misleading, it implies it is deleting a parent of the category, which it is not.

amarendra-dianomic:
@MarkRiddoch method name changed from delete_recursively_parent_category to delete_parent_category_recursively in all files. Pl check.

amarendra-dianomic:
@MarkRiddoch I have changed the name to "delete_children_categories" and have added a new route "DELETE /foglamp/category/<category>/children" in order to completely remove the ambiguity. Please check.

python/foglamp/services/core/routes.py
```diff
@@ -63,6 +63,7 @@ def setup(app):
     app.router.add_route('PUT', '/foglamp/category/{category_name}', api_configuration.update_configuration_item_bulk)
     app.router.add_route('POST', '/foglamp/category/{category_name}/children', api_configuration.create_child_category)
     app.router.add_route('GET', '/foglamp/category/{category_name}/children', api_configuration.get_child_category)
+    app.router.add_route('DELETE', '/foglamp/category/{category_name}/children', api_configuration.delete_children_categories)
```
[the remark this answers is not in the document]
praveen-garg: though ticket's scope is limited to internal calls only! i.e. delete service, delete task, delete notification.

Though if we add here, we should consider to add to core management api too `python/foglamp/services/core/server.py`

amarendra-dianomic:
@MarkRiddoch we need architectural clarity here. Please note that,
1) Earlier, we had wrongly reused and redefined "DELETE /foglamp/category/\<category_name>/parent" route. Earlier this route was being used to delete relationship between a parent and its children, in category_children table only and NOT from configuration table. This has now been restored.
2) Since we now want to delete a category (aka parent) and its children from BOTH configuration table and category_children table, we need to define a new route. I proposed and used "DELETE /foglamp/category/\<category_name>/children" route. @praveen-garg has proposed to use "DELETE /foglamp/category/\<category_name>" route instead. IMHO, it has the potential for being misunderstood with deleting ONLY the category and it may not reflect the correct intent to delete both the category and its children.
3) Although I agree with @praveen-garg's observation that if we add the DELETE route here, we should consider to add to core management api too python/foglamp/services/core/server.py.
4) Regarding method name, IMHO, "delete_children_categories" is good enough but I do not have any objection to using "delete_category_and_children_recursively".

Please advise.

python/foglamp/services/core/api/configuration.py
```diff
@@ -511,7 +511,32 @@
     except ValueError as ex:
         raise web.HTTPNotFound(reason=str(ex))
 
-    return web.json_response({"message": "Parent-child relationship for the parent-{} is deleted".format(category_name)})
+    return web.json_response({"message": "Parent category-{} is deleted".format(category_name)})
```
praveen-garg: Restore this message. This is just an old def with incorrect name. i.e. delete_parent_category > delete_parent_relation()

amarendra-dianomic:
@MarkRiddoch @praveen-garg So final changes are here as below:
1. Rest api endpoint exposing categories deletion has been removed. Now recursive deletion is a private def within configuration_manager.py.
2. Method name changed to **delete_category_and_children_recursively**.
3. Testing this PR is possible only vide delete_service, delete_task and delete_notification methods/rest end points. Multiple scenario should be constructed manually involving service/task/notitication.

Hope this is OK now.

praveen-garg:
Yeah, This is what its doing and handles all use cases. FOGL-2150 is to Add new endpoint ` Delete Category` to Admin (REST) / Core Management Api.

praveen-garg:
Yeah, This is what its doing and handles all use cases. FOGL-2150 is to Add new endpoint ` Delete Category` to Admin (REST) / Core Management Api and expose this method to them.

PR #1279, opened 2018-12-11
filters.py refactored to facilitate proper unit tests

amarendra-dianomic:


python/foglamp/services/core/api/filters.py
```diff
@@ -547,3 +437,65 @@ def new_items(lst1, lst2):
...
+            parent = k
+            _delete_keys_from_dict(v, lst_keys, deleted_values, parent)
+    return dict_del, deleted_values
+
+
+async def _delete_child_filters(storage: StorageClientAsync, cf_mgr: ConfigurationManager, service_name: str,
+                                   new_list: List[str], old_list: List[str] = []) -> None:
+    # Difference between pipeline and value from storage lists and then delete relationship as per diff
+    delete_children = _diff(new_list, old_list)
+    for l in delete_children:
```
praveen-garg: l?

python/foglamp/services/core/api/filters.py
```diff
@@ -547,3 +437,65 @@ def new_items(lst1, lst2):
...
+        try:
+            filter_child_category_name = "{}_{}".format(service_name, l)
+            await cf_mgr.delete_child_category(service_name, filter_child_category_name)
+        except:
+            pass
+        payload = PayloadBuilder().WHERE(['key', '=', "{}_{}".format(service_name, l)]).payload()
+        await storage.delete_from_tbl("configuration", payload)
```
praveen-garg: use _delete_configuration_category to delete from cache too

python/foglamp/services/core/api/filters.py
```diff
@@ -547,3 +437,65 @@ def new_items(lst1, lst2):
...
+
+
+async def _add_child_filters(storage: StorageClientAsync, cf_mgr: ConfigurationManager, service_name: str,
+                               filter_list: List[str], old_list: List[str] = []) -> None:
+    # Create children categories. Since create_category() does not expect "value" key to be
+    # present in the payload, we need to remove all "value" keys BUT need to add back these
+    # "value" keys to the new configuration.
+    for filter_name in filter_list:
+        filter_config = await cf_mgr.get_category_all_items(category_name=filter_name)
+        filter_desc = "Configuration of {} filter for user {}".format(service_name, filter_name)
```
praveen-garg: swap `service_name, filter_name` args

praveen-garg: I think `service_` should not be used, as filters can be applied with north instances too.

python/foglamp/services/core/api/filters.py
```diff
@@ -516,6 +405,7 @@ def new_items(lst1, lst2):
     else:
         return web.json_response({'result': "Filter {} deleted successfully".format(filter_name)})
 
+
 async def delete_filter_pipeline(request: web.Request) -> web.Response:
     """ DELETE filter pipeline
```
praveen-garg: We should not make http request here, its a call within same API. so better we clone to PUT empty or create a common def to be called for PUT and `DELETE  == PUT empty`

python/foglamp/services/core/api/filters.py
```diff
@@ -197,174 +183,77 @@
...
-
-        # Fetch up-to-date category items
+            await cf_mgr.create_category(category_name=service_name, category_value=new_item, keep_original_items=True)
+            await _add_child_filters(storage, cf_mgr, service_name, filter_list)
+
+        # Fetch up-to-date category item
         result = await cf_mgr.get_category_item(service_name, config_item)
         if result is None:
-            # Error config_item doesn't exist
             message = "No detail found for the category_name: {} and config_item: {}".format(service_name, config_item)
```
praveen-garg: better to say, `"No detail found for the user: {} and filter: {}"`

python/foglamp/services/core/api/filters.py
amarendra-dianomic: request object cloning does allow adding a PUT/POST payload, hence this approach. See https://docs.aiohttp.org/en/stable/web_reference.html#aiohttp.web.BaseRequest.clone.

amarendra-dianomic:
@praveen-garg feedback fixes done. Pl check.

PR #1289, opened 2018-12-18
FOGL-2199, 2200 - notification routes fixed. JSON return fixed. Tests fixed.

amarendra-dianomic:


amarendra-dianomic:
@MarkRiddoch this PR deals with two issues:
a) As @pintomax pointed out that we wrongly prefixed routes, meant for notification server only, with "/foglamp" which has now been fixed, and
b) The json returned by GET /notification/rules and GET /notification/delivery now returns the **list** of **dicts**, one per plugin, where each dict is the return value of plugin_info() call. Earlier each rest call returned a single large dict. As @pintomax mentioned, it was deviation from the doc which has now been fixed.

PR #1291, opened 2018-12-18
FOGL-2160 Make south poll rate dynamically configurable

MarkRiddoch:


C/services/south/south.cpp
```diff
@@ -507,7 +507,13 @@ void SouthService::configChange(const string& categoryName, const string& catego
 	{
 		m_configAdvanced = ConfigCategory(m_name+"Advanced", category);
 		try {
-			m_readingsPerSec = (unsigned long)strtol(m_configAdvanced.getValue("readingsPerSec").c_str(), NULL, 10);
+			unsigned long newval = (unsigned long)strtol(m_configAdvanced.getValue("readingsPerSec").c_str(), NULL, 10);
+			if (newval != m_readingsPerSec)
+			{
+				m_readingsPerSec = newval;
+				close(m_timerfd);
```
AmandeepArora: Closing and re-creating m_timerfd can interleave with timerfd_create and/or read on m_timerfd in the other thread. I am not sure if this interleaving would be thread-safe.

MarkRiddoch: The other thread gets an error and then continues with the new value. So it seems to be fine.

AmandeepArora: close(2) manpage says "It is probably unwise to close file descriptors while they may be in use by system calls in other threads in the same process. Since a file descriptor may be reused, there are some obscure race conditions that may cause unintended side effects.". 

Not sure what these side effects could be. They may be insignificant when dealing with a timer FD but significant when dealing with actual data files.

Also it is possible to just change the timer notification interval by using timerfd_settime on the same timer FD and avoid close & re-create.

I was also planning to make this change few days back. I was thinking of setting an atomic<bool> flag in configChange that the poll loop can see and do the needful then.

PR #1292, opened 2018-12-18
FOGL-2203: addition of scanPluginDir method

pintomax:
FOGL-2203: addition of scanPluginDir method

Example usage:

        PluginManager* plugins = PluginManager::getInstance();
        list<std::string> pList;
        plugins->scanPluginDir("notificationRule", pList);


C/services/common/plugin_manager.cpp
```diff
@@ -225,3 +226,39 @@ PLUGIN_HANDLE PluginManager::resolveSymbol(PLUGIN_HANDLE handle, const string& s
...
+	char *home = getenv("FOGLAMP_ROOT");
+	if (home)
+	{
+		struct dirent *entry;
+		DIR *dp;
+		string path = home;
+		path += "/plugins/" + type + "/";
+
+		// Open the plugins dir/type
+		dp = opendir(path.c_str());
```
AmandeepArora: May be check the return value of opendir() for NULL.

C/services/common/include/plugin_manager.h
```diff
@@ -25,23 +25,34 @@
...
+		PLUGIN_HANDLE	loadPlugin(const std::string& name,
+					   const std::string& type);
+		void		unloadPlugin(PLUGIN_HANDLE handle);
+		void*		resolveSymbol(PLUGIN_HANDLE handle,
+					      const std::string& symbol);
+		PLUGIN_HANDLE	findPluginByName(const std::string& name);
+		PLUGIN_HANDLE	findPluginByType(const std::string& type);
+		PLUGIN_INFORMATION
+				*getInfo(const PLUGIN_HANDLE);
+		void		scanPluginDir(const std::string& type,
```
MarkRiddoch: Please call tis something that reflects what we are doing not the mechanism it uses. E.g. getInstalledPlugins

C/services/common/plugin_manager.cpp
```diff
@@ -225,3 +226,39 @@ PLUGIN_HANDLE PluginManager::resolveSymbol(PLUGIN_HANDLE handle, const string& s
...
+
+		// Get all plugin name directories
+		while ((entry = readdir(dp)))
+		{
+			if (strcmp (entry->d_name, "..") != 0 &&
+			    strcmp (entry->d_name, ".") != 0)
+			{
+				// Load plugin, given its name
+				loadPlugin(entry->d_name, type);
```
MarkRiddoch: This could fail if the file is not a valid plugin or is of the wrong type.

Load plugin does not take a filename as an argument, it takes a plugin name, they are not the same.

C/services/common/plugin_manager.cpp
```diff
@@ -225,3 +226,39 @@ PLUGIN_HANDLE PluginManager::resolveSymbol(PLUGIN_HANDLE handle, const string& s
...
+				// Add name to ouput list
+				plugins.push_back(entry->d_name);
```
MarkRiddoch: This is adding the file name, notches plugin name.

pintomax: entry->d_name is the subdirectory name in plugins/type

plugins/filter/
                      delta/libdelta.so
                      scale/libscale.so

taking "delta" and passinf it to loadPlugin() works, at least with C plugins

Another example:

plugins/north/
                  PI_Server_V2/libPI_Server_V2.so

MarkRiddoch: ok, thanks

PR #1293, opened 2018-12-19
FOGL-2188: Memory leak in storage service duplicate tracking

AmandeepArora:
Implemented a LRU/MRU mechanism so that storage service stores a max of last 16 most recently used mappings from "thread ID" to "last seen HTTP request sequence number for that thread ID". So this implements ageing of these mapping entries indirectly.


AmandeepArora:
@MarkRiddoch : Code cleanup done.

PR #1294, opened 2018-12-19
FogL 2182 - safe mode startup option

ashish-jabble:
**Fixed items**

- [x] foglamp start --safe-mode added
- [x] text added when run in safe mode for `foglamp start and foglamp status`
- [x] safeMode added in /ping, related ticket for GUI https://scaledb.atlassian.net/browse/FOGL-2216
- [x] Changes required for Scheduler code as well
- [x] Audit message updated for both modes on start/stop
- [x]  Restart should be normal mode

### DIFF
--- a/python/foglamp/services/core/server.py
+++ b/python/foglamp/services/core/server.py
@@ -404,7 +404,7 @@ class Server:
         """Starts the scheduler"""
         _logger.info("start scheduler")
         cls.scheduler = Scheduler(cls._host, cls.core_management_port)
-        await cls.scheduler.start()
+        await cls.scheduler.start(cls.running_in_safe_mode)
 
     @staticmethod
     def __start_storage(host, m_port):
@@ -617,11 +617,10 @@ class Server:
             cls._configuration_manager = ConfigurationManager(cls._storage_client_async)
             cls._interest_registry = InterestRegistry(cls._configuration_manager)
 
-            if not cls.running_in_safe_mode:
-                # start scheduler
-                # see scheduler.py start def FIXME
-                # scheduler on start will wait for storage service registration
-                loop.run_until_complete(cls._start_scheduler())
+            # start scheduler
+            # see scheduler.py start def FIXME
+            # scheduler on start will wait for storage service registration
+            loop.run_until_complete(cls._start_scheduler())
 
             # start monitor
             loop.run_until_complete(cls._start_service_monitor())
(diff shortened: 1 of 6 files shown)


python/foglamp/services/core/server.py
```diff
@@ -582,8 +585,10 @@ def _check_readings_table(cls, loop):
 
     @classmethod
     def _start_core(cls, loop=None):
-        _logger.info("start core")
+        if cls.running_in_safe_mode:
```
amarendra-dianomic: An explanatory comment will be better to explain what behaviour is expected from core running in safe-mode.

scripts/foglamp
```diff
@@ -22,7 +22,7 @@ set -e
 #
 # This is the startup script for foglamp
 #
-USAGE="Usage: `basename ${0}` {start|stop|status|reset|kill|help|version}"
+USAGE="Usage: `basename ${0}` {start|start --safe-mode|stop|status|reset|kill|help|version}"
```
amarendra-dianomic: Seems that only "--" is prefixed to safe-mode only.

ashish-jabble: JIRA has asked for `--safe_mode` and I only replaced underscore with hyphen i.e `--safe-mode `

praveen-garg: @amarendra-dianomic  Yeah. As it is a  flag with start, hence `start --safe-mode`

python/foglamp/services/core/server.py
```diff
@@ -582,8 +585,10 @@ def _check_readings_table(cls, loop):
...
+            _logger.info("Starting in safe mode")
 
+        _logger.info("start core")
```
praveen-garg: Let's  log  once. `_logger.info("Starting ...") and _logger.info("Starting in SAFE MODE...")` based on condition

praveen-garg:
In safe mode, I see syslog as:

```
Dec 20 17:12:15 FogLAMP[10098] INFO: server: foglamp.services.core.server: start scheduler
Dec 20 17:12:15 FogLAMP[10098] INFO: scheduler: foglamp.services.core.scheduler.scheduler: Starting Scheduler: Management port received is 33715
Dec 20 17:12:15 FogLAMP[10098] INFO: scheduler: foglamp.services.core.scheduler.scheduler: Scheduled task for schedule 'purge' to start at 2018-12-20 18:12:15.390403
Dec 20 17:12:15 FogLAMP[10098] INFO: scheduler: foglamp.services.core.scheduler.scheduler: Scheduled task for schedule 'stats collection' to start at 2018-12-20 17:12:30.390403
Dec 20 17:12:15  FogLAMP[10098] INFO: scheduler: foglamp.services.core.scheduler.scheduler: Scheduled task for schedule 'certificate checker' to start at ...
```

We should not!

the next log statement to 
```
Dec 20 17:12:15 FogLAMP[10098] INFO: server: foglamp.services.core.server: start scheduler
```
should be

```
In safe mode, the scheduler will be in restricted mode. Only the Admin API operations and current state of schedules will be accessible; No jobs / processes will be triggered.
```

PR #1299, opened 2018-12-21
FOGL-2220 - JSON handling fixed for filter api

amarendra-dianomic:


praveen-garg:
FOGL-2229 should also be fixed with 4e994d5 .

PR #1304, opened 2018-12-27
FOGL-2180: system test added for end to end with csv plugin

Singhal-Vaibhav:
End to end test with CSV plugin to PI written in pytest.
**Advantages of test over bash test suite:**

- Less lines of code (~190 as compared to existing bash suit > 300)
- Reusable components (conftest.py)
- Assertions at every point
- Cleanup handled, test cleans up csv created and data sent to pi and subsequent test will start with fresh environment (both at FogLAMP side and PI side)
- Execution time ~45 sec as compared to bash test suit (>2 minutes)
- Uses arg parser, test can be configured with different arguments from command line
- Junit style framework, no need for pytest wrapper as compared to bash test suite
- CI friendly (CI expects tests to be written in junit style framework)

**Usage:**
```tests/system/python $ pytest test_end_to_end_csv.py --pi_db=<pi_db_name> --pi_host=<pi_server_host> --pi_admin=<pi_machine_user> --pi_passwd=<pi_machine_password> --pi_token=<pi_producer_token>"```

**Result:**
```====== 1 passed in 45.42 seconds =====```

### DIFF
--- a/tests/system/python/conftest.py
+++ b/tests/system/python/conftest.py
@@ -1,0 +1,108 @@
+# -*- coding: utf-8 -*-
+
+# FOGLAMP_BEGIN
+# See: http://foglamp.readthedocs.io/
+# FOGLAMP_END
+
+""" Configuration system/python/conftest.py
+
+"""
+import pytest
+import subprocess
+import os
+
+__author__ = "Vaibhav Singhal"
+__copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
+__license__ = "Apache 2.0"
+__version__ = "${VERSION}"
+
+
+@pytest.fixture
+def reset_foglamp_start():
+    assert os.environ.get('FOGLAMP_ROOT') is not None
+    subprocess.run(["$FOGLAMP_ROOT/scripts/foglamp kill"], shell=True, check=True)
+    subprocess.run(["echo YES | $FOGLAMP_ROOT/scripts/foglamp reset"], shell=True, check=True)
+    subprocess.run(["$FOGLAMP_ROOT/scripts/foglamp start"], shell=True)
+    stat = subprocess.run(["$FOGLAMP_ROOT/scripts/foglamp status"], shell=True, stdout=subprocess.PIPE)
+    assert "FogLAMP not running." not in stat.stdout.decode("utf-8")
+
+
+def pytest_addoption(parser):
+    parser.addoption("--foglamp_url", action="store", default="localhost:8081",
+                     help="foglmap client api url")
+    parser.addoption("--pi_host", action="store", default="pi-server",
+                     help="PI Server Host Name/IP")
+    parser.addoption("--pi_port", action="store", default="5460",
+                     help="PI Server PORT")
+    parser.addoption("--pi_db", action="store", default="pi-server-db",
+                     help="PI Server database")
+    parser.addoption("--pi_admin", action="store", default="pi-server-uid",
+                     help="PI Server user login")
+    parser.addoption("--pi_passwd", action="store", default="pi-server-pwd",
+                     help="PI Server user login password")
+    parser.addoption("--pi_token", action="store", default="omf_north_0001",
+                     help="OMF Producer Token")
+    parser.addoption("--south_plugin_repo", action="store", default="foglamp-south-playback",
+                     help="South Repo to clone")
+    parser.addoption("--south_plugin", action="store", default="playback",
+                     help="Name of the South Plugin")
+    parser.addoption("--north_plugin", action="store", default="PI_Server_V2",
+                     help="Name of the North Plugin")
+    parser.addoption("--asset_name", action="store", default="systemtest",
+                     help="Name of asset")
+
+
+@pytest.fixture
+def foglamp_url(request):
+    return request.config.getoption("--foglamp_url")
+
+
+@pytest.fixture
+def pi_host(request):
+    return request.config.getoption("--pi_host")
+
+
+@pytest.fixture
+def pi_port(request):
+    return request.config.getoption("--pi_port")
+
+
+@pytest.fixture
+def pi_db(request):
+    return request.config.getoption("--pi_db")
+
+
+@pytest.fixture
+def pi_admin(request):
+    return request.config.getoption("--pi_admin")
+
+
+@pytest.fixture
+def pi_passwd(request):
+    return request.config.getoption("--pi_passwd")
+
+
+@pytest.fixture
+def pi_token(request):
+    return request.config.getoption("--pi_token")
+
+
+@pytest.fixture
+def south_plugin_repo(request):
+    return request.config.getoption("--south_plugin_repo")
+
+
+@pytest.fixture
+def south_plugin(request):
+    return request.config.getoption("--south_plugin")
+
+
+@pytest.fixture
+def north_plugin(request):
+    return request.config.getoption("--north_plugin")
+
+
+@pytest.fixture
+def asset_name(request):
+    return request.config.getoption("--asset_name")
+
--- a/tests/system/python/conftest.py
+++ b/tests/system/python/conftest.py
@@ -7,9 +7,9 @@
 """ Configuration system/python/conftest.py
 
 """
-import pytest
 import subprocess
 import os
+import pytest
 
 __author__ = "Vaibhav Singhal"
 __copyright__ = "Copyright (c) 2018 Dianomic Systems"
--- a/tests/system/python/conftest.py
+++ b/tests/system/python/conftest.py
@@ -12,13 +12,13 @@ import subprocess
 import os
 
 __author__ = "Vaibhav Singhal"
-__copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
+__copyright__ = "Copyright (c) 2018 Dianomic Systems"
 __license__ = "Apache 2.0"
 __version__ = "${VERSION}"
 
 
 @pytest.fixture
-def reset_foglamp_start():
+def reset_and_start_foglamp():
     assert os.environ.get('FOGLAMP_ROOT') is not None
     subprocess.run(["$FOGLAMP_ROOT/scripts/foglamp kill"], shell=True, check=True)
     subprocess.run(["echo YES | $FOGLAMP_ROOT/scripts/foglamp reset"], shell=True, check=True)
@@ -42,8 +42,6 @@ def pytest_addoption(parser):
                      help="PI Server user login password")
     parser.addoption("--pi_token", action="store", default="omf_north_0001",
                      help="OMF Producer Token")
-    parser.addoption("--south_plugin_repo", action="store", default="foglamp-south-playback",
-                     help="South Repo to clone")
     parser.addoption("--south_plugin", action="store", default="playback",
                      help="Name of the South Plugin")
     parser.addoption("--north_plugin", action="store", default="PI_Server_V2",
@@ -85,11 +83,6 @@ def pi_passwd(request):
 @pytest.fixture
 def pi_token(request):
     return request.config.getoption("--pi_token")
-
-
-@pytest.fixture
-def south_plugin_repo(request):
-    return request.config.getoption("--south_plugin_repo")
 
 
 @pytest.fixture
--- a/tests/system/python/test_end_to_end_csv.py
+++ b/tests/system/python/test_end_to_end_csv.py
@@ -67,7 +67,7 @@ def _start_folamp_south(south_plugin, asset_name, foglamp_url):
 
 
 def _start_folamp_north(foglamp_url, pi_host, pi_port, north_plugin, pi_token):
-    """Start north service"""
+    """Start north task"""
 
     conn = http.client.HTTPConnection(foglamp_url)
     data = {"name": "North_Readings_to_PI",
@@ -199,7 +199,7 @@ def start_south_north(reset_and_start_foglamp, south_plugin, asset_name, foglamp
     # Start foglamp south service
     csv_file_path = _start_folamp_south(south_plugin, asset_name, foglamp_url)
 
-    # Start foglamp north service
+    # Start foglamp north task
     _start_folamp_north(foglamp_url, pi_host, pi_port, north_plugin, pi_token)
 
     # Provide the fixture value
--- a/tests/system/python/test_end_to_end_csv.py
+++ b/tests/system/python/test_end_to_end_csv.py
@@ -7,7 +7,6 @@
 """ Test system/python/end_to_end_csv.py
 
 """
-import pytest
 import os
 import subprocess
 import http.client
@@ -16,6 +15,9 @@ import time
 import shutil
 import base64
 import ssl
+import pytest
+from foglamp.common.common import _FOGLAMP_ROOT
+print(_FOGLAMP_ROOT)
 
 __author__ = "Vaibhav Singhal"
 __copyright__ = "Copyright (c) 2018 Dianomic Systems"
@@ -30,12 +32,14 @@ CSV_DATA = 2.5
 # Number of tries to make to read PI data and to read North configuration
 NUM_RETRIES = 3
 # Low sleep interval between each tries, used in internal foglamp process waits
-SLEEP_INTERVAL_LOW = 3
+SLEEP_INTERVAL_LOW = 5
 # High sleep interval between each tries, used in external pi webapi waits
 SLEEP_INTERVAL_HIGH = 10
+# Name of the North Task
+NORTH_TASK_NAME = "North_Readings_to_PI"
 
 
-def _start_folamp_south(south_plugin, asset_name, foglamp_url):
+def _start_foglamp_south(south_plugin, asset_name, foglamp_url):
     """Start south service"""
 
     south_config = {"assetName": {"value": "{}".format(asset_name)}, "csvFilename": {"value": "{}".format(CSV_NAME)},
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: tests/system/python/conftest.py [hunk 1]
```diff
@@ -0,0 +1,108 @@
...
+import os
+
+__author__ = "Vaibhav Singhal"
+__copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
+__license__ = "Apache 2.0"
+__version__ = "${VERSION}"
+
+
+@pytest.fixture
+def reset_foglamp_start():
```
praveen-garg: reset_and_start_foglamp

tests/system/python/conftest.py
```diff
@@ -0,0 +1,108 @@
...
+    assert os.environ.get('FOGLAMP_ROOT') is not None
```
praveen-garg: should we check what should be the exact value

tests/system/python/conftest.py
```diff
@@ -0,0 +1,108 @@
...
+    subprocess.run(["$FOGLAMP_ROOT/scripts/foglamp kill"], shell=True, check=True)
+    subprocess.run(["echo YES | $FOGLAMP_ROOT/scripts/foglamp reset"], shell=True, check=True)
+    subprocess.run(["$FOGLAMP_ROOT/scripts/foglamp start"], shell=True)
+    stat = subprocess.run(["$FOGLAMP_ROOT/scripts/foglamp status"], shell=True, stdout=subprocess.PIPE)
+    assert "FogLAMP not running." not in stat.stdout.decode("utf-8")
```
praveen-garg: Can we please assert its running with VERSION variable

tests/system/python/conftest.py
```diff
@@ -0,0 +1,108 @@
...
+    parser.addoption("--pi_db", action="store", default="pi-server-db",
+                     help="PI Server database")
+    parser.addoption("--pi_admin", action="store", default="pi-server-uid",
+                     help="PI Server user login")
+    parser.addoption("--pi_passwd", action="store", default="pi-server-pwd",
+                     help="PI Server user login password")
+    parser.addoption("--pi_token", action="store", default="omf_north_0001",
+                     help="OMF Producer Token")
+    parser.addoption("--south_plugin_repo", action="store", default="foglamp-south-playback",
+                     help="South Repo to clone")
```
praveen-garg: South plugin github repository to clone

tests/system/python/conftest.py
```diff
@@ -0,0 +1,108 @@
...
+    parser.addoption("--south_plugin", action="store", default="playback",
```
praveen-garg: we should not need `--south_plugin_repo` as this can be used with `foglamp-south-$NAME`

tests/system/python/test_end_to_end_csv.py
```diff
@@ -0,0 +1,196 @@
...
+    r.read().decode()
+
+    data = {"schedule_name": "{}".format("North_Readings_to_PI")}
+    conn.request("PUT", '/foglamp/schedule/enable', json.dumps(data))
+    r = conn.getresponse()
+    assert 200 == r.status
+    r.read().decode()
+    yield start_south_north
+    os.remove(os.path.join(os.path.expandvars('$FOGLAMP_ROOT/data/{}'.format(CSV_NAME))))
+    shutil.rmtree("/tmp/{}".format(south_plugin_repo), ignore_errors=True)
```
praveen-garg: can this be broken into 3-4 smaller internal def ?

tests/system/python/test_end_to_end_csv.py
```diff
@@ -0,0 +1,196 @@
...
+    headers = {'Authorization': 'Basic %s' % userAndPass}
+    conn = http.client.HTTPSConnection(host, context=ssl._create_unverified_context())
+
+    try:
+        conn.request("GET", '/piwebapi/assetservers', headers=headers)
+        res = conn.getresponse()
+        r = json.loads(res.read().decode())
+        dbs = r["Items"][0]["Links"]["Databases"]
+
+        if dbs is not None:
```
praveen-garg: what if this is None

tests/system/python/test_end_to_end_csv.py
```diff
@@ -0,0 +1,196 @@
...
+            conn.request("GET", dbs, headers=headers)
+            res = conn.getresponse()
+            r = json.loads(res.read().decode())
+            for el in r["Items"]:
+                if el["Name"] == pi_database:
+                    element = el["Links"]["Elements"]
+
+        if element is not None:
+            conn.request("GET", element, headers=headers)
+            res = conn.getresponse()
```
praveen-garg: as above

tests/system/python/test_end_to_end_csv.py
```diff
@@ -0,0 +1,196 @@
...
+            r = json.loads(res.read().decode())
+            url_elements_list = r["Items"][0]["Links"]["Elements"]
+
+        if url_elements_list is not None:
```
praveen-garg: as above

tests/system/python/test_end_to_end_csv.py
```diff
@@ -0,0 +1,196 @@
...
+            conn.request("GET", url_elements_list, headers=headers)
+            res = conn.getresponse()
+            r = json.loads(res.read().decode())
+            items = r["Items"]
+            for el in items:
+                if el["Name"] == asset:
+                    url_assets_list = el["Links"]["EndValue"]
+                    web_id = el["WebId"]
+
+        if url_assets_list is not None:
```
praveen-garg: as above

tests/system/python/test_end_to_end_csv.py
```diff
@@ -0,0 +1,196 @@
...
+    shutil.rmtree("/tmp/{}".format(south_plugin_repo), ignore_errors=True)
+
+
+def _read_data_from_pi(host, port, admin, password, pi_database, asset):
+    """ This method reads data from pi web api """
+    dbs = None
+    element = None
+    url_elements_list = None
+    url_assets_list = None
+    web_id = None
```
praveen-garg: what is a web_id ?

tests/system/python/test_end_to_end_csv.py
```diff
@@ -0,0 +1,196 @@
...
+    conn = http.client.HTTPConnection(foglamp_url)
+    time.sleep(10)
+    conn.request("GET", '/foglamp/asset')
+    r = conn.getresponse()
+    assert 200 == r.status
+    r = r.read().decode()
+    retval = json.loads(r)
+    assert retval[0]["assetCode"] == asset_name
+    assert retval[0]["count"] == 1
+    conn.request("GET", '/foglamp/asset/{}'.format(asset_name))
```
praveen-garg: Add a new line

tests/system/python/conftest.py
Singhal-Vaibhav: We cant, as we are not sure where will the end user clone foglamp, it can be /home/ubuntu/FogLAMP or /home/bla/Development/FogLAMP

tests/system/python/conftest.py
Singhal-Vaibhav: foglamp start is a shell command with multi line output. Assertion usually fails as we don't get the timely response and hence making the test flaky

tests/system/python/test_end_to_end_csv.py
Singhal-Vaibhav: It can be due to many reasons, PI server parameters not given correctly or pi web api is not working or the response is too slow. This is when retrieving data from pi web api. So if this is None, we return (None) and try again, till we reach max retries.

tests/system/python/test_end_to_end_csv.py
praveen-garg: Please add comment / docstring

tests/system/python/test_end_to_end_csv.py
```diff
@@ -0,0 +1,209 @@
...
+NUM_RETRIES = 3
+# Sleep interval between each tries
+SLEEP_INTERVAL = 10
+
+
+@pytest.fixture
+def start_south_north(reset_and_start_foglamp, south_plugin, asset_name, foglamp_url, pi_host, pi_port,
+                      north_plugin, pi_token):
+    """ This fixture clone a south repo and starts both south and north instance """
+    assert os.environ.get('FOGLAMP_ROOT') is not None
```
praveen-garg: Add newline

tests/system/python/test_end_to_end_csv.py
Singhal-Vaibhav: Its handled:
```    if data_from_pi is None or retry_count == NUM_RETRIES:
        assert False, "Failed to read data from PI"```

File: tests/system/python/test_end_to_end_csv.py [hunk 14]
```diff
@@ -0,0 +1,209 @@
...
+    r = conn.getresponse()
+    assert 200 == r.status
+    r.read().decode()
+
+    data = {"schedule_name": "{}".format("North_Readings_to_PI")}
+    conn.request("PUT", '/foglamp/schedule/enable', json.dumps(data))
+    r = conn.getresponse()
+    assert 200 == r.status
+    r.read().decode()
+    yield start_south_north
```
praveen-garg: why yield, please leave a comment

tests/system/python/test_end_to_end_csv.py
```diff
@@ -0,0 +1,209 @@
...
+    retry_count = 0
+    data_from_pi = None
+    while data_from_pi is None and retry_count < NUM_RETRIES:
+        data_from_pi = _read_data_from_pi(pi_host, pi_port, pi_admin, pi_passwd, pi_db, asset_name)
+        retry_count += 1
+        time.sleep(SLEEP_INTERVAL)
+
+    if data_from_pi is None or retry_count == NUM_RETRIES:
+        assert False, "Failed to read data from PI"
+    assert data_from_pi == str(CSV_DATA)
```
praveen-garg: Add a newline at EOF

tests/system/python/test_end_to_end_csv.py
```diff
@@ -0,0 +1,209 @@
...
+# Number of tries to make to read PI data and to read North configuration
+NUM_RETRIES = 3
+# Sleep interval between each tries
+SLEEP_INTERVAL = 10
+
+
+@pytest.fixture
+def start_south_north(reset_and_start_foglamp, south_plugin, asset_name, foglamp_url, pi_host, pi_port,
+                      north_plugin, pi_token):
+    """ This fixture clone a south repo and starts both south and north instance """
```
praveen-garg: Please add newline after each docstring

tests/system/python/test_end_to_end_csv.py
```diff
@@ -0,0 +1,209 @@
...
+__author__ = "Vaibhav Singhal"
+__copyright__ = "Copyright (c) 2018 Dianomic Systems"
+__license__ = "Apache 2.0"
+__version__ = "${VERSION}"
+
+
+CSV_NAME = "sample.csv"
+CSV_HEADER = "temperature"
+CSV_DATA = 2.5
+SENSOR_NAME = CSV_HEADER
```
praveen-garg: why its a copy of `CSV_HEADER`

tests/system/python/test_end_to_end_csv.py
```diff
@@ -0,0 +1,243 @@
...
+                      north_plugin, pi_token):
+    """ This fixture clone a south repo and starts both south and north instance """
+
+    # Test that FOGLAMP_ROOT environment variable is set
+    assert os.environ.get('FOGLAMP_ROOT') is not None
+
+    # Start foglamp south service
+    csv_file_path = _start_folamp_south(south_plugin, asset_name, foglamp_url)
+
+    # Start foglamp north service
```
praveen-garg: service?

tests/system/python/conftest.py
```diff
@@ -0,0 +1,101 @@
...
+    assert os.environ.get('FOGLAMP_ROOT') is not None
+    subprocess.run(["$FOGLAMP_ROOT/scripts/foglamp kill"], shell=True, check=True)
+    subprocess.run(["echo YES | $FOGLAMP_ROOT/scripts/foglamp reset"], shell=True, check=True)
+    subprocess.run(["$FOGLAMP_ROOT/scripts/foglamp start"], shell=True)
+    stat = subprocess.run(["$FOGLAMP_ROOT/scripts/foglamp status"], shell=True, stdout=subprocess.PIPE)
+    assert "FogLAMP not running." not in stat.stdout.decode("utf-8")
+
+
+def pytest_addoption(parser):
+    parser.addoption("--foglamp_url", action="store", default="localhost:8081",
```
ashish-jabble: action="store" means?

tests/system/python/test_end_to_end_csv.py
```diff
@@ -0,0 +1,243 @@
...
+        return None
+
+
+@pytest.fixture
+def start_south_north(reset_and_start_foglamp, south_plugin, asset_name, foglamp_url, pi_host, pi_port,
+                      north_plugin, pi_token):
+    """ This fixture clone a south repo and starts both south and north instance """
+
+    # Test that FOGLAMP_ROOT environment variable is set
+    assert os.environ.get('FOGLAMP_ROOT') is not None
```
ashish-jabble: Not needed as its already in reset_and_start_foglamp fixture

tests/system/python/test_end_to_end_csv.py
```diff
@@ -0,0 +1,243 @@
...
+    conn.request("POST", '/foglamp/service', json.dumps(data))
+    r = conn.getresponse()
+    assert 200 == r.status
+    r = r.read().decode()
+    retval = json.loads(r)
+    assert retval["name"] == "play"
+    return csv_file_path
+
+
+def _start_folamp_north(foglamp_url, pi_host, pi_port, north_plugin, pi_token):
```
ashish-jabble: _start_foglamp_north

tests/system/python/test_end_to_end_csv.py
```diff
@@ -0,0 +1,243 @@
...
+
+# Number of tries to make to read PI data and to read North configuration
+NUM_RETRIES = 3
+# Low sleep interval between each tries, used in internal foglamp process waits
+SLEEP_INTERVAL_LOW = 3
+# High sleep interval between each tries, used in external pi webapi waits
+SLEEP_INTERVAL_HIGH = 10
+
+
+def _start_folamp_south(south_plugin, asset_name, foglamp_url):
```
ashish-jabble: _start_foglamp_south

tests/system/python/test_end_to_end_csv.py
```diff
@@ -0,0 +1,243 @@
...
+@pytest.fixture
+def start_south_north(reset_and_start_foglamp, south_plugin, asset_name, foglamp_url, pi_host, pi_port,
+                      north_plugin, pi_token):
+    """ This fixture clone a south repo and starts both south and north instance """
+
+    # Test that FOGLAMP_ROOT environment variable is set
+    assert os.environ.get('FOGLAMP_ROOT') is not None
+
+    # Start foglamp south service
+    csv_file_path = _start_folamp_south(south_plugin, asset_name, foglamp_url)
```
ashish-jabble: _start_foglamp_south

tests/system/python/test_end_to_end_csv.py
```diff
@@ -0,0 +1,243 @@
...
+    """Start south service"""
+
+    south_config = {"assetName": {"value": "{}".format(asset_name)}, "csvFilename": {"value": "{}".format(CSV_NAME)},
+                    "ingestMode": {"value": "batch"}}
+    data = {"name": "play", "type": "South", "plugin": "{}".format(south_plugin), "enabled": "true",
+            "config": south_config}
+
+    conn = http.client.HTTPConnection(foglamp_url)
+    subprocess.run(["rm -rf /tmp/foglamp-south-{}".format(south_plugin)], shell=True, check=True)
+    subprocess.run(["rm -rf -d ${FOGLAMP_ROOT}/python/foglamp/plugins/south/*/"], shell=True, check=True)
```
ashish-jabble: Don't do "*",  instead `${FOGLAMP_ROOT}/python/foglamp/plugins/south/{south_plugin}

tests/system/python/test_end_to_end_csv.py
```diff
@@ -0,0 +1,243 @@
...
+    conn.request("PUT", '/foglamp/category/North_Readings_to_PI', json.dumps(data))
+    r = conn.getresponse()
+    assert 200 == r.status
+    r.read().decode()
+
+    data = {"schedule_name": "{}".format("North_Readings_to_PI")}
+    conn.request("PUT", '/foglamp/schedule/enable', json.dumps(data))
+    r = conn.getresponse()
+    assert 200 == r.status
+    r.read().decode()
```
ashish-jabble: We can do enable task with good config in **2 steps**:

1) Fetch north plugins with their default config and get north_plugin key from the JSON => 

> http://localhost:8081/foglamp/plugins/installed?type=north&config=true

2) Update the config from point1 and  then add north task with updated config via below
> curl -sX POST http://localhost:8081/foglamp/scheduled/task -d
     '{"name": "PI-2",
     "plugin": "pi_server",
     "type": "north",
     "schedule_type": 3,
     "schedule_day": 0,
     "schedule_time": 0,
     "schedule_repeat": 30,
     "schedule_enabled": true,
     "config": {
        "producerToken": {"value": "uid=180905062754237&sig=kx5l+"},
        "URL": {"value": "https://10.2.5.22:5460/ingress/messages"}}}'
    """

tests/system/python/conftest.py
Singhal-Vaibhav: store the value of arg value passed in variable

sstefano69:
@Singhal-Vaibhav 

I tried having a cleaned PI-Server and I have got an error
it worked fine at the second attempt, having the data already in the PI-Server.

I changed to :
	NUM_RETRIES = 60

and it worked fine, I propose to extend the value.

MarkRiddoch:
Whilst this test looks like it does the right thing and does test the system, we now have a slight problem.

The original request was to write a system test for a particular circumstance, not to create a new test framework for running system tests. Undoubtedly the new framework works, however we now have two system test frameworks and two sets of tests, one for each framework. This means we have to do one of the following:

- Rewrite the existing system test in pytest

- Rewrite this test in the existing framework

- Support both frameworks going forwards

What we need is more tests, not more test frameworks. Ideally a test framework should make it easy to write tests, that is the most important factor when we want to maintain good test coverage. The current, bash script based system test was intended to be modular and allow reuse of test components to make that process easy. It is not clear if switching to pytest is going to exacerbate that goal or make it easier for all developers, including outside plugin developers, to create new system test cases and increase our coverage.

I suspect, in the interests of increased coverage we are going to have to accept this pytest variant of a system test in the short term, but we should make a clear decision before adopting it as the way to do all systems tests going forwards. We clearly **should not** introduce any new test frameworks and instead concentrate on expanding the test coverage.

Singhal-Vaibhav:
@sstefano69 I will tune the max retries once we have this setup running on CI.

PR #1305, opened 2018-12-28
FOGL-2134 : Adding and removing filter pipelines from services should be dynamic

AmandeepArora:
Filter pipeline abstraction done and added support for dynamic reconfiguration of filter pipeline.

This branch cannot be merged yet into develop since corresponding changes in C++ sending process are also required, otherwise sending process won't compile.


MarkRiddoch:
> C++ sending process won't compile with the changes in this PR. It needs to use the new/updated filter pipeline abstraction & related classes to make it work. There is JIRA ticket FOGL-2248 for this work. So probably after that is also done, we can merge both into develop.

Thanks Aman. We need to be careful about the ordering of the checkin's, or at least make sure both go in together to avoid breaking develop.

AmandeepArora:
@MarkRiddoch - Asset tracker related code came in when FOGL-2110 was merged into this branch.

PR #1307, opened 2019-01-02
Category creation ordering altered for new pipeline creation

amarendra-dianomic:


python/foglamp/services/core/api/filters.py
```diff
@@ -243,8 +243,8 @@
         else:  # No existing filters, hence create new item 'config_item' and add the "pipeline" array as a string
             new_item = dict({config_item: {'description': 'Filter pipeline', 'type': 'JSON', 'default': {}}})
             new_item[config_item]['default'] = json.dumps({'pipeline': filter_list})
-            await cf_mgr.create_category(category_name=user_name, category_value=new_item, keep_original_items=True)
             await _add_child_filters(storage, cf_mgr, user_name, filter_list)
+            await cf_mgr.create_category(category_name=user_name, category_value=new_item, keep_original_items=True)
```
praveen-garg: As callback triggers as soon as category is created, we need to define / add children items first!

PR #1315, opened 2019-01-08
FOGL-2260 - Add log message to Python south service for lack of filter support

amarendra-dianomic:


python/foglamp/services/south/ingest.py
```diff
@@ -170,6 +170,10 @@ class Ingest(object):
         })
         cls._parent_service._core_microservice_management_client.create_configuration_category(config_payload)
 
+        # Check and warn if pipeline exists in South service
+        if 'filter' in cls._parent_service.config:
+            _LOGGER.warning('Filter pipeline is not supported on Python South service [%s]', cls._parent_service.config['plugin']['value'])
```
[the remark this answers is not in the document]
amarendra-dianomic: done

praveen-garg:
> @praveen-garg You will need to a) add a South service e.g. sinusoid, b) add a filter, and c) restart service/server if filter was added in service enabled mode. This is because this check is part of Ingest initialization process and which happens only at the start of the service. If we add this check in add_readings() method, it will flood syslog with unnecessary messages.
> 
> `Jan 8 13:51:08 asinha-ThinkPad-E450 FogLAMP[9141] WARNING: ingest: foglamp.services.south.ingest: South Service [sinusoid] does not support the use of a filter pipeline.`

then we should have it in python.foglamp.services.south.server.Server. `change` callback if newConfig should have filter config item.

amarendra-dianomic:
@praveen-garg, @MarkRiddoch The advantage of having this check in ingest is that this warning will appear each time the service is started, whereas in change callback, it will appear only once, at the time of adding a filter. Please let me know which one will be beneficial.

praveen-garg:
> @praveen-garg, @MarkRiddoch The advantage of having this check in ingest is that this warning will appear each time the service is started, whereas in change callback, it will appear only once, at the time of adding a filter. Please let me know which one will be beneficial.

@amarendra-dianomic  User should not need to disable / enable *restart* the south service to see this log! `filter` pipeline config item gets merged to south service category hence change callback is the correct place to listen that change and log, that its not supported.

amarendra-dianomic:
>@amarendra-dianomic User should not need to disable / enable restart the south service to see this log! filter pipeline config item gets merged to south service category hence change callback is the correct place to listen that change and log, that its not supported.
@praveen-garg if informing user is the intent, then it will be much better to consider this as a validation error and throw error/exception at the time of adding filter itself, rather than making a post facto log notification. Also, because it is highly unlikely that user will refer to log after a successful config change. So better to either a) add this in ticket's scope or b) create another ticket should we decide to do it later and must have clarity on one time log or log on each start.

praveen-garg:
@amarendra-dianomic  let's not make it complex. these both modules/ classes will be dead code, once we are done with FOGL-2271. Only scope of this ticket is to log on filter pipeline addition.

python/foglamp/services/south/ingest.py
```diff
@@ -170,6 +170,10 @@ class Ingest(object):
         })
         cls._parent_service._core_microservice_management_client.create_configuration_category(config_payload)
 
+        # Check and warn if pipeline exists in South service
+        if 'filter' in cls._parent_service.config:
+            _LOGGER.warning('South Service [%s] does not support the use of a filter pipeline.', cls._parent_service.config['plugin']['value'])
```
praveen-garg: it should be the service name not the plugin name

praveen-garg:
LGTM! both scenarios works perfectly fine.

python/foglamp/services/south/ingest.py
praveen-garg: Fixed!

PR #1326, opened 2019-01-10
FOGL-2301 Initial turning down of log verbosity

MarkRiddoch:
Initial toning down of log verbosity. More will no doubt follow as part of this Jira ticket.


MarkRiddoch:
@AmandeepArora Yes, agreed. I think having some configuration to allow this to be set would be good at some point.

MarkRiddoch:
Yes, @AmandeepArora Before applying this to sending process we need to do something to allow for advanced configuration options in the sending process. There are already a log of options and I don't want to crowd them further.

PR #1331, opened 2019-01-15
FOGL-2308 Fix reading of Python timestamp in different timezones

MarkRiddoch:


C/common/reading.cpp
```diff
@@ -283,3 +267,70 @@ ostringstream assetTime;
...
+		int h, m;
+		int sign = (*ptr == '+' ? -1 : +1);
+		ptr++;
+		sscanf(ptr, "%02d:%02d", &h, &m);
+		ts->tv_sec += sign * ((3600 * h) + (60 * m));
+	}
+	// Now convert to the local time
+	extern long timezone;
+	tzset();
+	ts->tv_sec -= timezone;
```
AmandeepArora: Will this be subtraction or addition? I think local time is "timezone" seconds ahead of UTC.

MarkRiddoch: Yes, for timezones east of Greenwich timezone is a negative number and timezones west of Greenwich it is a positive number. 

```
foglamp@foglamp-18:~$ cat t.c
main()
{
	extern long timezone;
	tzset();
	printf("Timezone: %ld\n", timezone);
}
foglamp@foglamp-18:~$ export TZ=EET
foglamp@foglamp-18:~$ ./t
Timezone: -7200
foglamp@foglamp-18:~$ export TZ=CET
foglamp@foglamp-18:~$ ./t
Timezone: -3600
foglamp@foglamp-18:~$ export TZ=EST
foglamp@foglamp-18:~$ ./t
Timezone: 18000
foglamp@foglamp-18:~$ 
```

MarkRiddoch: Given the nature of this issue it would be good if you could try it in all your time zones.

AmandeepArora: Ok, got it. Thanks.

PR #1338, opened 2019-01-18
FOGL-2318: Handle plugin_init or plugin_reconfigure failures/exceptions

AmandeepArora:
In case, plugin_init or plugin_reconfigure fails (or raises exception) and it is not recoverable, then shutdown the south service. This change is applicable for both C and python south plugins.

Change in scheduler code is done by Amarendra.


C/services/south/south.cpp
```diff
@@ -211,7 +211,9 @@ void SouthService::start(string& coreAddress, unsigned short corePort)
 		m_config = m_mgtClient->getCategory(m_name);
 		if (!loadPlugin())
 		{
-			logger->fatal("Failed to load south plugin.");
+			logger->fatal("Failed to load south plugin, exiting...");
+			bool rv = m_mgtClient->unregisterService();
```
praveen-garg: I think we should not un-register, but shutdown. 

Otherwise service registry will have no entry for this, and schedule shall remain enabled. there must be some mechanism to *try* restart this service (which will actually try disable and enable schedule for this). 

/cc @MarkRiddoch

MarkRiddoch: We should at least try a restart, although is it likely to fail again.

AmandeepArora: I think plugin knows best how to try to recover from the failure, what else to try and how to space those retries. And if all those re-attempts fail, it would just return a NULL handle to indicate init/reconf failure. So I guess further retries from this code are probably not required.

Other point is what Praveen raised. Does it make sense to unregister service or just shutdown service (so that its entry remains in the service registry) for now and leave it possible for it to be restarted later (probably manually)?

MarkRiddoch: I suggest we do not unregister it, ultimately we would like to ave it turn up as failed, if it is not registered then it can not be shown as failed. So go down the route of shutdown with unregister, we may have to look at what happens with retries of failed services, but there is already a Jira for this I believe.

PR #1341, opened 2019-01-21
FOGL-2304  Config item script upload and change callback fix

praveen-garg:


pintomax:
Currently testing:

- South Service with C++ plugin "random" and Python35 filter: calling shutdown is ok
- South Service with Python plugin "sinusoid" and Python35 filter: calling shutdown is ok as well

PR #1350, opened 2019-01-28
FOGL-2342 - Add task status into north api response

amarendra-dianomic:
Now hitting http://localhost:8081/foglamp/north gives:
```
[
	{
		"exclusive": true,
		"id": "844135a5-7dd5-414a-9b2d-d3803080e639",
		"processName": "north_c",
		"sent": -1,
		"repeat": 30,
		"name": "Pi",
		"enabled": true,
		"day": null,
		"taskStatus": {
			"endTime": "",
			"reason": "",
			"exitCode": "",
			"state": "Running",
			"startTime": "2019-01-28 16:53:34.612"
		}
	}
]
``` 

It may be noted that correct status code could be printed only after completion of [FOGL-2343](https://scaledb.atlassian.net/browse/FOGL-2343) which will take care of C++ sending process. But we need to make similar changes to Python sending process also as Python sending process is still in use.

Secondly, we may need to define additional status code for tasks table to account for misc failure codes as envisaged in [FOGL-2136](https://scaledb.atlassian.net/browse/FOGL-2136). This may require further changes in scheduler->entities.py.


python/foglamp/services/core/api/north.py
```diff
@@ -48,16 +70,25 @@
...
                 'day': sch.day,
                 'enabled': sch.enabled,
-                'exclusive': sch.exclusive
+                'exclusive': sch.exclusive,
+                'taskStatus': None if task is None else {
+                    'state': [t.name.capitalize() for t in list(Task.State)][int(task['state']) - 1],
+                    'startTime': str(task['start_time']),
+                    'endTime': str(task['end_time']),
+                    'exitCode': task['exit_code'],
+                    'reason': task['reason'],
```
ashish-jabble: exit_code and reason keys for debug?

praveen-garg: There is no change in `tasks` table  for reason and exit code yet so far, but once those are set by north egress, REST API will return them as is.

PR #1358, opened 2019-01-30
FOGL-2380 correct child relationship for filters category

MarkRiddoch:


MarkRiddoch:
This does no harm, and solves the initial problem. Will merge and investigate the multiple parents in a different ticket

PR #1364, opened 2019-02-04
FOGL-2407 async_ingest pymodule lib removed on make clean

ashish-jabble:


.gitignore
```diff
@@ -17,10 +17,12 @@ data/tmp
 
 # SQLite3 default db location and after migration
 data/*.db
+data/foglamp.db-journal
```
praveen-garg: As sqlite file name can set via DEFAULT_SQLITE_DB_FILE, hence please make it `data/*.db-journal`

PR #1370, opened 2019-02-04
FOGL-2422: Always use timestamp with timezone when inserting/updating…

pintomax:
… foglamp.tasks table

FOGL-2422: Always use timestamp with timezone when inserting/updating foglamp.tasks table


python/foglamp/services/core/scheduler/scheduler.py
```diff
@@ -28,6 +28,7 @@
 from foglamp.services.core.service_registry.service_registry import ServiceRegistry
 from foglamp.services.core.service_registry import exceptions as service_registry_exceptions
 from foglamp.services.common import utils
+from foglamp.plugins.common import utils
```
praveen-garg: utils as?

PR #1374, opened 2019-02-05
FOGL-2399: Add support for IN operations in the storage API W.I.P

pintomax:
FOGL-2399: Add support for IN operations in the storage API W.I.P

Addition of IN and NOT IN condition

First commit is for SQLite only


C/plugins/storage/sqlite/connection.cpp
```diff
@@ -2775,6 +2775,66 @@ bool Connection::jsonWhereClause(const Value& whereClause,
 		else
 			sql.append(" seconds')"); // Get value in UTC by asking for no timezone
 	}
+	else if (!cond.compare("in") || !cond.compare("not in"))
+	{
+		// Check we have a non empty array
+		if (whereClause["value"].IsArray() &&
+		    whereClause["value"].Size())
+		{
+			for (auto & c: cond) c = toupper(c);
```
MarkRiddoch: Not really required.

C/plugins/storage/sqlite/connection.cpp
```diff
@@ -2775,6 +2775,66 @@ bool Connection::jsonWhereClause(const Value& whereClause,
...
+					raiseError("where clause", message.c_str());
+					return false;
+				}
+			}
+			sql.append(" )");
+		}
+		else
+		{
+			string message("The \"value\" of an \"" + \
+					cond + "\" condition must be an array.");
```
MarkRiddoch: Since you also check Size() then I would include in the error "and must not be empty" or ave a separate error when the array is empty.

PR #1375, opened 2019-02-05
FOGL-2426: use timedateUtils alias

pintomax:
FOGL-2426: use timedateUtils alias


python/foglamp/services/core/scheduler/scheduler.py
```diff
@@ -28,7 +28,7 @@
 from foglamp.services.core.service_registry.service_registry import ServiceRegistry
 from foglamp.services.core.service_registry import exceptions as service_registry_exceptions
 from foglamp.services.common import utils
-from foglamp.plugins.common import utils
+from foglamp.plugins.common import utils as timedateUtils
```
praveen-garg: this is too ugly to see plugins refernce here! @MarkRiddoch @pintomax /cc @amarendra-dianomic

MarkRiddoch: @pintomax Please raise a ticket to migrate the localtimestamp call to common. Make it low priority and merge what we have now. The priority is to restore the functionality ASAP.

pintomax: We can migrate the utils code for the local timestamp to common code with another ticket

PR #1376, opened 2019-02-05
[FOGL-2425] test script make jobs defaults to -j 4; use -j 2 or -j 1 as needed in quotes or without space

praveen-garg:


praveen-garg:
@Singhal-Vaibhav  the PR tester for this is running on Ubuntu 16?

PR #1386, opened 2019-02-07
FOGL-2393 Include displayName in default config category (for 1.5.0RC)

MarkRiddoch:
Prevent display names from disappearing on first execution of a south service


praveen-garg:
should this be rebased with 1.5.0RC itself? Its carrying FOGL-2384 changes as well (which is okay, if planned).

MarkRiddoch:
Removed population caused by merging develop into this branch

PR #1395, opened 2019-02-11
FOGL-2451 Make no data to purge a log message and not an error return

MarkRiddoch:
from the service

Note, the error message about lack of support for purge by size remains as this is an error from the plugin and is the only way this information can be communicated back to the caller.

### DIFF
--- a/C/plugins/storage/sqlite/connection.cpp
+++ b/C/plugins/storage/sqlite/connection.cpp
@@ -2202,8 +2202,8 @@ int blocks = 0;
 
 	result = "{ \"removed\" : 0, ";
 	result += " \"unsentPurged\" : 0, ";
-	result += " \"unsentRetained\", 0, ";
-    	result += " \"readings\" 0 }";
+	result += " \"unsentRetained\" : 0, ";
+    	result += " \"readings\" : 0 }";
 
 
 	logger->info("Purge starting...");
@@ -2292,7 +2292,7 @@ int blocks = 0;
 
 		if (rowidLimit == 0)
 		{
- 			raiseError("purge", "No data to purge");
+ 			logger->info("No data to purge");
 			return 0;
 		}
 


C/plugins/storage/sqlite/connection.cpp
```diff
@@ -2202,8 +2202,8 @@ int blocks = 0;
 
 	result = "{ \"removed\" : 0, ";
 	result += " \"unsentPurged\" : 0, ";
-	result += " \"unsentRetained\", 0, ";
-    	result += " \"readings\" 0 }";
+	result += " \"unsentRetained\" : 0, ";
+    	result += " \"readings\" : 0 }";
```
praveen-garg: I'm surprised, we could never catch this. I'm sure json parser should have failed with this.

C/plugins/storage/sqlite/connection.cpp
```diff
@@ -2292,7 +2292,7 @@ int blocks = 0;
 
 		if (rowidLimit == 0)
 		{
- 			raiseError("purge", "No data to purge");
+ 			logger->info("No data to purge");
```
praveen-garg: Postgres and in memory plugins handles it already?

C/plugins/storage/sqlite/connection.cpp
praveen-garg: https://github.com/foglamp/FogLAMP/pull/1366/commits/e666cf799164bc6cbeeddc5f1be8e1f935f33bf5

Looks like we were assigning new (string) values on actual purge. and when we had no rows to purge, there was bare exception raised by plugin,  hence parser purge process parser never read these default.

C/plugins/storage/sqlite/connection.cpp
MarkRiddoch: They do not have this code that shortcuts if no rows to purge, therefore it is not needed.

praveen-garg: Okay

PR #1396, opened 2019-02-11
FOGL-2453 bulk update for previous value in stats history task run

ashish-jabble:


praveen-garg:
It would be great to record what performance gain we see here with SQLite. And what we are expecting / actually gain after FOGL-1973.

PR #1405, opened 2019-02-18
FOGL-2432 - Fix the utils import code in scheduler.py

amarendra-dianomic:


python/foglamp/services/core/api/notification.py
```diff
@@ -9,7 +9,7 @@
 import aiohttp
 from aiohttp import web
 
-from foglamp.common import utils
+from foglamp.common import utils as common_utils
```
praveen-garg: No changes required in this module.

python/foglamp/services/core/api/service.py
```diff
@@ -10,7 +10,7 @@
 from aiohttp import web
 
 from typing import Dict
-from foglamp.common import utils
+from foglamp.common import utils as common_utils
```
praveen-garg: No changes required in this module.

python/foglamp/services/core/api/task.py
```diff
@@ -6,10 +6,9 @@
 
 import datetime
 import uuid
-
 from aiohttp import web
 
-from foglamp.common import utils
+from foglamp.common import utils as common_utils
```
praveen-garg: No changes required in this module.

python/foglamp/services/core/scheduler/scheduler.py
```diff
@@ -17,18 +17,19 @@
...
-from foglamp.services.core.scheduler.exceptions import *
 from foglamp.common.storage_client.exceptions import *
 from foglamp.common.storage_client.payload_builder import PayloadBuilder
 from foglamp.common.storage_client.storage_client import StorageClientAsync
+from foglamp.common.configuration_manager import ConfigurationManager
+from foglamp.services.core.scheduler.entities import *
+from foglamp.services.core.scheduler.exceptions import *
 from foglamp.services.core.service_registry.service_registry import ServiceRegistry
 from foglamp.services.core.service_registry import exceptions as service_registry_exceptions
 from foglamp.services.common import utils
```
praveen-garg: remove this import as as you already imported now

python/foglamp/tasks/statistics/statistics_history.py
```diff
@@ -15,7 +15,7 @@
 from foglamp.common.storage_client.payload_builder import PayloadBuilder
 from foglamp.common import logger
 from foglamp.common.process import FoglampProcess
-from foglamp.plugins.common import utils
+from foglamp.common import utils as common_utils
```
praveen-garg: No changes required in this module.

python/foglamp/services/core/scheduler/scheduler.py
amarendra-dianomic: Nope. This is **foglamp.services.common.utils** and the already imported is **foglamp.common.utils**!

amarendra-dianomic:
@praveen-garg Re your comments "No changes required in this module.", pl note that these changes are meant to mere make the code more readable. They do not have any impact on functionality, whatsoever!

praveen-garg:
yeah, let's keep the stuff within scope and avoid introducing regression risk and need to exhaustive testing. The way these exists today also readable and clearly explicit.

python/foglamp/services/core/scheduler/scheduler.py
praveen-garg: was not it better, if we have added  `local_timestamp` def to services/common/utils itself (as requested in JIRA), as this module also belongs to services package services/core/scheduler/scheduler.py and avoid extra import

amarendra-dianomic: FOGL-2432 (for services) says that fix for this ticket fixes FOGL-2433 (for plugins) also. Since services and plugins are two separate folders at the same level in hierarchy, therefore it makes sense that the common fix to both should come above of both these folders and therefore foglamp.common.utils is the ideal home for common methods.

amarendra-dianomic: Also, scheduler starts (and imports) only once during the life time of a foglamp server instance. Hence this additional import cost is minuscule and affordable.

python/foglamp/common/utils.py
```diff
@@ -34,3 +35,22 @@ def check_reserved(string):
         if s in reserved:
             return False
     return True
+
+
+def get_diff(old, new):
```
praveen-garg: We don;t need it anywhere

python/foglamp/services/core/scheduler/scheduler.py
praveen-garg: No we don;t want to change plugins import for plugins for now. It will be removed later, but as along as it stay it will be from .plugins path only

praveen-garg: python/foglamp/tasks/statistics/statistics_history.py also use the plugin path. We should change that with test file (2 more file changes only, from initial commit) so that `from foglamp.common import utils as common_utils` import becomes more reasonable;

tests/unit/python/foglamp/common/test_common_utils.py
```diff
@@ -0,0 +1,44 @@
...
+        actual = common_utils.check_reserved(test_string)
+        assert expected == actual
+
+    @pytest.mark.parametrize("test_input_old, test_input_new, expected", [
+        ({'a': 1, 'b': 2, 'c': 3}, {'a': 11, 'b': 22, 'c': 33}, ['a', 'b', 'c']),
+        ({'a': 1, 'b': 2, 'c': 3}, {'a': 11, 'b': 22, 'd': 44}, ['a', 'b', 'd']),
+        ({'a': 1, 'b': 2, 'c': 3}, {'a': 11, 'b': 22}, ['a', 'b']),
+        ({'a': 1, 'b': 2, 'c': 3}, {'d': 11, 'e': 22, 'c': 33}, ['d', 'e', 'c'])
+    ])
+    def test_get_diff(self, test_input_old, test_input_new, expected):
```
praveen-garg: We  are not using this until now from actual code, which is introduced in this PR only. So let's remove that and this test as well.

python/foglamp/services/core/scheduler/scheduler.py
amarendra-dianomic: done

tests/unit/python/foglamp/common/test_common_utils.py
amarendra-dianomic: done

PR #1409, opened 2019-02-18
FOGL-2501 : fixed lib reference in cmake

sstefano69:


AmandeepArora:
@sstefano69 , the shared libraries that are linked in by unit test cases are created within tests/unit/C/lib. So it doesn't need to refer the libraries from cmake_build/C/lib.

sstefano69:
@AmandeepArora 
we discovered that the lib dir is created executing  :  

/home/foglamp/Development/FogLAMP/tests/unit/C/scripts/RunAllTests.sh

so it is required before executing standalone unit tests,

PR #1410, opened 2019-02-19
FOGL-2467 : tmp

sstefano69:


sstefano69:
opened maybe  for error.

PR #1415, opened 2019-02-20
FOGL-2493

pintomax:
FOGL-2493: use statistics instead of readings in /foglamp/south API


pintomax:
The GUI now shows the total number of ingested readings, not the buffered ones

pintomax:
NOTE:
this is change of functionality,: the GUI will report total ingested readings instead of currently buffered ones.

MarkRiddoch (approved):
Happy with this in principle, will make a big improvement in performance. The only concern is this is a chance in functionality, in particular the GUI displays different data and should be run past @billehunt

praveen-garg (approved):

PR #1422, opened 2019-02-22
FOGL-2511-B - Fix shim.py->plugin_init() to retain JSON type items as str and not dict

amarendra-dianomic:


python/foglamp/plugins/common/shim/shim.py
```diff
@@ -43,7 +43,18 @@ def plugin_info():
 def plugin_init(config):
     _LOGGER.info("plugin_init called")
     handle = _plugin.plugin_init(json.loads(config))
-    return handle
+    # South C server sends "config" argument as string in which all JSON type items' components,
+    # 'default' and 'value', gets converted to dict during json.loads(). Hence we need to restore
+    # them to str, which is the required format for configuration items.
+    revised_handle = {}
+    for k, v in handle.items():
+        if v['type'] == 'JSON':
```
praveen-garg: Looks okay. We should add a `# TODO: or  FIXME: with FOGL-1827`

PR #1423, opened 2019-02-22
FOGL-2435 (patch) -y added to apt install

ashish-jabble:


requirements.sh
```diff
@@ -23,9 +23,9 @@
 
 set -e
```
praveen-garg:
```
sudo apt update
sudo apt -y upgrade
```

praveen-garg: https://github.com/foglamp/FogLAMP/pull/1426

PR #1424, opened 2019-02-22
FOGL-2490: limit the number of worker threads in all reading wrappers

pintomax:
FOGL-2490: code cleanup around reading wrappers


pintomax:
All reading wrappers now deal with threads number limit

AmandeepArora:
Can you briefly explain the original issue that you are trying to address here and what results you got with these changes? The multiple threads' contention is also indirectly handled by the number of DB connections. I haven't seen them grow beyond 4-5. And with WAL mode, DB read and writes can happen in parallel in many cases. Multiple readers can work in parallel. The issue is when there is more than 1 writer or too many readers on a table starving a writing thread. And even in this case of DB lock contention, the other contending threads can prepare the SQL queries by parsing JSON payloads, e.g. case of 5K readings being inserted in a single request/transaction or generating the JSON payload after reading back a readings' batch from DB.

pintomax:
ubuntu@ubuntu16-server-virt:~/source/develop/FogLAMP$ ab -n 1000 -c 3 127.0.0.1:8081/foglamp/asset/Random

Nothing logged

ubuntu@ubuntu16-server-virt:~/source/develop/FogLAMP$ ab -n 1000 -c 6 127.0.0.1:8081/foglamp/asset/Random


Feb 22 12:06:08 ubuntu16-server-virt FogLAMP Storage[19244]: WARNING: Storage API: readingQueryWrapper can not be run by a thread right now, maximum number of 5 allowed threads hit.

pintomax:
@AmandeepArora

Task description says:
- We should probably allow worker threads for all operations, currently it is limited to a small number.
   I added a new one readingsFetch
- ... put in place a mechanism, to limit the number of worker threads that are allowed at any one time.
 Hence the limit of 5 in this work.

The limit is hit only if we have many concurrent connections that spawns new threads.

Assuming we have one thread for each operation we might end up with 10 threads.
This can be too much for a small device.

AmandeepArora:
It depends on what the new incoming thread is going to do. It doesn't matter (mostly) if a new reader thread is coming in. And for writer threads that are trying to write to readings table, probably better interleaving can be done in the Connection class methods since DB access is just part of the task to be done by the incoming thread. Other part is preparing the transaction payload from JSON payload and vice versa. Also DB level arbitration can be better done in the Connection class code, rather than not allowing a new incoming thread completely. Multiple writers is an issue, not multiple readers.

In your example, "readingQueryWrapper" thread was not allowed even though it was not going to add anything to DB access contention in sqlite WAL mode.

For handling contention between writer threads to readings table, I am making changes in FOGL-2489-1 branch.

C/services/storage/storage_api.cpp
```diff
@@ -112,25 +115,80 @@ void on_error(__attribute__((unused)) shared_ptr<HttpServer::Request> request, _
...
-void readingAppendWrapper(shared_ptr<HttpServer::Response> response, shared_ptr<HttpServer::Request> request)
+void readingAppendWrapper(shared_ptr<HttpServer::Response> response,
+			  shared_ptr<HttpServer::Request> request)
 {
 	StorageApi *api = StorageApi::getInstance();
+#if WORKER_THREADS
+	std::atomic<int>* cnt = &(api->m_workers_count);
+	// Check rurrent number of workers and increase count
+	if (std::atomic_load(cnt) < MAX_WORKER_THREADS &&
+	    std::atomic_fetch_add(cnt, 1) < MAX_WORKER_THREADS)
```
AmandeepArora: In this case, if MAX_WORKER_THREADS  is 5 and cnt is 4, then first part of the '&&' condition is true but second half is false and we don't allow this new thread. But we have already increased cnt to 5. Is that expected behaviour?

C/services/storage/storage_api.cpp
```diff
@@ -112,25 +115,80 @@ void on_error(__attribute__((unused)) shared_ptr<HttpServer::Request> request, _
...
+#else
 	api->readingAppend(response, request);
+#endif
 }
 
 /**
  * Wrapper function for the reading fetch API call.
  */
-void readingFetchWrapper(shared_ptr<HttpServer::Response> response, shared_ptr<HttpServer::Request> request)
+void readingFetchWrapper(shared_ptr<HttpServer::Response> response,
```
AmandeepArora: Do we want to count reading thread towards the worker thread limit?

C/services/storage/storage_api.cpp
```diff
@@ -112,25 +115,80 @@ void on_error(__attribute__((unused)) shared_ptr<HttpServer::Request> request, _
...
+			 shared_ptr<HttpServer::Request> request)
 {
 	StorageApi *api = StorageApi::getInstance();
+#if WORKER_THREADS
+	std::atomic<int>* cnt = &(api->m_workers_count);
+	// Check rurrent number of workers and increase count
+	if (std::atomic_load(cnt) < MAX_WORKER_THREADS &&
+	    std::atomic_fetch_add(cnt, 1) < MAX_WORKER_THREADS)
```
AmandeepArora: Same comment.

C/services/storage/storage_api.cpp
pintomax: @AmandeepArora I'll remove the limit and just print the counter with next commits.

This change is only for [readingQueryWrapper] removal

pintomax:
@AmandeepArora @MarkRiddoch I've reverted changes for limiting the creation of a new thread.

The three methods that can be run by a thread now log a warning message if the current worker threads number is greater than 5, the hardcoded value in #define

With seven south services running I haven't seen the waring message.

With some debug loggings I've seen only two as maximum value.

C/services/storage/storage_api.cpp
```diff
@@ -112,25 +115,84 @@ void on_error(__attribute__((unused)) shared_ptr<HttpServer::Request> request, _
...
-void readingAppendWrapper(shared_ptr<HttpServer::Response> response, shared_ptr<HttpServer::Request> request)
+void readingAppendWrapper(shared_ptr<HttpServer::Response> response,
+			  shared_ptr<HttpServer::Request> request)
 {
 	StorageApi *api = StorageApi::getInstance();
+#if WORKER_THREADS
+	std::atomic<int>* cnt = &(api->m_workers_count);
+	// Check rurrent number of workers and log if threshold value is hit
+	int tVal = std::atomic_load(cnt);
+	if (tVal > MAX_WORKER_THREADS)
```
AmandeepArora: Should this check be ">=" rather than ">"?

PR #1428, opened 2019-02-26
FOGL-2521:investigate why the unit test /tests/unit/C/services/storage/postgres/tests.cpp behaves differently locally and on the automated test

sstefano69:


sstefano69:
activity not needed.

PR #1431, opened 2019-02-28
FOGL-2511 Minor fix in shim.py when custom fields are added to config (Fixes FOGL-2535 and FOGL-2536)

amarendra-dianomic:


python/foglamp/plugins/common/shim/shim.py
```diff
@@ -82,10 +82,11 @@ def _revised_config_for_json_item(config):
     # them to str, which is the required format for configuration items.
     revised_config_handle = {}
     for k, v in config.items():
-        if v['type'] == 'JSON':
-            if isinstance(v['default'], dict):
-                v['default'] = json.dumps(v['default'])
-            if isinstance(v['value'], dict):
-                v['value'] = json.dumps(v['value'])
+        if isinstance(v, dict):
```
praveen-garg: I believe this will also handle FOGL-2535 ; as fix is also applicable for re-configure

PR #1438, opened 2019-03-05
FOGL-2471 - Add certificate based authentication to FogLAMP API

amarendra-dianomic:
Steps to test:
1) sudo make clean. make. scripts/foglamp reset.

2) Start foglamp. On Configuration, select "authentication - mandatory". Save. Stop server. Restart.

3) Create a local CA:

        cd data/etc/certs
        openssl genrsa -out ca.key 4096
        openssl req -new -x509 -days 365 -key ca.key -out ca.cert -subj '/C=US/CN=MY-CA'

   **Must:** In "CN" value, it must NOT be "foglamp" as it is already taken by foglamp.cert. Should we decide to use foglamp.cert as CA, it will create conflicts later.

4) Create a user certificate:
   
         openssl genrsa -out user.key 1024
         openssl req -new -key user.key -out user.csr -subj '/C=IN/CN=user'

   **Must:** In "CN" value, provide an existing username. For testing, it can be "user" which is an existing user.

5) Sign user certificate with CA certificate:

        openssl x509 -req -days 365 -in user.csr -CA ca.cert -CAkey ca.key -set_serial 01 -out user.cert

6) Now we need to login to generate a token for an user session based on user certificate. Start postman. In a tab, select "POST http://localhost:8081/foglamp/login". In Body, select option "raw". Choose "Text" from dropdown. Now copy paste contents of user.cert in Body. Press "Send". Kind of below response appears:

  ```
    {
       "admin": false,
       "token": "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJleHAiOjE1NTE4MDI2NTksInVpZCI6Mn0.UcjDhR47Bj1MLNSrrW7tvdBqJUM2QNV-P8nY-ULinY8",
       "uid": 2,
       "message": "Logged in successfully"
    }
   ```

   With curl:

`curl -T data/etc/certs/user.cert -X POST http://localhost:8081/foglamp/login -k | jq
`

7) In another postman tab, select "GET http://localhost:8081/foglamp/schedule". In Headers, add "Authorization" with value of token above. Press "Send". A list of schedule appears.

   With curl:

`curl -H "Authorization:eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ1aWQiOjIsImV4cCI6MTU1MTgwNjkzN30.LtMNZZrKMJNCz8cL3DB84xlAFstfIKvfubdok_XKyFI" -X GET http://localhost:8081/foglamp/schedule -k | jq
`

8) Try step 6) with incorrect client certificate, message appears like,

```
401: Authentication failed
```

TODO (separately): 

1) GUI page to generate client certificate.
2) GUI login page to add client certificate contents.
3) GUI page to revoke client certificate including expiry
4) Mechanism to revoke client certificate including expiry

[FOGL-2557](https://scaledb.atlassian.net/browse/FOGL-2557) has been created for 1, 2, 3 above.


python/foglamp/services/core/user_model.py
```diff
@@ -347,6 +350,49 @@ class Objects:
 
             return uid, jwt_token, False
 
+        @classmethod
+        async def certificate_login(cls, username, host):
+            """
+            Args:
+                user_cert: user_cert
```
praveen-garg: No, this is not

python/foglamp/services/core/user_model.py
```diff
@@ -347,6 +350,49 @@ class Objects:
...
+            except StorageServerError as ex:
+                if ex.error["retryable"]:
+                    pass  # retry INSERT
+                raise ValueError(ERROR_MSG)
+
+            # TODO remove hard code role id to return is_admin info
+            if int(found_user['role_id']) == 1:
+                return uid, jwt_token, True
+
+            return uid, jwt_token, False
```
praveen-garg: looks like this is 80% duplicate of login def. Please put common code in _login def?

python/requirements.txt
```diff
@@ -7,3 +7,6 @@ pyjwt==1.6.4
 
 # Transformation of data, Apply JqFilter
 pyjq==2.2.0
+
+# openssl
+pyopenssl==19.0.0
```
praveen-garg: what is  the build size impact?

python/foglamp/common/web/tls_helper.py
```diff
@@ -0,0 +1,146 @@
...
+class VerificationError(ValueError):
+    pass
+
+
+class BaseCert:
+    @classmethod
+    def from_pem(cls, pem_data):
+        try:
+            if not isinstance(pem_data, str):
+                raise AssertionError("pem_data is not string: %s", pem_data)
```
praveen-garg: not sure `AssertionError` will be stripped? if we use python -O or -OO flag, as for the assert statement/s

python/foglamp/services/core/api/auth.py
```diff
@@ -62,35 +68,61 @@
...
-    if not username or not password:
-        _logger.warning("Username and password are required to login")
-        raise web.HTTPBadRequest(reason="Username or password is missing")
+        user_cert = data
+        try:
+            cert = BaseCert.from_pem(user_cert)
+        except ValueError:
+            cert = None
+        if not cert:
+            raise web.HTTPBadRequest(reason="Invalid Certificate")
```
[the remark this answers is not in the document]
praveen-garg: i.e. 
1. no need to have an extra variable user_crt = data as its not being reused.
2. cert is assigned default to None. that will handle the case even if an unknown error raised

python/foglamp/services/core/server.py
```diff
@@ -306,6 +316,24 @@ def get_certificates(cls):
...
+                cls.is_auth_required = True if config['authentication']['value'] == "mandatory" else False
+            except KeyError:
+                _logger.error("error in retrieving authentication info")
+                raise
+
+            try:
+                cls.auth_method = config['authMethod']['value']
+                if cls.is_auth_required:
+                    if cls.auth_method not in ['password', 'certificate']:
+                        raise ValueError("For authentication = mandatory, authMethod must be either password or certificate")
```
praveen-garg: if auth is required the  authMethod allows to restrict to use password or certificate or you can use any!

Hence this check is wrong. 
if any is the  auth_method (which is the default too) user should be able to login either by username + password Or certificate.

if auth_method == certificate: user can only use certificate to enter into the system
if auth_method == password: user can only use password to enter into the system

python/foglamp/services/core/server.py
```diff
@@ -314,6 +342,9 @@ def get_certificates(cls):
 
             try:
                 cls.is_rest_server_http_enabled = False if config['enableHttp']['value'] == 'false' else True
+                # https is a must if authentication method is certificate
+                if cls.auth_method == 'certificate':
+                    cls.is_rest_server_http_enabled = False
```
praveen-garg: why this is a must?

python/foglamp/services/core/user_model.py
```diff
@@ -347,6 +355,30 @@ class Objects:
...
             return uid, jwt_token, False
 
+        @classmethod
+        async def certificate_login(cls, username, host):
+            """
+            Args:
+                username: username
+                host:     IP address
+            Returns:
+                  return token
```
praveen-garg: not the token only

praveen-garg:
> Start foglamp. On Configuration, select "certificate" as authentication method. Save. Stop server. Restart. From this point onward, it will not be possible to test via GUI as GUI does not have option to "Login via certificate".

@amarendra-dianomic  The only thing we need to do is make authentication mandatory, as authMethod is default to set `any`
then user will be required to, either enter password or use certificate, for auth.

**Please update the description**

NOTE: However for testing, we should also try authMethod: password / certificate only settings as well.

python/foglamp/services/core/server.py
amarendra-dianomic: Because certificate cannot be allowed to be uploaded via insecure channel even if it is a public key.

python/foglamp/services/core/server.py
amarendra-dianomic: Basically "authMethod" is for GUI use only - to direct to appropriate page/action based upon its value. Backend will act per whatever method is choosen by GUI to login.

praveen-garg: No GUI, will provide both option. 

if auth is mandatory, GUI will throw the user on login screen. and allow to choose any. if server allows only password or certificate only then user will be informed.

amarendra-dianomic: Yes, same intent was meant by my comment. It is GUI which will decide how to login based upon "authentication" and "authMethod".

praveen-garg:
1) what certificate name config item is there for cert based authentication? the current certificateName config item is for HTTPS. 

2) shall more than 1 user shall be able to login using same certificate, if shared?

3) GUI has a basic certificates store page. which allows to upload / delete / overwrite the certificates. in HTTP mode only so that foglamp can be started to be in https with desired / default certificate.

Please elaborate more why, commit a7bb8bc, it is a must have to be on https to allow auth with certificate.

amarendra-dianomic:
Commit a7bb8bc was required because of this error:
```
FogLAMP (FOGL-2471 %) $ curl -T data/etc/certs/user.crt -X POST https://localhost:8081/foglamp/login -k | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (35) gnutls_handshake() failed: An unexpected TLS packet was received.

```

praveen-garg:
> Commit [a7bb8bc](https://github.com/foglamp/FogLAMP/commit/a7bb8bc1c4c3415f3dde58fd22ca044a717bb2df) was required because of this error:
> 
> ```
> FogLAMP (FOGL-2471 %) $ curl -T data/etc/certs/user.crt -X POST https://localhost:8081/foglamp/login -k | jq
>   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
>                                  Dload  Upload   Total   Spent    Left  Speed
>   0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
> curl: (35) gnutls_handshake() failed: An unexpected TLS packet was received.
> ```

Looks you got the error, even when  httpS is enabled!

python/foglamp/common/web/tls_helper.py
amarendra-dianomic: AssertionError extends Exception class and as such it will NEVER be hidden by -O or -OO flags.

python/requirements.txt
amarendra-dianomic: @Singhal-Vaibhav can you please help in ascertaining pre and post FOGL-2471 build sizes?

python/foglamp/services/core/server.py
```diff
@@ -163,6 +165,14 @@ class Server:
             'displayName': 'Auth Providers',
             'order': '8'
         },
+        'authMethod': {
```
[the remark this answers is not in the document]
amarendra-dianomic: authProviders is the source of auth info - user or LDAP. We may choose to use LDAP in future, hence we should retain it.

amarendra-dianomic:
@praveen-garg re your queries:
>1. what certificate name config item is there for cert based authentication? the current certificateName config item is for HTTPS.

  Not particular abt certificate name. Only requirement is that CN should be same as username. Since user will be required to either a) load certificate content vide GUI or b) upload certificate vide -T option in curl.

>2. shall more than 1 user shall be able to login using same certificate, if shared?
  
yes, this should not be a problem.

praveen-garg:
Well, we should not hardcode to use ca.crt admin should be allowed to upload any to certificate store with key, as we do allow for enable https! Better we provide a authCertificateName config item and use that; this will allow to use existing GUI interface and allow admin to upload and use new signing certificates

python/foglamp/services/core/server.py
```diff
@@ -163,13 +165,28 @@ class Server:
...
             'description': 'Certificate file name',
             'type': 'string',
             'default': 'foglamp',
             'displayName': 'Certificate Name',
             'order': '4'
         },
+        'authCertificateName': {
+            'description': 'CA Certificate file name',
+            'type': 'string',
+            'default': 'ca.cert',
```
praveen-garg: without extension it should be. And let's internally allow crt and cert both; later upload mechanism will allow .crt too and we should think of supporting .pem files.

the same logic we have for https

python/foglamp/services/core/server.py
```diff
@@ -163,13 +165,28 @@ class Server:
...
+        },
         'certificateName': {
             'description': 'Certificate file name',
             'type': 'string',
             'default': 'foglamp',
             'displayName': 'Certificate Name',
             'order': '4'
         },
+        'authCertificateName': {
+            'description': 'CA Certificate file name',
```
praveen-garg: CA Certificate file name for authentication

python/foglamp/services/core/server.py
```diff
@@ -163,13 +165,28 @@ class Server:
...
+            'type': 'string',
+            'default': 'ca.cert',
+            'displayName': 'CA Certificate file name',
```
praveen-garg: Auth Certificate Name

python/foglamp/services/core/api/auth.py
```diff
@@ -7,15 +7,20 @@
...
+import os
+import json
 from collections import OrderedDict
 
 from aiohttp import web
 from foglamp.services.core.user_model import User
 from foglamp.common.web.middleware import has_permission
 from foglamp.common import logger
+from foglamp.common.web.tls_helper import BaseCert, VerificationError
+from foglamp.common.configuration_manager import ConfigurationManager
```
praveen-garg: Should not we use user_model for db interaction? its already imported there. We can call the verify_certificate(cert, certs_dir): defined in user_model

amarendra-dianomic:
@praveen-garg all final changes done and tested ok. Pl check.

amarendra-dianomic:
@ashish-jabble changes done. Pl check.

python/foglamp/services/core/server.py
```diff
@@ -178,21 +173,43 @@ class Server:
...
+            'type': 'enumeration',
+            'options': ["any", "password", "certificate"],
+            'default': 'any',
+            'displayName': 'Authentication method',
+            'order': '6'
+        },
+        'authCertificateName': {
+            'description': 'Auth Certificate name',
+            'type': 'string',
+            'default': 'ca',
```
ashish-jabble: readonly?

praveen-garg: No, Please refer  https://github.com/foglamp/FogLAMP/pull/1438#issuecomment-470190204

ashish-jabble:
As per JIRA acceptance criteria
> The options of “Password” and “Certificate” will allow just the prescribed method and block the other method in each case

Below scenarios should be handled properly and revisit exception handling especially 500 cases:

 *auth mandatory + authMethod (password) -* Block certificate login
 *auth mandatory + authMethod (certificate) -* Block password login

For example:

curl -X POST -d '-----BEGIN CERTIFICATE' http://localhost:8081/foglamp/login
{"error": {"message": "[JSONDecodeError] Expecting value: line 1 column 1 (char 0)"}}

python/foglamp/services/core/api/auth.py
```diff
@@ -62,35 +64,61 @@
         curl -X POST -d '{"username": "user", "password": "foglamp"}' http://localhost:8081/foglamp/login
     """
 
-    data = await request.json()
-    username = data.get('username')
-    password = data.get('password')
+    data = await request.text()
+    if data.startswith("-----BEGIN CERTIFICATE-----"):
```
ashish-jabble: Add `&& data.endswith('-----END CERTIFICATE-----')`
Is it better check to deal with?

python/foglamp/services/core/api/auth.py
```diff
@@ -62,35 +64,61 @@
...
-    username = str(username).lower()
+        try:
+            await User.Objects.verify_certificate(cert)
+            username = cert.get_common_name()
+            uid, token, is_admin = await User.Objects.certificate_login(username, host)
+            # set the user to request object
+            request.user = await User.Objects.get(uid=uid)
+            # set the token to request
+            request.token = token
+        except (VerificationError, User.DoesNotExist) as e:
```
ashish-jabble: Move `VerificationError` to the next except block

python/foglamp/services/core/api/auth.py
```diff
@@ -62,35 +64,61 @@
...
+            raise web.HTTPUnauthorized(reason="Authentication failed")
```
ashish-jabble: It should be 404 for `User.DoesNotExist`

praveen-garg: In the case of certificate login, it should be HTTPUnauthorized only. Or we should find a mechanism to know username and check that first and based on that return DoesNotExist  and if exists and VerificationError then   HTTPUnauthorized.
**TO keep things simple, on login with certificate - HTTPUnauthorized is good enough, as all info is embedded in certificate only**

python/foglamp/services/core/api/auth.py
amarendra-dianomic: We can add but will it not be an overkill?

python/foglamp/services/core/api/auth.py
amarendra-dianomic: This is intentional.

python/foglamp/services/core/api/auth.py
praveen-garg: agree with @amarendra-dianomic!  we just want to know login is with certificate? and verification is being done for certificate

python/foglamp/services/core/api/auth.py
```diff
@@ -61,36 +62,68 @@
     :Example:
         curl -X POST -d '{"username": "user", "password": "foglamp"}' http://localhost:8081/foglamp/login
```
praveen-garg: Please update the curl example with certificate

PR #1441, opened 2019-03-06
FOGL-2548 Allow readings per minute and per hour as well as per second

MarkRiddoch:


C/services/south/south.cpp
```diff
@@ -599,15 +629,15 @@ int SouthService::createTimerFd(int usecs)
 	   Logger::getLogger()->error("clock_gettime");
 
 	new_value.it_value.tv_sec = now.tv_sec;
```
AmandeepArora: I think "rate.tv_sec" would also be added on the LHS. new_value.it_value is for the initial expiry and new_value.it_interval is the periodic interval for each subsequent timer expiry.

MarkRiddoch: Well spotted, the initial all would have been made early.

PR #1444, opened 2019-03-07
FOGL-2150 - Create a new endpoint to DELETE category

amarendra-dianomic:


ashish-jabble:
What if we delete the category of running `south service & north instance` (Simliar for Active filter)? On next start up service/task is crashed. I think we should restrict some cases for deletion of category.

praveen-garg:
Role based restrictions will be there.

PR #1446, opened 2019-03-07
FOGL-2488: Various SQLite related optimizations and other fixes/enhancements

AmandeepArora:
Have tested extensively on a local constrained VM, Raspberry Pi and a low end AWS instance. It is working fine on all platforms. Slight tweaking of purge block size can be done for low end platforms to allow better interleaving between simultaneous writers to DB. Would add code in future to determine this block size dynamically at runtime.


praveen-garg:
lgtm! @Singhal-Vaibhav please run existing system tests on this PR.

AmandeepArora:
Changes done in this PR:
-------------------------

1. Reduced lock contention between readers/writers in Sqlite plugin code
2. DB retry interval granularity reduced from 2s to 100 ms to take advantage of better interleaving in WAL mode
3. Between the 2 writing threads - ingest and purge - added mutex and condition_variable so that purge takes permission from the ingest thread (if required) for each block deletion
4. Used SQLITE_OPEN_NOMUTEX so that database connection are opened in the multi-thread threading mode
5. Added sqlite PRAGMAs for using 4MB cache, enabling WAL journaling mode, switching off secure delete & limiting journal size to 4MB when possible
6. Set SQLite DB page_size to 4KB for new SQLite databases created using foglamp reset (default page size is 1 KB upto sqlite v3.11)
7. Bulk insert support added in Postgres & Sqlite (Massimiliano)
8. Added code to update statistics_history table using the new bulk insert feature (Massimiliano)
9. Corrected handling of return status of bulk insert/update transactions
10. Optimized purge related queries based on age of reading row. Index on readings table based on user_ts were not being used since user_ts is internally stored as a string, so it cannot support numerical operators like less than, greater than, etc. So implemented binary search to find the most recent reading that is old enough to be purged.
11. Mark also provided some code for purge query optimization/cleanup
12. Modified datetime function params used in purge procedure to properly consider local TZ and UTC timestamps in DB.
13. If purge of a block takes more than 150 ms, there is a penalty/sleep done by purge thread. This is to avoid WAL mode checkpoint starvation and to give higher priority to other threads.
14. Added code in Connection::SQLexec() to rollback aborted transaction in case of bulk update/insert failure due to DB lock contention, so that subsequent retries work fine.
15. Added timestamp of query in storage profiling code besides time taken, so that nearby logs can be checked in /var/log/syslog for better correlation.
16. Enabled worker threads in Storage API code (Massimiliano)
17. Core/scheduler waits for 5 mins for purge to complete before it times out and then the scheduler can fire another purge request at next scheduled time. In earlier code, at times, storage was running mutiple purge threads simultaneously. Added code to reject new purge requests if one is already ongoing.
18. Removed few expensive queries from core/purge python code that were not really required or could be replaced with alternate simpler queries (Praveen)
19. Dropped readings table index on uuid field (Massmimiliano)

Singhal-Vaibhav:
No regression seen after running system tests.

ashish-jabble:
Also please add below files in _.gitignore_ file
```
data/*.db-shm
data/*.db-wal

OR

data/*.db*
```

Singhal-Vaibhav:
Happy paths in the pull are fine and seems to be no issues, however there are few concerns:
1. On Sqlite, this pull creates 2 additional files apart from foglamp.db, foglamp.db-shm and foglamp.db-wal. If these files are important, should we consider changing the logic of creating the support bundle to include these files?
2. Checked on a heavy DB (1M rows) on raspberry PI, purge runs forever, and if purge has started its almost impossible to interact with foglamp instance. foglamp stop doesn't work
3. Extensive logging like:
```
Mar 14 17:42:55 ubuntu FogLAMP Storage[25389]: INFO: Purge loop slept for 168 msecs since removal of a block took 686324 usecs
``` 
should be in debug mode instead of INFO
4. Behaviour changes like mentioned in FOGL-2493 should have been a part of separate ticket. If not, its worth mentioning that in pull description and release notes.
5. We need to quantify the performance gain we get from these changes.

praveen-garg:
> Happy paths in the pull are fine and seems to be no issues, however there are few concerns:
> 
> 1. On Sqlite, this pull creates 2 additional files apart from foglamp.db, foglamp.db-shm and foglamp.db-wal. If these files are important, should we consider changing the logic of creating the support bundle to include these files?

  We are just copying configuration, audit_log, schedules, scheduled_processes tables in support bundle. so it should not matter.

but I'm not sure about **backup and restore** ; and yes we should add them to .gitignore

> 2. Checked on a heavy DB (1M rows) on raspberry PI, purge runs forever, and if purge has started its almost impossible to interact with foglamp instance. foglamp stop doesn't work

on develop, we are good with the same scenario?

>  3. Extensive logging like:
> 
> 
> ```
> Mar 14 17:42:55 ubuntu FogLAMP Storage[25389]: INFO: Purge loop slept for 168 msecs since removal of a block took 686324 usecs
> ```
> 
> should be in debug mode instead of INFO

agree

> 4. Behaviour changes like mentioned in FOGL-2493 should have been a part of separate ticket. If not, its worth mentioning that in pull description and release notes.

I think FOGL-2493 is a story under this FOGL-2488 epic. so even we will need to consider all such stuff for documentation.

if I understand correctly, the change is, We should always see increasing numbers for readings (stats) in ping, not currently buffered; so readings = buffered + purged

> 5. We need to quantify the performance gain we get from these changes.

agree! hopefully the job runs on jenkins will help on that. and also we should record the SQL profiling and no. of n/w requeste results  w.r.t. develop.

AmandeepArora:
>     1. On Sqlite, this pull creates 2 additional files apart from foglamp.db, foglamp.db-shm and foglamp.db-wal. If these files are important, should we consider changing the logic of creating the support bundle to include these files?
Aman: Please create ticket for this and/or for backup/restore, as required. Also I am not sure how you can get a clean atomic snapshot of these DB files whie foglamp is running. Added these 2 new files to .gitignore.

> 
>     2. Checked on a heavy DB (1M rows) on raspberry PI, purge runs forever, and if purge has started its almost impossible to interact with foglamp instance. foglamp stop doesn't work
Aman: Will investigate. 
> 
>     3. Extensive logging like: 
> ```
> Mar 14 17:42:55 ubuntu FogLAMP Storage[25389]: INFO: Purge loop slept for 168 msecs since removal of a block took 686324 usecs
> ```
> should be in debug mode instead of INFO
Aman: Relegated to debug level.

PR #1447, opened 2019-03-07
FOGL-2306: Separate sending process configuration into advanced and basic configuration blocks

pintomax:
FOGL-2306: Separate sending process configuration into advanced and basic configuration blocks

blockSize, duration, sleepInterval, memoryBufferSize configuration items, used by sending process only, are now part of $categoryNameAdvanced

Merge to develop can be done when FOGL-2567 is done


pintomax:
Ready to merge to develop ...

PR #1448, opened 2019-03-07
FOGL-2397: taskStatus issue on recreating north instance with same name

sstefano69:
**Defined direction**
We are going to prevent the recreation of a task with the same name, so the step 3 of the below repro will fail, the right solution will be to create a new task with a different name, like for example 'P1_v2'.

We will check the tasks table at the creation of a task to evaluate if there are rows already having that schedule_name and in this case we will abort the creation.

message : 
```
400: Invalid name north bound, there are tasks already executed with this name.
```


python/foglamp/services/core/api/task.py
```diff
@@ -180,6 +180,21 @@
...
+        payload = PayloadBuilder() \
+            .SELECT(["id", "schedule_name"]) \
+            .WHERE(['schedule_name', '=', name]) \
+            .LIMIT(1) \
+            .payload()
+
+        result = await storage.query_tbl_with_payload('tasks', payload)
+
+        if result['count'] >= 1:
+            msg = 'Invalid name {0}, there are tasks already executed with this name.'.format(name)
```
MarkRiddoch: I don't think this message is very meaningful to the user. Can I suggest something along the lines of "Unable to reuse name {}, already used by a previous task."

sstefano69: thanks for the piece of advice, fixed

PR #1462, opened 2019-03-18
FOGL-2609 - Remove microseconds details from ping time, and keep that in seconds only

amarendra-dianomic:


python/foglamp/services/core/api/common.py
```diff
@@ -86,7 +86,7 @@ def services_health_litmus_test():
     status_color = services_health_litmus_test()
     safe_mode = True if server.Server.running_in_safe_mode else False
 
-    return web.json_response({'uptime': since_started,
+    return web.json_response({'uptime': int(since_started),
```
praveen-garg: thanks for this (I will update jira for this reference too)

Also please fix for `python.foglamp.services.core.server.Server#ping`

README.rst
```diff
@@ -91,6 +91,10 @@ To build FogLAMP run the command ``make`` in the top level directory. This will
 Once the *make* has completed you can decide to test FogLAMP from your development environment or you can install it. 
 |br| |br|
 
+
+**NOTE:** *Please note that openssl toolkit is a requirement if we want to use https based REST client and certificate based authentication.*
```
praveen-garg: Instead of another NOTE, please put as bullet point under the existing NOTE for GCC in this section

amarendra-dianomic: done

python/foglamp/services/core/api/common.py
ashish-jabble: We have missed to fix tests for `$FogLAMP_ROOT/tests/unit/python/foglamp/services/core/api/test_common_ping.py`; as these were failing on local environment ONLY.
```
 assert 0.0 < content_dict["uptime"]
```
Let say Previously O/p of  `UPTIME.... 0.07508373260498047` And now `0`. So above assertion is failing as it executes fast ( and less than 1sec) on local setup.

praveen-garg: do you think better option will be `asyncio.sleep(1)` and verify `> 1` ?

PR #1466, opened 2019-03-19
FOGL-2573 (system test) added wait to read data from kakfa broker

praveen-garg:


tests/system/python/e2e/test_e2e_kafka.py
```diff
@@ -128,25 +128,41 @@ def test_end_to_end(self, start_south_north, foglamp_url, wait_time, kafka_host,
...
+        self._clean_kafka_instance(host, rest_port)
+
+        conn = http.client.HTTPConnection("{}:{}".format(host, rest_port))
         # Assertions on Kafka topic, consumer and subscription
         self._verify_kafka_topic(conn, topic)
-
         self._verify_kafka_topic_by_name(conn, topic)
-
         # Create a consumer group and instance
         self._verify_kafka_consumer_group_and_instance(conn)
```
praveen-garg: with latest confluent kafka setup, there is a Assertion failure with 415 status code.

tests/system/python/e2e/test_e2e_kafka.py
```diff
@@ -128,25 +128,41 @@ def test_end_to_end(self, start_south_north, foglamp_url, wait_time, kafka_host,
...
+        :param kafka_rest_port:
+        :return:
+        """
+        self._close_consumer(kafka_host, kafka_rest_port)
 
-        # Close the consumer (DELETE) to make it leave the group and clean up its resources
-        self._close_consumer(host, rest_port)
+    def _read_from_kafka(self, host, rest_port, topic, wait_time):
 
+        self._clean_kafka_instance(host, rest_port)
```
praveen-garg: _clean_kafka_instance should not be just after ingest into kafka (egress from foglamp north) and before verify! it should be just before the fogbench is executed and north task send that.

PR #1468, opened 2019-03-20
FOGL-2345: System test for hierarchical FogLAMPs

Singhal-Vaibhav:
This test runs FogLAMP on 2 machines. This tests adds sinusoid, expression and playback and HTTP North plugin (with asset filter) on FogLAMP instance 1. Send these via HTTP north to a second FogLAMP (instance 2) that then sends this data onward to a PI system

**Test Assumptions:**
1. You are able to ssh into remote machine without password (using ssh keys)
```
If you already have an SSH key, you can skip this step… Just hit Enter for the key and both passphrases:
$ ssh-keygen -t rsa -b 2048
Generating public/private rsa key pair.
Enter file in which to save the key (/home/username/.ssh/id_rsa): 
Enter passphrase (empty for no passphrase): 
Enter same passphrase again: 
Your identification has been saved in /home/username/.ssh/id_rsa.
Your public key has been saved in /home/username/.ssh/id_rsa.pub.

Copy your keys to the target server:
$ ssh-copy-id id@server
id@server's password: 

Now try logging into the machine, with ssh 'id@server', and check in:
.ssh/authorized_keys
to make sure we haven’t added extra keys that you weren’t expecting.

Finally check logging in…
$ ssh id@server
id@server:~$ 
```

2. FogLAMP is already cloned and built on remote machine 

**To Run:**
```
pytest -s -vv test_e2e_foglamp_pair.py --remote-user=<remote_username> --remote-ip=<remote_ip> --key-path=<key_file for auth to remote machine> --pi-db=<pi_db> --pi-host=<pi_host> --pi-admin=<pi_user> --pi-passwd=<pi_pwd> --pi-token="<pi_token>"
```
==== 1 passed in 214.70 seconds ====


tests/system/python/scripts/install_python_plugin_remote
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env bash
+set -e
+
+__author__="Vaibhav Singhal"
+__copyright__="Copyright (c) 2019 Dianomic Systems"
+__license__="Apache 2.0"
+__version__="1.0.0"
+
+###########################################################################################################
+# Usage text for this script
```
praveen-garg: can not we just copy the install_python_plugin script to remote and execute that; Also the exitCode of that should give us info for successful execution.

Singhal-Vaibhav: Done. Nice suggestion!

PR #1470, opened 2019-03-20
FOGL-2585 - Do not allow deletion of Notification service if notifications exist

amarendra-dianomic:


python/foglamp/services/core/api/service.py
```diff
@@ -137,7 +145,7 @@ def get_service_records():
 
         if name is None:
             raise web.HTTPBadRequest(reason='Missing name property in payload.')
-        if utils.check_reserved(name) is False:
+        if utils.check_reserved(name) is False or utils.check_foglamp_reserved(name) is False:
             raise web.HTTPBadRequest(reason='Invalid name property in payload.')
```
praveen-garg: We should make this message better.

python/foglamp/common/utils.py
```diff
@@ -37,6 +37,27 @@ def check_reserved(string):
...
+        'foglamp',
+        'general',
+        'advanced',
+        'notifications',
+        'north',
+        'south',
+        'filter',
+        'notify',
+        'rule',
+        'delivery',
```
praveen-garg: rule and delivery are also reserved?

amarendra-dianomic: yes, these have a pre defined meaning in Foglamp, so should not be used as service name.

python/foglamp/services/core/api/service.py
amarendra-dianomic: done

amarendra-dianomic:
@praveen-garg all done, pl check.

PR #1473, opened 2019-03-21
Fogl 2636: fix a possible memory leak in the usage of the 'escape' function in the postgress and SQLITE

sstefano69:


sstefano69:
Closed.

PR #1478, opened 2019-03-25
FOGL-2571: Add missing assertion of verify statistics and ping in system tests

Singhal-Vaibhav:
Fixes the following:

- [X] Assertion added to verify ping response
- [X] Assertion added to verify statistics response
- [X] verify_egress method added to wrap assertions from north api
- [X] flag added to verify data at north side (default value is true)

### DIFF
--- a/tests/system/python/conftest.py
+++ b/tests/system/python/conftest.py
@@ -9,6 +9,7 @@
 """
 import subprocess
 import os
+import sys
 import fnmatch
 import http.client
 import json
@@ -22,6 +23,8 @@ __author__ = "Vaibhav Singhal"
 __copyright__ = "Copyright (c) 2019 Dianomic Systems"
 __license__ = "Apache 2.0"
 __version__ = "${VERSION}"
+
+sys.path.append(os.path.join(os.path.dirname(__file__), 'helpers'))
 
 
 @pytest.fixture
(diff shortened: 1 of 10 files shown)


tests/system/python/conftest.py
```diff
@@ -291,6 +291,9 @@ def pytest_addoption(parser):
                      help="Generic wait time between processes to run")
     parser.addoption("--retries", action="store", default=3, type=int,
                      help="Number of tries for polling")
+    # TODO: Temporary fixture, to be used with value False for environments where PI Web API is not stable
+    parser.addoption("--verify-north-data", action="store", default='true',
```
praveen-garg: we should change it to verify-north-interface, verify-north-data may be understood with locally stored data

praveen-garg: `action="store_false", default=True`

tests/system/python/conftest.py
```diff
@@ -381,6 +384,15 @@ def remote_foglamp_path(request):
     return request.config.getoption("--remote-foglamp-path")
 
 
+@pytest.fixture
+def verify_north_data(request):
+    v = request.config.getoption("--verify-north-data")
+    if v.lower() in ('no', 'false', 'f', 'n', '0'):
```
praveen-garg: this should not be needed with store_false action

tests/system/python/e2e/test_e2e_coap_PI.py
```diff
@@ -24,6 +24,43 @@
...
+def get_statistics_map(foglamp_url):
+    _connection = http.client.HTTPConnection(foglamp_url)
+    _connection.request("GET", '/foglamp/statistics')
+    r = _connection.getresponse()
+    assert 200 == r.status
+    r = r.read().decode()
+    jdoc = json.loads(r)
+    actual_stats_map = {}
+    for itm in jdoc:
+        actual_stats_map[itm['key']] = itm['value']
```
praveen-garg: can we make this a common `utils::serialize_stats`, to avoid rework

PR #1484, opened 2019-03-27
FOGL-1802: addition of ORDER BY column number

pintomax:
FOGL-1802: addition of ORDER BY column number

First implementation:

-  curl -sX GET http://localhost:8081/foglamp/asset/Flow/random/series OK
- unit tests: OK

SELECT min(("reading"->>'random')::float) AS "min", max(("reading"->>'random')::float) AS "max", avg(("reading"->>'random')::float) AS "average", to_char("user_ts", 'YYYY-MM-DD HH24:MI:SS') AS "timestamp" FROM foglamp.readings WHERE "asset_code" = 'Flow' AND reading ? 'random' AND reading ? 'random' AND reading ? 'random' GROUP BY to_char("user_ts", 'YYYY-MM-DD HH24:MI:SS') ORDER BY 4 desc LIMIT 20;

Postgres test "group_time.json" handles this change


C/plugins/storage/postgres/connection.cpp
```diff
@@ -1980,9 +1995,27 @@ bool Connection::jsonModifiers(const Value& payload, SQLBuffer& sql)
...
+			    groupColumn.compare(sortBy["column"].GetString()) == 0 &&
+			    nAggregates)
+			{
+				// Increment number of columns
+				// as the GROUP BY column is added as last one
+				// in the column names for SELECT
+				// The ORDER BY column name is replaced by the column
+				// number, without diuble quotes
+				// example: SELECT ... ORDER BY 4
+				sql.append(++nAggregates);
```
MarkRiddoch: I would have used nAggregates + 1 here and not done the increment as it means nAggregates is now a misleading number. But it works intros case as we never use nAggregates again.

pintomax: Fixed

PR #1488, opened 2019-03-29
FOGL-2544: merge the SQlite plugin and the SQlite in memory pluing in…

pintomax:
… a single set of sources

FOGL-2544: merge the SQlite plugin and the SQlite in memory pluing in a single set of sources

Fist implementation.

sqlite/connection.cpp will be spit in common_tables.ccp and readings_table.cpp in next commits


pintomax:
Errors in the tests, a ne one will come

PR #1489, opened 2019-03-29
FOGL-2650: Fix postgres unit tests for the execution on the remote testing system

sstefano69:
Select the proper storage.json file for the tests.

### DIFF
--- a/tests/unit/C/services/storage/postgres/tests.cpp
+++ b/tests/unit/C/services/storage/postgres/tests.cpp
@@ -8,7 +8,7 @@ using namespace std;
 int main(int argc, char **argv) {
     testing::InitGoogleTest(&argc, argv);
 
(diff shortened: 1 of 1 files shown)


tests/unit/C/services/storage/postgres/tests.cpp
```diff
@@ -2,31 +2,26 @@
...
-/**
- * Select the proper storage.json file for the tests
- */
-void cache_file_select()
-{
+	testing::InitGoogleTest(&argc, argv);
 
-	string foglamp_root = getenv("FOGLAMP_ROOT");
-	string foglamp_data = foglamp_root + "/tests/unit/C/services/storage/postgres";
+	testing::GTEST_FLAG(repeat) = 5000;
```
pintomax: We can use 500 or 1000 here

PR #1491, opened 2019-03-29
FOGL-2544: merge the SQlite plugin and the SQlite in memory pluing in…

pintomax:
… a single set of sources

FOGL-2544: merge the SQLite plugin and the SQLite in memory plugin in a single set of sources

Tests compared with "develop" are ok


MarkRiddoch (approved):
Review of this is pretty much impossible given the changes. On the basis that all the tests are passing we will go with this change and then cleanup remaining issues with it later.

PR #1492, opened 2019-04-01
FOGL-2653: Add table snapshot facility to storage layer

pintomax:
FOGL-2653: Add table snapshot facility to storage layer

Added new entry points in Storage API

Added three new API entry points for storage plugins

Added 1.1.0 for Plugin interface API version to Postgres plugin
Still 1.0.0 version for  Postgres plugin


C/plugins/storage/postgres/connection.cpp
```diff
@@ -2546,3 +2546,32 @@ void Connection::logSQL(const char *tag, const char *stmt)
...
+ * @param id		The snapshot id
+ * @return		-1 on error, >= 0 on success
+ *
+ * The new created table name has the name:
+ * table_id
+ */
+int Connection::create_table_snapshot(const string& table, const string& id)
+{
+	string query = "SELECT * INTO TABLE foglamp.";
+	query += table + "_" +  id + " FROM foglamp.configuration";
```
MarkRiddoch: Should be table, not "foglamp.confoiguration"

pintomax: Yes, oversight!

C/plugins/storage/postgres/connection.cpp
```diff
@@ -2546,3 +2546,87 @@ void Connection::logSQL(const char *tag, const char *stmt)
...
+ *
+ * @param table		The table to fill
+ * @param id		The snapshot id of the table
+ * @return		-1 on error, >= 0 on success
+ *
+ */
+int Connection::load_table_snapshot(const string& table, const string& id)
+{
+	string query = "INSERT INTO foglamp." + table;
+	query += " SELECT * FROM foglamp." + table + "_" + id;
```
AmandeepArora: Though not sure, but it seems data from snapshot table is being appended to original table? Is it possible that there might be duplicate entries after the append operation?

C/plugins/storage/postgres/plugin.cpp
```diff
@@ -39,7 +39,7 @@ static PLUGIN_INFORMATION info = {
 	"1.0.0",                  // Version
 	SP_COMMON|SP_READINGS,    // Flags
 	PLUGIN_TYPE_STORAGE,      // Type
-	"1.0.0"                   // Interface version
+	"1.1.0"                   // Interface version
```
AmandeepArora: These 3 new functions in storage plugin should be called only if the interface version is >= "1.1.0". I didn't find that check in this PR.

C/plugins/storage/postgres/connection.cpp
pintomax: The command above creates a new table "aaa_123" with the content of "aaa"

C/plugins/storage/postgres/plugin.cpp
pintomax: There is no check for that right now

C/plugins/storage/postgres/connection.cpp
AmandeepArora: Isn't this restore function (load_table_snapshot) - bringing data back from snapshot table to original table? I am saying original table might have some data already. Is that possible?

pintomax: Oh yes.

If you try to "restore" configuration table, currently not empty, there will be some insert failures

pintomax: The operation PUT then returns the error message from database.
Same for POST (existing snapshot) or DELETE (non existing snapshot): errors are rerturned

PR #1497, opened 2019-04-02
FOGL-2676: C Unit test /storage/sqlite - fix test id 42

sstefano69:


C/plugins/storage/sqlite/connection.cpp
```diff
@@ -2465,8 +2465,11 @@ int blocks = 0;
 		{
 			unsigned long midRowId = 0;
 			unsigned long prev_m = m;
-		    m = l + (r - l) / 2;
-			if (prev_m == m) break;
+		    	m = l + (r - l) / 2;
+
+			// It applies the check if there a more than a very limited number of rows
+			if ( (rowidLimit - minrowidLimit) > 10 )
```
AmandeepArora: I don't follow the logic here. The break in original code happens when the binary search ends. With this new 'if' condition, how would binary search end?

For what values of rowidLimit, minrowidLimit and other params to this function, do you see the original  issue? With these changes, the normal purge process works?

sstefano69: talked/defined with Aman

sstefano69:
#closed and reopen

PR #1499, opened 2019-04-03
FOGL-2652, 2654 - configuration, schedules tables snapshot API

amarendra-dianomic:
```
$ curl -X POST http://localhost:8081/foglamp/snapshot/category | jq
{
  "id": "1555066394",
  "table": "configuration"
}

$ curl -X PUT http://localhost:8081/foglamp/snapshot/category/1555066394 | jq
{
  "id": "1555066394",
  "table": "configuration"
}

$ curl -X GET http://localhost:8081/foglamp/snapshot/category | jq
{
  "snapshots": [
    {
      "id": "1555066394"
    },
    {
      "id": "1555061750"
    },
    {
      "id": "1555061734"
    }
  ]
}

$ curl -X DELETE http://localhost:8081/foglamp/snapshot/category/1555061734 | jq
{
  "id": "1555061734",
  "table": "configuration"
}

$ curl -X GET http://localhost:8081/foglamp/snapshot/category | jq
{
  "snapshots": [
    {
      "id": "1555066394"
    },
    {
      "id": "1555061750"
    }
  ]
}
```


python/foglamp/services/core/routes.py
```diff
@@ -175,6 +176,12 @@ def setup(app):
     app.router.add_route('PUT', '/foglamp/notification/{notification_name}', notification.put_notification)
     app.router.add_route('DELETE', '/foglamp/notification/{notification_name}', notification.delete_notification)
 
+    # Snapshot config
+    app.router.add_route('GET', '/foglamp/config/snapshot', snapshot_config.get_snapshot)
```
MarkRiddoch: let's make the route /foglamp/snapshot/category - that should solve the issue with conflicts with the existing API and avoids the mixing of category and config. 

We can make all snaps calls use /foglamp/snapshot/...

python/foglamp/services/core/api/snapshot_config.py
```diff
@@ -0,0 +1,85 @@
+# -*- coding: utf-8 -*-
```
praveen-garg: We should have `core/api/snapshot/tables.py` and map the routes for supported table names [configuration, schedules].

python/foglamp/services/core/api/snapshot_config.py
```diff
@@ -0,0 +1,85 @@
...
+async def get_snapshot(request):
+    """ get list of available snapshots
+
+    :Example:
+        curl -X GET http://localhost:8081/foglamp/category/snapshot
+    """
+    _storage = get_storage_async()  # from foglamp.services.core.connect
+    try:
+        retval = await _storage.get_snapshot('configuration')
+    except Exception as ex:
```
praveen-garg: Please handle StorageServer Error, raised by storage client (for all endpoint def)

python/foglamp/services/core/api/snapshot_config.py
```diff
@@ -0,0 +1,85 @@
...
+        raise web.HTTPInternalServerError(reason='Category table snapshots could not be fetched. {}'.format(str(ex)))
+    else:
+        return web.json_response(retval)
```
praveen-garg: I suggest to parse retval and return more human friendly messages

python/foglamp/services/core/api/snapshot_config.py
MarkRiddoch: No, the snapshot will work with any table, so there is no need to do this.

praveen-garg: You mean we should not restrict to `[configuration, schedules]` tables?

MarkRiddoch: Correct. The only table you can not snapshot is the readings. That is because there is no API to do it in the storage layer.

amarendra-dianomic: @MarkRiddoch now there is some mismatch here vis-a-vis specs. Per specs mentioned in FOGL-2652 and FOGL-2654, we are required to define table specific routes for the tables configuration and schedules. Meaning that table snapshot route definitions are not generic and are bound to a specific table. If we want routes for table snapshot to be generic, then we must define routes accordingly. Something like:

POST /foglamp/snapshot/table/{table_name}
GET /foglamp/snapshot/table/{table_name}
GET /foglamp/snapshot/tables
PUT /foglamp/snapshot/table/{table_name}/id/{id}
DELETE /foglamp/snapshot/table/{table_name}/id/{id}

Am I missing something?

praveen-garg:
The response for GET has “rows” key, it looks bit awkward to me.  And should not we know which id belongs to what table? So that we know what route (for PUT and DELETE) to call, category for configuration or schedules! 

How do we know, which one is the latest or oldest?  (Creation time)

amarendra-dianomic:
@praveen-garg the snapshot id is the int part of unix timestamp, hence it will always increase with time. Hence the largest snapshot id will be the latest one and so on.

python/foglamp/services/core/api/snapshot/table.py
```diff
@@ -0,0 +1,111 @@
...
+
+from aiohttp import web
+
+from foglamp.services.core.connect import *
+from foglamp.common.storage_client.exceptions import StorageServerError
+from foglamp.common import logger
+
+
+__author__ = "Amarendra K Sinha"
+__copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
```
ashish-jabble: 2019 Dianomic Systems

python/foglamp/common/storage_client/storage_client.py
```diff
@@ -311,6 +312,97 @@ def disconnect(self):
...
+        post_url = '/storage/table/{tbl_name}/snapshot'.format(tbl_name=tbl_name)
+        data = {"id": str(int(time.time()))}
+
+        url = 'http://' + self.base_url + post_url
+        async with aiohttp.ClientSession() as session:
+            async with session.post(url, data=json.dumps(data)) as resp:
+                status_code = resp.status
+                jdoc = await resp.text()
+                if status_code not in range(200, 209):
+                    _LOGGER.info("POST %s", post_url)
```
ashish-jabble: If `post_url` info is required, then it should be before this condition or if its for debug purpose either remove it or with .debug. Same comments applied to all endpoints

python/foglamp/services/core/api/snapshot/table.py
```diff
@@ -0,0 +1,111 @@
...
+_help = """
+    -------------------------------------------------------------------------------
+    | GET POST        | /foglamp/snapshot/category                                |
+    | PUT DELETE      | /foglamp/snapshot/category/{id}                           |
+    | GET POST        | /foglamp/snapshot/schedule                                |
+    | PUT DELETE      | /foglamp/snapshot/schedule/{id}                           |
+    -------------------------------------------------------------------------------
+"""
+
+_logger = logger.setup()
```
ashish-jabble: `_logger = logger.setup(__name__)`

python/foglamp/common/storage_client/storage_client.py
praveen-garg: In my opinion this is okay as-is. Only in the case of error, we would prefer to see (info) what url was called and what error is there, even in production system.

amarendra-dianomic:
Re feedback from @Singhal-Vaibhav above,
1 - Unable to reproduce as postgres snapshot is working fine at my local.
2 - Error messages fixed with correct response code.
3, 4 - @MarkRiddoch please advise.

python/foglamp/services/core/api/snapshot/table.py
```diff
@@ -67,7 +68,7 @@
     except (StorageServerError, Exception) as ex:
         raise web.HTTPInternalServerError(reason='{} table snapshot could not be created. {}'.format(table, str(ex)))
     else:
-        return web.json_response(retval)
+        return web.json_response(retval["created"])
```
ashish-jabble: Better to handle like below
```
result = retval["created"]
......
except KeyError as ex:
        raise web.HTTPNotFound(reason='{}........'.format(str(ex)))
else:

```
Also same for retval["loaded"], ["deleted"] etc.

amarendra-dianomic: Yes, this is yet another valid approach. But let us stick to on relying upon StorageServerError for storage layer errors, for the sake of uniformity, as we have been doing at other places.

MarkRiddoch:
> 2\. As a user, I created multiple snapshots 1 year back, now i want to restore to a snapshot. What I dont know is which snapshot to restore to. Are we expecting an admin user to keep track of snapshot id with the purpose of that snapshot? I think we should add a description with each snapshot

No, snapshots are not intended for users to use, they are there to facilitate functionality in the management service which allows an update of a series of configuration categories, possibly across a number of FogLAMP instances, to be rolled back if an update fails. They are not intended as long term backup mechanisms.

python/foglamp/common/storage_client/storage_client.py
```diff
@@ -311,6 +312,97 @@ def disconnect(self):
...
+                status_code = resp.status
+                jdoc = await resp.text()
+                if status_code not in range(200, 209):
+                    _LOGGER.info("DELETE %s", delete_url)
+                    _LOGGER.error("Error code: %d, reason: %s, details: %s", resp.status, resp.reason, jdoc)
+                    raise StorageServerError(code=resp.status, reason=resp.reason, error=jdoc)
+        return json.loads(jdoc)
+
+    async def get_snapshot(self, tbl_name):
+        """Delete a table snapshot
```
praveen-garg: Get a table snapshot

python/foglamp/services/core/api/snapshot/table.py
```diff
@@ -0,0 +1,149 @@
...
+    "schedule": "schedules"
+}
+
+
+@has_permission("admin")
+async def get_snapshot(request):
+    """ get list of available snapshots
+
+    :Example:
+        curl -d '{"username": "user", "password": "foglamp"}' -X POST http://localhost:8081/foglamp/login
```
praveen-garg: Please remove this.

python/foglamp/services/core/api/snapshot/table.py
```diff
@@ -0,0 +1,149 @@
...
+
+        curl -X GET http://localhost:8081/foglamp/snapshot/category
+        curl -X GET http://localhost:8081/foglamp/snapshot/schedule
+        
+        When auth is mandatory:
```
praveen-garg: Okay, but these examples are not really required.

python/foglamp/services/core/api/snapshot/table.py
```diff
@@ -0,0 +1,149 @@
...
+    except (StorageServerError, Exception) as ex:
+        raise web.HTTPInternalServerError(reason='{} table snapshots could not be fetched. {}'.format(table, str(ex)))
+    else:
+        newlist = sorted(retval["rows"], key=lambda k: k['id'], reverse=True)
+        return web.json_response({"snapshots": newlist})
+
+
+@has_permission("admin")
+async def post_snapshot(request):
+    """ Create a snapshot  by name
```
praveen-garg: remove `by name`

PR #1502, opened 2019-04-04
FOGL-2676: C Unit test /storage/sqlite - fix test id 42

sstefano69:
tested : 
* timezone                    :Etc/UTC:
* timezone                    :Europe/Rome:

* Storage
* in memory

* U16
* U18
* Raspibian


C/plugins/storage/sqlite/common/readings.cpp
```diff
@@ -820,11 +820,14 @@ int blocks = 0;
 			{
 				// search in earlier/left half
 				r = m - 1;
+
+				// The m position should be skipped as midRowId is 0
+				m = r;
```
AmandeepArora: 'm' position is being skipped anyway, since for next iteration, we are setting right end as 'm-1' here. And then at beginning of next iteration, 'm' would be recalculated as middle of range (l..r).

sstefano69: the assignment is for handling the case in which it is the very last row that should be skipped as it is newer than age, actually the test id 102.

PR #1511, opened 2019-04-08
FOGL-2704 System Test for egress to PI with scale-set filter plugin

praveen-garg:


tests/system/python/e2e/test_e2e_pi_scaleset.py
```diff
@@ -0,0 +1,166 @@
...
+            data_from_pi = read_data_from_pi(pi_host, pi_admin, pi_passwd, pi_db, asset_name_with_prefix, {READ_KEY})
+            retry_count += 1
+            time.sleep(wait_time * 2)
+
+        if data_from_pi is None or retry_count == retries:
+            assert False, "Failed to read data from PI"
+
+        assert READ_KEY in data_from_pi
+        assert isinstance(data_from_pi[READ_KEY], list)
+        assert OUTPUT in data_from_pi[READ_KEY]
```
Singhal-Vaibhav: Test fails with output:
```
assert OUTPUT in data_from_pi[READ_KEY]
E       assert 69.80000000000001 in [69.7999989986]
```

praveen-garg: Fixed!

PR #1494, opened 2019-04-02
Fogl 2655 - Plugins snapshot

amarendra-dianomic:
```
$ curl -X GET http://localhost:8081/foglamp/snapshot/plugins | jq
{
  "snapshots": []
}

$ curl -X POST http://localhost:8081/foglamp/snapshot/plugins | jq
{
  "message": "snapshot id=1554808178, file=snapshot-plugin-1554808178.tar.gz created successfully."
}

$ curl -X GET http://localhost:8081/foglamp/snapshot/plugins | jq
{
  "snapshots": [
    {
      "id": "1554808178",
      "name": "snapshot-plugin-1554808178.tar.gz"
    }
  ]
}

$ rm -rf C/plugins
$ ll C/plugins
ls: cannot access 'C/plugins': No such file or directory

$ curl -X PUT http://localhost:8081/foglamp/snapshot/plugins/1554808178 | jq
{
  "message": "snapshot snapshot-plugin-1554808178.tar.gz restored successfully."
}

$ ll C/plugins
total 28
drwxrwxr-x 7 asinha asinha 4096 Mar 25 11:52 ./
drwxrwxr-x 8 asinha asinha 4096 Apr  9 16:42 ../
drwxrwxr-x 3 asinha asinha 4096 Apr  8 15:25 common/
drwxrwxr-x 3 asinha asinha 4096 Mar 25 11:52 filter/
drwxrwxr-x 4 asinha asinha 4096 Mar 25 11:52 north/
drwxrwxr-x 6 asinha asinha 4096 Apr  3 17:13 storage/
drwxrwxr-x 2 asinha asinha 4096 Apr  3 17:13 utils/
$ 

$ curl -X DELETE http://localhost:8081/foglamp/snapshot/plugins/1554808178 | jq
{
  "message": "snapshot snapshot-plugin-1554808178.tar.gz deleted successfully."
}

$ curl -X GET http://localhost:8081/foglamp/snapshot/plugins | jq
{
  "snapshots": []
}
```

PR #1515, opened 2019-04-09
FOGL-2651 FogLAMP API to install plugins

ashish-jabble:
**Assumptions:**
- This solution requires that the .tar/.tar.gz files strictly follow the structure as explained in the examples below.
- This solution will work only with "sudo make install" i.e. this solution is not available for development environment in case of deb format (we need foglamp DEBIAN installation is mandatory as foglamp.sudoers is being copied to /etc/sudoers.d/foglamp through it) but we may test in case of tar format

**Known issues**
~Not working for the case If plugin is having any external dep to install~
~Walk through FOGL-2741 comments~
See FOGL-2760

**Fixed items:**
- [x] POST route added to add plugins
- [x] Added a new params -- **type** which is required, because we need some identifier what the plugin is type so it would be better to pass explicitly in param
- [x] Checksum only with **md5Sum**. Do we need to support check with any other?

**Examples**

See the below structure for plugin upload and code is working with the same 

# Python-Plugin
```
$ tree coap
coap --- (Naming convention should be followed and it will exactly same as plugin_name)
├── coap.py
├── __init__.py
├── README.rst
└── requirements.txt   --- OPTIONAL (If any pip dependency required to install plugin) AND if any apt package needed then (KNOWN issue, it can be handle in future)
```
```
$ cat coap/requirements.txt
aiocoap==0.3
cbor2==4.0.0
```

```
curl -sX POST http://localhost:8081/foglamp/plugins -d '{"url":"http://localhost:5000/download/coap.tar", "format":"tar", "compressed":false, "type": "south", "checksum": "5626d5625ca742435ea58809373573c6"}' | jq
{
  "message": "['Successfully downloaded coap.tar'] and installed"
}
```
**Another Python example**

```
tree mqtt_sparkplug
mqtt_sparkplug
├── __init__.py
├── mqtt_sparkplug.py
├── requirements-mqtt_sparkplug.txt
└── sparkplug_b
    ├── __init__.py
    ├── sparkplug_b_pb2.py
    └── sparkplug_b.py

curl -sX POST http://localhost:8081/foglamp/plugins -d '{"url":"http://localhost:5000/download/mqtt_sparkplug.tar.gz", "type": "south", "format":"tar", "checksum": "77b74584e09fc28467599636e47f3fc5", "compressed": "true"}' | jq
{
  "message": "['Successfully downloaded mqtt_sparkplug.tar.gz'] and installed"
}

```

**And it will save and download under `FOGLAMP_ROOT/data/plugins` and move to in relative the plugins directory.**

# C - Plugin
```
 $ tree Benchmark
Benchmark    (Naming convention should be followed and it will exactly same as required i.e PLUGIN_NAME from  ----- lib<PLUGIN_NAME>.so )
├── libBenchmark.so
└── libBenchmark.so.1
```

```
curl -sX POST http://localhost:8081/foglamp/plugins -d '{"url":"http://localhost:5000/download/c/Benchmark1.tar.gz", "format":"tar", "compressed":true, "type": "south", "checksum": "4015c2dea1cc71dbf70a23f6a203eeb6"}' | jq
{
  "message": "['Successfully downloaded Benchmark1.tar.gz'] and installed"
}

```
```
 $ tree dht11_V2/
dht11_V2/
├── libdht11_V2.so
├── libdht11_V2.so.1
└── requirements.sh      (i.e sudo apt install -y wiringpi) -- KNOWN ISSUE (FOGL-2760)
```

**Pending items:**

~Handling for  notification delivery and rules plugins~
~[ ] .deb format handling~
~- [ ] Remove extra logger which is for debug purpose~
~Add unit tests~


.gitignore
```diff
@@ -14,6 +14,8 @@ data/etc/storage.json
 data/etc/certs/*
 data/var
 data/tmp
+data/scripts
```
ashish-jabble: We missed it in previous work for upload script; so added in this pull request

python/foglamp/services/core/api/plugins/install.py
```diff
@@ -0,0 +1,179 @@
...
+    tar = tarfile.open(_PATH + file_name, mode)
+    _LOGGER.info("Extracted to {}".format(_PATH))
+    tar.extractall(_PATH)
+    _LOGGER.info("Extraction Done!!")
+    # _LOGGER.info(tar.getnames())
+    return tar.getnames()
+
+
+def install_debian(file_name: str):
+    # FIXME: Not working seems like we need to manipulate in /etc/sudoers.d/foglamp file
```
MarkRiddoch: Possibly, or recreate a shell script that will install packages and make it set user id on execution chmod 455 <script>

The same trick can be done to install a tar archive if the FogLAMP is installed via sudo make install. Th make install rule will need to set the ownership and permissions on that script as part of the make install

ashish-jabble: yeah will try the solution which is provided in https://github.com/foglamp/FogLAMP/pull/1494

ashish-jabble:
@MarkRiddoch Please check the C-util program provided by @AmandeepArora in the same pull request. By this util program my problems are resolved. Also please check FOGL-2741 comments.

ashish-jabble:
> Looks good to me. I think we will need to look into external dependencies in the future, but this allows us to start with the functionality we want for adding plugins via the management. We may have find not all plugins can be installed via tar files, but the Debian packages will need to be used.

@MarkRiddoch apt install fixes with `scripts/extras/foglamp.sudoers`

PR #1524, opened 2019-04-15
FOGL-2655: Utility program to extract plugin snapshot tar archive

AmandeepArora:
Setuid bit would be set on its executable.


C/plugins/utils/extract_plugin_snapshot.cpp
```diff
@@ -0,0 +1,48 @@
...
+ */
+int main(int argc, char *argv[])
+{
+	if(argc < 2)
+	{
+		printf("Usage: %s <archive filename>\n", argv[0]);
+		return 1;
+	}
+
+        char *args[]={(char *) "/bin/tar", (char *) "-C", (char *) "/usr/local/foglamp/", (char *) "-xf", (char *) "PLACEHOLDER", NULL};
```
MarkRiddoch: It might not be /usr/local/foglamp. We should also cope with installations in different places that are reflect with FOGLAMP_ROOT environment variable.

AmandeepArora: Yeah, considered that in discussions earlier today. The problem with reading FOGLAMP_ROOT environment variable in this program, is that someone could set this to say "/", set the tar.gz file contents with files to be overwritten in /usr/bin, /bin, or /sbin, etc and this setuid executable is available to modify contents of these critical directories as root user.

One possibility is to not allow executables in tar archive  (except perhaps links with 0777 permissions), or such alternate checks. Will think more about it.

AmandeepArora: @MarkRiddoch - I added code to read FOGLAMP_ROOT env variable and verify foglamp installation at that path before proceeding with archive extraction. Please check.

PR #1537, opened 2019-04-26
FOGL-2751: Run Foglamp and plugins/filters on RHEL 7

AmandeepArora:


requirements.sh
```diff
@@ -31,4 +31,5 @@ sudo apt install -y cmake g++ make build-essential autoconf automake uuid-dev
 sudo apt install -y libtool libboost-dev libboost-system-dev libboost-thread-dev libpq-dev libssl-dev libz-dev
 sudo apt install -y python-dbus python-dev python3-dev python3-pip
 sudo apt install -y sqlite3 libsqlite3-dev
+sudo apt install -y pkg-config
```
praveen-garg: Please open a corresponding pull request for https://github.com/foglamp/foglamp-pkg `control` file `Depends:`  field.

praveen-garg:
@Singhal-Vaibhav  this FOGL-2751-aman branch is  supposed to go into FOGL-2751. I believe we should merge it with sanity testing with deb packaging for ub 16 and RPi and make/make install on RHEL (AWS instance). And do more more aggressive testing - running system tests, manual deb, RHEL checks for the  FOGL-2751 branch (PR) updated with latest develop.

requirements.sh
AmandeepArora: "pkg-config" is required only at compile time by cmake to find other packages, it is not required at run-time. So probably it is not required to add pkg-config to control fie "Depends" field ?

praveen-garg: Yeah. It should be fine then.

tests/unit/C/README.rst
```diff
@@ -19,6 +22,12 @@ Unfortunately this does not install the libraries and a manual build set is requ
 - sudo cmake --build build
 - sudo cp build/libgtest* /usr/lib
 
+Red Hat Enterprise Linux:
+-------------------------
+
+- sudo yum install gtest gtest-devel
```
Singhal-Vaibhav: We need to add the following line before installing gtest:
```
sudo yum install https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
```

AmandeepArora: Updated.

AmandeepArora:
@Vaibhav, that issue is because "scl enable" state is lost during cascading shell spawning via "sudo make install". Massimiliano would take care of this issue in further commits.

PR #1543, opened 2019-04-30
Correct Copyright holder

billehunt:


LICENSE
```diff
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2017 OSIsoft, LLC
+   Copyright 2019 Dianomic Systems
```
MarkRiddoch: Marshall's last comment was it should be Dianomic Systems Inc

I have a script ready to go for all the repositories to make that change, will exclude this one.

PR #1545, opened 2019-05-02
Fogl 2756: specific actions either for debian or redhat

sstefano69:
Changes in the foglamp makefile will be in a different branch.

**tested**
+ rpm creation on Centos
+ Centos - installation
+ RH     - installation

+ full rpm creation on RH - fresh machine
+ Centos - installation
+ RH     - installation

+ Ubuntu - fresh machine - .deb creation
+ Ubuntu - fresh machine - .deb installation


scripts/foglamp
```diff
@@ -19,6 +19,19 @@
 set -e
 #set -x
 
+# RedHat specific
+platform=$([ `cat /etc/os-release | grep -c -i -E "centos|red hat"` -eq 0 ] && echo debian || echo redhat)
```
AmandeepArora: Please update the check here also, if possible.

scripts/foglamp
```diff
@@ -19,6 +19,19 @@
...
+if [[ $platform == "redhat" ]]
+then
+
+	export PATH=/opt/rh/rh-python36/root/usr/bin${PATH:+:${PATH}}
```
AmandeepArora: These 5 steps are already done in /opt/rh/rh-python36/enable. Can also be done via "source scl_source enable rh-python36", rather than using absolute paths in "/opt/rh".

extras/scripts/foglamp.service
```diff
@@ -1,8 +1,13 @@
 #!/bin/sh
 
-# kFreeBSD do not accept scripts as interpreters, using #!/bin/sh and sourcing.
-if [ true != "$INIT_D_SCRIPT_SOURCED" ] ; then
-    set "$0" "$@"; INIT_D_SCRIPT_SOURCED=true . /lib/init/init-d-script
+# Debian specific
+platform=$([ `cat /etc/os-release | grep -c -i -E "centos|red hat"` -eq 0 ] && echo debian || echo redhat)
```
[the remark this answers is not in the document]
sstefano69: solved

scripts/foglamp
sstefano69: switch to use get_platform

scripts/foglamp
sstefano69: the point is that when FogLAMP is started as a service we are using : 

sudo -u ${FOGLAMP_USER} "${FOGLAMP_ROOT}/bin/foglamp" start > /dev/null

and using a bourne - #!/bin/sh

and to solve this situation we have identify that solution with Mark.

sstefano69: I identified the problem and solved - now using - source scl_source enable rh-python36

scripts/common/get_platform.sh
```diff
@@ -0,0 +1,29 @@
...
+__author__="Stefano Simonelli"
+__version__="1.0"
+
+# Identifies the platform on which FogLAMP reside
+# output :
+#          not empty - Centos or RedHat
+#          empty     - Debian/Ubuntu
+get_platform() {
+
+	(lsb_release -ds 2>/dev/null || cat /etc/*release 2>/dev/null | head -n1 || uname -om) | egrep '(Red Hat|CentOS)' || echo ""
```
AmandeepArora: May be we can return the output of this command "(lsb_release -ds 2>/dev/null || cat /etc/*release 2>/dev/null | head -n1 || uname -om)" and let the calling script do the grep for the required pattern. That would be more flexible approach.

sstefano69: solved

PR #1548, opened 2019-05-07
Fix for CentOS / RedHat

pintomax:
Fix for CentOS / RedHat


AmandeepArora:
Please remove "SUDOERS_SRC_RH" for now. And please explore if there is any other alternate approach to make scl enabled at all times. @praveen-garg mentioned that it is possible to do so, but I am not sure if it is desirable and better solution.

pintomax:
@AmandeepArora removed, it was an oversight.
We need to test and merge this. I'll add SQLIte3 changes after that.

I think we can explore an alternative to scl enabled when all the needed work is done.

pintomax:
@Singhal-Vaibhav @AmandeepArora 

I've found this solution, using this branch


[ec2-user@ip-10-0-0-93 FogLAMP]$ sudo -s
[root@ip-10-0-0-93 FogLAMP]# source scl_source enable rh-python36 && make install

pintomax:
This seems OK as well in Makefile:


CMAKE := source scl_source enable rh-python36 && cmake

pintomax:
Added fix:

just type "sudo make install"  where FogLAMP source has been pulled from git

PR #1551, opened 2019-05-10
FOGL-2810: Add support for CentOS 7 in requirements.txt

pintomax:
FOGL-2810: Add support for CentOS 7 in requirements.txt


pintomax:
Maybe forcing system updates, yum check-update or sudo apt update
is something the user doesn't want and we can remove them.

Any opinion?

MarkRiddoch (approved):
Builds when tested on my CentOS. FogLAMP starts after setting Python 3.6 with scl enable, we should really document the Python 3.6 prominently. Passes a few basic sanity tests.

PR #1552, opened 2019-05-10
Fix unit tests in RHEL 7

pintomax:
Fix unit tests in RHEL 7


tests/unit/C/cmake_sqlite/CMakeLists.txt
```diff
@@ -31,6 +31,13 @@ add_library(${PROJECT_NAME} SHARED ${SOURCES} ${COMMON_SOURCES})
 add_definitions(-DSQLITE_SPLIT_READINGS=1)
 add_definitions(-DPLUGIN_LOG_NAME="SQLite 3 in_memory")
 
-target_link_libraries(${PROJECT_NAME} -lsqlite3)
+if(EXISTS ${FOGLAMP_SQLITE3_LIBS})
+	include_directories(${FOGLAMP_SQLITE3_LIBS})
+	target_link_libraries(${PROJECT_NAME} -L"${FOGLAMP_SQLITE3_LIBS}/.libs" -lsqlite3)
```
AmandeepArora: FOGLAMP_SQLITE3_LIBS points to "/tmp/foglamp-sqlite3-pkg/src"? If so, contents of /tmp are lost on each reboot. Or is it getting built every time unit tests are run?

pintomax: SQLITE3 static library in /tmp/foglamp-sqlite3-pkg/src is re-built following steps in requirements.sh

So after reboot that location goes away.

If jenkins PR-checker downloads and compiles the sqlite3 this is not an issue 

Let's wait @Singhal-Vaibhav  feedback on that.

pintomax: Right now this SQLITE3 static library in /tmp/foglamp-sqlite3-pkg/src is only needed by RHEL/Centos

So the RHEL server where PR checker will be installed could easily take care of this

PR #1553, opened 2019-05-10
FOGL-2810: Support for centOS 7 and other fixes

Singhal-Vaibhav:
Fixes:
1. removed sudo from commands, the script needs to be executed with sudo
2. removed update/upgrade. We dont want to upgrade user environment from this script
3. Supports RPI/ Ub 18/16, Centos 7 and RHEL 7. If we try to run on RHEL 8, it gives error
4. No need to run any shell command `scl enable rh-python36 bash` before running foglamp, this script takes care of it.


requirements.sh
```diff
@@ -17,47 +17,44 @@
...
+	yum install -y rsyslog
+	yum install -y openssl-devel
+	yum install -y postgresql-devel
+	yum install -y wget
+	yum install -y zlib-devel
+	yum install -y git
+	yum install -y cmake
+	yum install -y libuuid-devel
+	yum install -y dbus-devel
+	echo "source scl_source enable rh-python36" >> /home/${SUDO_USER}/.bashrc
```
pintomax: The PR looks good.

Just asking wether the permanent change in  /home/${SUDO_USER}/.bashrc is something the user could complain about

Singhal-Vaibhav: If we dont do this, every time user closes his shell terminal and opens up a new, the user needs to run `scl enable rh-python36 bash`

PR #1554, opened 2019-05-10
Initial commit of a check for segfault test

pintomax:
Initial commit of a check for segfault test

Run in Ubuntu first. RHEL needs FOGL-2806


pintomax:
The test fails because a segfault come from one test.

Verify that Jenkins takes all xml files of the tests including the one with segfault, where the failure is described

pintomax:
Use commit 4df3be9 to check the segfault is trapped and all tests are ok

If these solution is ok, other JIRA tasks will come with code that protects possibel segfauls in tests

PR #1558, opened 2019-05-15
requirements.sh run without sh as per new changes in

ashish-jabble:


ashish-jabble:
@Singhal-Vaibhav I ran custom job on [CI](http://35.170.30.120:8080/job/foglamp_nightly_build_ub18/49/console) with given fix and it passes.

PR #1559, opened 2019-05-17
FOGL-2641 - avahi dbus name collision handling issue, does not allow more than 1 FogLAMP instances discoverable over the network 

amarendra-dianomic:


python/foglamp/services/core/server.py
```diff
@@ -732,10 +732,11 @@ def _start_core(cls, loop=None):
...
             # All services are up so now we can advertise the Admin and User REST API's
-            cls.admin_announcer = ServiceAnnouncer(cls._service_name, cls._ADMIN_API_SERVICE, service_server_port,
-                                                   [cls._service_description])
-            cls.user_announcer = ServiceAnnouncer(cls._service_name, cls._USER_API_SERVICE, service_server_port,
-                                                  [cls._service_description])
+            # cls.admin_announcer = ServiceAnnouncer(cls._service_name, cls._ADMIN_API_SERVICE, service_server_port,
+            #                                        [cls._service_description])
+            # cls.user_announcer = ServiceAnnouncer(cls._service_name, cls._USER_API_SERVICE, service_server_port,
+            #                                       [cls._service_description])
+
```
MarkRiddoch: Why are these being removed?

python/foglamp/services/common/service_announcer.py
```diff
@@ -6,52 +6,61 @@
...
 
-    _group = None
-    """ The Avahi group """
+class ServiceAnnouncer:
+    def __init__(self, sname, stype, port, txt):
+        host_name = socket.gethostname()
+        host = self.get_ip()
+        service_name = "_" + sname.lower() + "._tcp.local."
+        desc = {'path': '/~paulsm/'}  # TODO: Change
+        _LOGGER.error(">>>>>>>>>>>>>>>>> %s", service_name)
```
MarkRiddoch: Please do not checkin with this in place.

amarendra-dianomic: Fixed

python/foglamp/services/core/server.py
amarendra-dianomic: Fixed

amarendra-dianomic:
@MarkRiddoch, @praveen-garg feedback changes done. Please check.

python/foglamp/services/common/service_announcer.py
```diff
@@ -6,52 +6,60 @@
...
-    _service_name = None
-    """ The name of the service to advertise """
 
-    _group = None
-    """ The Avahi group """
+class ServiceAnnouncer:
+    def __init__(self, sname, stype, port, txt):
+        host_name = socket.gethostname()
+        host = self.get_ip()
+        service_name = "{}_{}.{}".format(host_name, sname, stype)
```
praveen-garg: can we simpy keep it the name, as passed? hostname we are already putting in for `server`

MarkRiddoch:
Looks like a good way forward.

PR #1561, opened 2019-05-20
FOGL-2780: Update OMF north plugin to handle array data

pintomax:
FOGL-2780: Update OMF north plugin to handle array data

Any unsupported OMF data type will be stripped out by the OMF plugin and not sent to the PI Server.

Unit tests changes/updates will come with next commits


C/plugins/common/omf.cpp
```diff
@@ -1569,3 +1651,24 @@ bool OMF::getCreatedTypes(const string& key)
...
 }
+
+/**
+ * Check whether input Datapoint type is supported by OMF class
+ *
+ * @param    dataPoint		Input data
+ * @return			True is fupported, false otherwise
+ */ 
+
+static bool isTypeSupported(DatapointValue& dataPoint)
```
MarkRiddoch: Please make this a private method of the class.

pintomax: It's used by two classes.

PR #1562, opened 2019-05-20
Name conflict resolution in Service Announcer

amarendra-dianomic:


python/foglamp/services/common/service_announcer.py
```diff
@@ -44,17 +45,32 @@ def __init__(self, sname, stype, port, txt):
...
+                    socket.inet_aton(host),
+                    port,
+                    0,
+                    0,
+                    desc,
+                    "{}.local.".format(host_name),
+                )
+                if zeroconf.get_service_info(info.type, info.name) is not None:
+                    raise NonUniqueNameException
+            except NonUniqueNameException:
```
praveen-garg: allow_name_change =True is there, we don't need this.

amarendra-dianomic:
No longer needed in view of latest changes in FOGL-2641.

PR #1567, opened 2019-05-22
FOGL-2847 syslog file path fixes for RHEL/ CentOS

praveen-garg:
set `/var/log/messages` permissions to 644.


python/foglamp/services/core/api/support.py
```diff
@@ -18,7 +19,12 @@
 
 _FOGLAMP_DATA = os.getenv("FOGLAMP_DATA", default=None)
 _FOGLAMP_ROOT = os.getenv("FOGLAMP_ROOT", default='/usr/local/foglamp')
+
 _SYSLOG_FILE = '/var/log/syslog'
+
+if ('centos' in platform.platform()) or ('redhat' in platform.platform()):
+    _SYSLOG_FILE = '/var/log/message'
```
MarkRiddoch: /var/log/messages

python/foglamp/services/core/support.py
```diff
@@ -30,6 +31,9 @@
 _NO_OF_FILES_TO_RETAIN = 3
 _SYSLOG_FILE = '/var/log/syslog'
 
+if ('centos' in platform.platform()) or ('redhat' in platform.platform()):
+    _SYSLOG_FILE = '/var/log/message'
```
MarkRiddoch: /var/log/messages

python/foglamp/services/core/api/support.py
praveen-garg: Thanks, fixed!

python/foglamp/services/core/support.py
praveen-garg: Thanks, fixed!

PR #1568, opened 2019-05-23
FOGL-2847 syslog file path fixes for RHEL/ CentOS (v1.6.0)

praveen-garg:


praveen-garg:
done via cherry-pick 1dffabd7

PR #1570, opened 2019-05-23
FOGL-2783: Extend directory paths from where Foglamp picks plugins

AmandeepArora:
Tested south & north plugins, filters and notification delivery/rules plugins with help from Ashish. Ashish's changes are in PR 1565 under FOGL-2805.


C/services/common/plugin_manager.cpp
```diff
@@ -41,14 +41,58 @@ PluginManager::PluginManager()
...
+
 /**
  * Load a given plugin
  */
 PLUGIN_HANDLE PluginManager::loadPlugin(const string& name, const string& type)
 {
 PluginHandle *pluginHandle = NULL;
 PLUGIN_HANDLE hndl;
-char          buf[128];
+char		buf[128];
```
MarkRiddoch: We should probably make this MAXPATHLEN now and not just 128

PR #1576, opened 2019-06-04
FOGL-2738 notification instance update error handling fixes

praveen-garg:


tests/system/python/api/test_notification.py
```diff
@@ -173,8 +173,6 @@ def test_create_valid_notification_instance(self, foglamp_url):
         assert 2 == len(jdoc['rules'])
 
     @pytest.mark.parametrize("test_input, expected_error", [
-        pytest.param({"name": "Test8"}, '400: Name update is not allowed.', marks=pytest.mark.xfail(reason="FOGL-2673")),
```
praveen-garg: name is marked as readonly property, now.

PR #1580, opened 2019-06-11
FOGL-2882: removed PQ from sqlite cmake

sstefano69:


praveen-garg:
@sstefano69  Thanks, Running the tests now..

PR #1586, opened 2019-06-18
FOGL-2883

Singhal-Vaibhav:
- Temporary skipped test TestE2eExprPi test. This needs to be fixed for consistency
- Increased timeout in Scaleset set

### DIFF
--- a/tests/system/python/e2e/test_e2e_expr_pi.py
+++ b/tests/system/python/e2e/test_e2e_expr_pi.py
@@ -31,6 +31,7 @@ SVC_NAME = "Expr #1"
 ASSET_NAME = "Expression"
 
 
+@pytest.mark.skip(reason="Flaky test, to be fixed")
 class TestE2eExprPi:
     def get_ping_status(self, foglamp_url):
         _connection = http.client.HTTPConnection(foglamp_url)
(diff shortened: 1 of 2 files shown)


tests/system/python/e2e/test_e2e_expr_pi.py
```diff
@@ -31,6 +31,7 @@
 ASSET_NAME = "Expression"
 
 
+@pytest.mark.skip(reason="FOGL-2883: Flaky test, to be fixed")
```
praveen-garg: Can we skip based on Redhat platform condition?

praveen-garg: and more to it, can we skip just sent verification instead of full test. I'm not sure if this is not able to send data or verify sent

Singhal-Vaibhav: fail randomly on ub16/18 as well. Nothing specific to a particular environment

praveen-garg: Okay

tests/system/python/e2e/test_e2e_expr_pi.py
```diff
@@ -151,6 +151,7 @@ def _verify_egress(self, read_data_from_pi, pi_host, pi_admin, pi_passwd, pi_db,
         assert "Expression" in data_from_pi
         assert isinstance(data_from_pi["name"], list)
         assert isinstance(data_from_pi["Expression"], list)
-        assert "value" in data_from_pi["name"]
+        # TODO: FOGL-2883: Test fails randomly in below assertion needs to be fixed
+        # assert "value" in data_from_pi["name"]
         # FOGL-2438 values like tan(45) = 1.61977519054386 gets truncated to 1.6197751905 with ingest
         assert 1.6197751905 in data_from_pi["Expression"]
```
praveen-garg: will this also not fail?

Singhal-Vaibhav: this fails too, commented out. Thanks

PR #1589, opened 2019-06-19
FOGL-2876: Add a modbus simulator in CI network to be accessed by modbus system test

Singhal-Vaibhav:

### DIFF
--- a/tests/system/python/e2e/test_e2e_modbus_c_pi.py
+++ b/tests/system/python/e2e/test_e2e_modbus_c_pi.py
@@ -63,7 +63,7 @@ class TestE2EModbusCPI:
             add_south: Fixture that adds a south service with given configuration with enabled or disabled mode
             remove_directories: Fixture that remove directories created during the tests
         """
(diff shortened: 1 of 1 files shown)


tests/system/python/e2e/test_e2e_modbus_c_pi.py
```diff
@@ -64,7 +64,8 @@ def start_south_north(self, reset_and_start_foglamp, add_south, remove_directori
             remove_directories: Fixture that remove directories created during the tests
         """
         self.check_connect(modbus_host, modbus_port)
-        cfg = {"protocol": {"value": "TCP"}, "asset": {"value": ""}, "port": {"value": "502"},
+        cfg = {"protocol": {"value": "TCP"}, "asset": {"value": ""}, "address": {"value": modbus_host},
+               "port": {"value": "{}".format(modbus_port)},
```
praveen-garg: `"{}".format(modbus_port)` could just  be `str(modbus_port)`.  Though we are good as is!

PR #1597, opened 2019-06-28
FOGL-2070 password type configuration secret text in REST API side ONLY

ashish-jabble:
**Response:**
1) Category having password type returns **"****"** at ADMIN API side

```
 $ curl -sX GET http://localhost:8081/foglamp/category/C1 | jq
{
  "p2": {
    "default": "P2",
    "type": "password",
    "value": "****",
    "description": "P2"
  },
  "p1": {
    "default": "P1",
    "type": "password",
    "value": "****",
    "description": "P1"
  },
  "test1": {
    "default": "",
    "type": "script",
    "value": "",
    "file": "",
    "description": "d"
  }
}

```
2) Category having password type returns **plain text** when microservice management API

```
$ curl -sX GET http://localhost:35359/foglamp/service/category/C1 | jq
{
  "p2": {
    "default": "P2",
    "value": "foglamp",
    "type": "password",
    "description": "P2"
  },
  "p1": {
    "default": "P1",
    "value": "FogLAMP",
    "type": "password",
    "description": "P1"
  },
  "test1": {
    "file": "",
    "default": "",
    "value": "",
    "type": "script",
    "description": "d"
  }
}

```

Fixed for below endpoints
- [x] GET /foglamp/category/{category_name}
- [x]  GET  /foglamp/category/{category_name}/{config_item}
- [x]  PUT  /foglamp/category/{category_name}/{config_item}
- [x]  DELETE  /foglamp/category/{category_name}/{config_item}/value
- [x] POST /foglamp/category
- [x] PUT /foglamp/category/{category_name} -- fyi, bulk update only exists at Admin API side

### DIFF
--- a/python/foglamp/services/core/api/configuration.py
+++ b/python/foglamp/services/core/api/configuration.py
@@ -4,11 +4,12 @@
 # See: http://foglamp.readthedocs.io/
 # FOGLAMP_END
 
-
+import copy
 from aiohttp import web
 import binascii
 import urllib.parse
 import os
+from typing import Dict
 
 from foglamp.services.core import connect
 from foglamp.common.configuration_manager import ConfigurationManager, _optional_items
@@ -94,6 +95,11 @@ async def get_category(request):
     if category is None:
         raise web.HTTPNotFound(reason="No such Category found for {}".format(category_name))
 
+    try:
+        request.is_core_mgt
+    except AttributeError:
+        category = hide_password(category)
+
     return web.json_response(category)
 
 
@@ -143,6 +149,10 @@ async def create_category(request):
         if data.get('children'):
             r = await cf_mgr.create_child_category(category_name, data.get('children'))
             result.update(r)
+        try:
+            request.is_core_mgt
+        except AttributeError:
+            result['value'] = hide_password(result['value'])
     except (KeyError, ValueError, TypeError) as ex:
         raise web.HTTPBadRequest(reason=str(ex))
     except LookupError as ex:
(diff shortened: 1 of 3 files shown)


python/foglamp/services/core/api/configuration.py
```diff
@@ -596,3 +625,15 @@
...
         return web.json_response(result)
+
+
+def hide_password(config: dict) -> Dict:
+    new_config = copy.deepcopy(config)
+    try:
+        for k, v in new_config.items():
+            if v['type'] == 'password':
+                v['value'] = "****"
+    except TypeError:
```
praveen-garg: This is to handle the case of single config item vs category with multiple config items?!

tests/unit/python/foglamp/services/core/api/test_configuration.py
```diff
@@ -207,16 +210,20 @@ def reset_singleton(self):
                 assert 200 == resp.status
                 r = await resp.text()
                 json_response = json.loads(r)
+                if hide_password:
+                    result['value'] = "****"
```
praveen-garg: why password hide is manual process here, is this not the expected result? and to be done by the executed code.

ashish-jabble: `expected result Vs Actual result` - There is no manipulation on actual result, it's the way to modify the data setup for tests and there are n number of way to deal it.

python/foglamp/services/core/api/configuration.py
ashish-jabble: yeah, for both cases the instance type is dict but with different structures.

tests/unit/python/foglamp/services/core/api/test_configuration.py
praveen-garg: Got it. I missed to read. `@pytest.mark.parametrize("result` this result is expectedResult. I got it confused with manipulation on actual result.

PR #1601, opened 2019-06-29
FOGL-2612: Add system API tests for auth

Singhal-Vaibhav:
Tests positive scenarios with authentication set to mandatory.
Test Results:
===== 18 passed in 38.37 seconds =====

### DIFF
--- a/tests/system/python/api/test_authentication.py
+++ b/tests/system/python/api/test_authentication.py
@@ -26,229 +26,229 @@ TOKEN = None
 
 @pytest.fixture
 def change_to_auth_mandatory(reset_and_start_foglamp, foglamp_url, wait_time):
-        # Wait for foglamp server to start
-        time.sleep(wait_time)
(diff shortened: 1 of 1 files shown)


tests/system/python/api/test_authentication.py
```diff
@@ -0,0 +1,212 @@
...
+                assert "403: Forbidden" == r
+
+                # Delete User
+                conn.request("DELETE", "/foglamp/admin/2/delete", headers={"authorization": _token})
+                r = conn.getresponse()
+                assert 403 == r.status
+                r = r.read().decode()
+                assert "403: Forbidden" == r
+
+        def test_login_with_certificate(self, foglamp_url):
```
praveen-garg: It would be great to have a test_admin_login_with_certificate, and rename this to test_user_login_with_certificate.

And create some cert at runtime and use that to test valid cert login

PR #1610, opened 2019-07-16
FOGL-3000  deprecated plugin mechanism added for both C and python plugins including their unit tests

ashish-jabble:
**Fixed items:**
- [x] deprecated mechanism added for both C & python plugins
- [x] C -code changes (SP_DEPRECATED bit mask added & flag key now return in C-utility program to get plugin info) and for python plugins if flag key exists then only action will be taken.
**'flag' KEY has taken for consistency on Both sides**
- [x] new unit tests added for plugin discovery and utils function for bit position check

**More info**
- [x] Have tested with all plugin type supported (south, north, filter and notify delivery and rules)
- [x] Also tested with FOGLAMP_PLUGIN_PATH environment set
- [x] No regression in system API tests

**Testing Notes:**
1) Changes in sinusoid.py file
```
@@ -123,7 +123,8 @@ def plugin_info():
         'mode': 'poll',
         'type': 'south',
         'interface': '1.0',
-        'config': _DEFAULT_CONFIG
+        'config': _DEFAULT_CONFIG,
+        'flag': utils.DEPRECATED_BIT_MASK_VALUE
     }


OR

'flag': 32
```

2) Changes in plugin.cpp for foglamp-south-csv-async
```
 @@ -42,7 +42,7 @@ extern "C" {
 static PLUGIN_INFORMATION info = {
        PLUGIN_NAME,              // Name
        VERSION,                  // Version
-       SP_ASYNC,                 // Flags
+       SP_ASYNC | SP_DEPRECATED,                 // Flags
        PLUGIN_TYPE_SOUTH,        // Type
        "1.0.0",                  // Interface version
        CONFIG                    // Default configuration
```
3) Changes in plugin.cpp for foglamp-south-csv

```
@@ -40,7 +40,7 @@ extern "C" {
 static PLUGIN_INFORMATION info = {
        PLUGIN_NAME,              // Name
        VERSION,                  // Version
-       0,                        // Flags
+       SP_DEPRECATED,                // Flags
        PLUGIN_TYPE_SOUTH,        // Type
        "1.0.0",                  // Interface version
        CONFIG                    // Default configuration
```

So When I hit `curl -sX GET http://localhost:8081/foglamp/plugins/installed`, **deprecated plugins are NOT in response and logged warn in syslogs**

**Syslogs**

```
WARNING: plugin_discovery: foglamp.common.plugin_discovery: "sinusoid" plugin is deprecated
WARNING: plugin_discovery: foglamp.common.plugin_discovery: "Csv" plugin is deprecated
WARNING: plugin_discovery: foglamp.common.plugin_discovery: "CSV-Async" plugin is deprecated
```

### DIFF
--- a/C/services/common/include/plugin_api.h
+++ b/C/services/common/include/plugin_api.h
@@ -16,7 +16,8 @@ typedef struct {
         unsigned int	options;
         const char	*type;
         const char	*interface;
-	const char	*config;
+	    const char	*config;
+	    bool        deprecated;
 } PLUGIN_INFORMATION;
  
 typedef struct {
--- a/C/services/common/include/plugin_api.h
+++ b/C/services/common/include/plugin_api.h
@@ -16,8 +16,7 @@ typedef struct {
         unsigned int	options;
         const char	*type;
         const char	*interface;
-	    const char	*config;
-	    bool        deprecated;
+	const char	*config;
 } PLUGIN_INFORMATION;
  
 typedef struct {
@@ -36,7 +35,8 @@ typedef void * PLUGIN_HANDLE;
 #define SP_ASYNC	0x0004
 #define SP_PERSIST_DATA	0x0008
 #define SP_INGEST	0x0010
- 
+#define SP_DEPRECATED 0x0020
+
 /**
  * Plugin types
  */
--- a/python/foglamp/common/plugin_discovery.py
+++ b/python/foglamp/common/plugin_discovery.py
@@ -10,6 +10,8 @@ import os
 from foglamp.common import logger
 from foglamp.services.core.api import utils
 from foglamp.services.core.api.plugins import common
+from foglamp.plugins.common import utils as api_utils
+
 
 __author__ = "Amarendra K Sinha, Ashish Jabble"
 __copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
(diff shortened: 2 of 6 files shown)


C/services/common/include/plugin_api.h
```diff
@@ -16,7 +16,8 @@ typedef struct {
         unsigned int	options;
         const char	*type;
         const char	*interface;
-	const char	*config;
+	    const char	*config;
+	    bool        deprecated;
```
MarkRiddoch: This was not meant to change the plugin_info structure. The Jira was to add a new flag value to the flags bit mask.

ashish-jabble: What will be the value for flags bit mask? And what about for python plugins as there is NO Flag key under plugin_info routine.

MarkRiddoch: Starting with the Python side. Here the approach you have taken is fine as it is a dict and inherently flexible.

On the C side, the current defined flag bits are
#define SP_COMMON       0x0001
#define SP_READINGS     0x0002
#define SP_ASYNC        0x0004
#define SP_PERSIST_DATA 0x0008
#define SP_INGEST       0x0010

in the file C/services/common/include/plugin_api.h

I would suggest we use the bit 

#define SP_DEPRECATED 0x0020

ashish-jabble:
@MarkRiddoch feedback changes done. Kindly have a look.

python/foglamp/common/plugin_discovery.py
```diff
@@ -10,6 +10,8 @@
 from foglamp.common import logger
 from foglamp.services.core.api import utils
 from foglamp.services.core.api.plugins import common
+from foglamp.plugins.common import utils as api_utils
```
praveen-garg: This should be as common_utils. 

We already have `from foglamp.services.core.api import utils` above

tests/unit/python/foglamp/common/test_plugin_discovery.py
```diff
@@ -13,6 +13,7 @@
 from foglamp.common.plugin_discovery import PluginDiscovery, _logger
 from foglamp.services.core.api import utils
 from foglamp.services.core.api.plugins import common
+from foglamp.plugins.common import utils as api_utils
```
praveen-garg: as above

PR #1617, opened 2019-07-22
FOGL-2934 plugin update API

ashish-jabble:
`curl -X PUT http://localhost:8081/foglamp/plugins/south/sinusoid/update`

- [x] https://github.com/foglamp/foglamp-north-http/pull/27
- [x] https://github.com/foglamp/foglamp-south-http/pull/31
- [x] https://github.com/foglamp/foglamp-north-http-c/pull/25


ashish-jabble:
@MarkRiddoch I believe here `{pluginName}` means directory of a plugin we installed. So in line with that `{plugin_type}` param needed in endpoint.
For example: `curl -X PUT http://localhost:8081/foglamp/plugins/south/sinusoid/update`

Also there is inconsistency in naming of plugins between  _installed dir_ and _package name_

| installed_plugin_dir  | package name | Action needed
| ------------- | ------------- | ------------- |
| http_south  | foglamp-south-http  | Y
| http_north | foglamp-north-http | Y
| wind_turbine | foglamp-south-wind-turbine | N
| mqtt_sparkplug | foglamp-south-mqtt-sparkplug | N
| ModbusC (libModbusC.so) | foglamp-south-modbus | Y
| FlirAX8 (libFlirAX8.json) | foglamp-south-flirax8 | N
| HttpNorthC (libHttpNorthC.so) | foglamp-north-httpc | Y


**One possible solution** for this 
1) Fix package naming on individual repos 
a) for given python plugins we need to fix as same for `wind turbine and mqtt_sparkplug` (**underscore** in installed dir and **hyphen** in packaging). So that internally we will manage easily for update plugin API.
b) Not sure about C-plugins how could we manage? Seems like we need to fix package name for most of the repos (target only south/north type first as plugin update API support only for them)

Kindly advise!

**NOTE:**
1) For endpoint `curl -X GET http://localhost:8081/foglamp/plugins/available` we used 
`sudo apt list` command so package name always returns in lowercase; irrespective of name defined in the configured repo.
2) Except `2` python plugins (i.e south-http, north-http) we are okay, but C-plugins are many with above problem
3) Don't we need `service update` API ? But with separate JIRA ticket.

ashish-jabble:
@MarkRiddoch Please have a look on updated check-in changes. Also fixed inconsistency package naming for python plugins ONLY.
https://github.com/foglamp/foglamp-north-http/pull/27
https://github.com/foglamp/foglamp-south-http/pull/31

MarkRiddoch:
@ashish-jabble For the C plugins we should change the name of the directory, and the library that is built to drop the unnessary extra bits, e.g. HttpNorthC (libHttpNorthC.so) does need the north part, it's installed in plugins/north so it is clear what it is. The package can still be called foglamp-north-httpc but the installed directory should be https and the library libhttpc.so.

I think we should only keep the C part if there is also an existing Python plugin that we need to differentiate. Otherwise it is again not really carrying any useful information.

I would say we lower case all the package names, so if with have libXyZ.so in directory XyZ then the package is foglamp-...-xyz

This is going to break a few people that might have already installed packages with the old name, but I think we need to do it for this and also the other issue with the architecture. Maybe we can even have checks for consistency in the packaging scripts.

ashish-jabble:
> The package can still be called foglamp-north-httpc but the installed directory should be https and the library libhttpc.so.

https://github.com/foglamp/foglamp-north-http-c/pull/25

Keeping that fact if someone has already installed this plugin and then update. Now they will see two http north plugins (HttpNorthC & httpc) in plugin discovery as the old `libHttpNorthC` is still there (and related stuff if any)

python/foglamp/services/core/api/plugins/update.py
```diff
@@ -0,0 +1,144 @@
...
+    pkg_mgt = 'yum' if 'centos' in _platform or 'redhat' in _platform else 'apt'
+    _PATH = _FOGLAMP_DATA + '/plugins/' if _FOGLAMP_DATA else _FOGLAMP_ROOT + '/data/plugins/'
+
+    stdout_file_path = _PATH + "output.txt"
+    if not os.path.exists(_PATH):
+        os.makedirs(_PATH)
+
+    cmd = "sudo {} update > {} 2>&1".format(pkg_mgt, stdout_file_path)
+    ret_code = os.system(cmd)
+    msg = ""
```
ashish-jabble: We should catch reason for above `cmd` as well

python/foglamp/services/core/api/plugins/update.py
```diff
@@ -0,0 +1,144 @@
...
+        _type = _type.lower()
+        if _type not in ['north', 'south']:
+            raise ValueError("Invalid plugin type. Must be 'north' or 'south'")
+
+        # Check requested plugin name is installed or not
+        installed_plugins = PluginDiscovery.get_plugins_installed(_type, False)
+        installed_plugin_name = [p_name["name"] for p_name in installed_plugins]
+        if name not in installed_plugin_name:
+            raise KeyError("{} plugin is not yet installed. So update is not possible.".format(name))
+
```
ashish-jabble: As per FOGL-3034, we should stop updating plugin only when category named `Installation` whose config item `upgradeOnInstall` value is false.
For above we don't want to hold this up, will be taken care itself in FOGL-3034

PR #1625, opened 2019-07-26
FOGL-3076 Add optional throttle mechanism to poll ingess rate

MarkRiddoch:


C/services/south/south.cpp
```diff
@@ -673,3 +711,61 @@ int SouthService::createTimerFd(struct timeval rate)
...
+	}
+	double desired = m_desiredRate.tv_sec + ((double)m_desiredRate.tv_usec / 1000000);
+	gettimeofday(&now, NULL);
+	timersub(&now, &m_lastThrottle, &res);
+	if (m_ingest->queueLength() > m_highWater && res.tv_sec > SOUTH_THROTTLE_DOWN_INTERVAL)
+	{
+		double rate = m_currentRate.tv_sec + ((double)m_currentRate.tv_usec / 1000000);
+		rate *= (1.0 + ((double)SOUTH_THROTTLE_PERCENT / 100.0));
+		m_currentRate.tv_sec = (long)rate;
+		m_currentRate.tv_usec = (rate - m_currentRate.tv_sec) * 1000000;
```
AmandeepArora: If rate is initially say 1.9 secs, then rate becomes 2.09. Then at line 738, m_currentRate.tv_usec becomes more than 1 sec. There may be a need to normalize m_currentRate structure.

C/services/south/south.cpp
```diff
@@ -673,3 +711,61 @@ int SouthService::createTimerFd(struct timeval rate)
...
+	else if (m_throttled && m_ingest->queueLength() < m_lowWater && res.tv_sec > SOUTH_THROTTLE_UP_INTERVAL)
+	{
+		// We are current throttle back but the queue is below the low water mark
+		timersub(&m_desiredRate, &m_currentRate, &res);
+		if (res.tv_sec != 0 || res.tv_usec != 0)
+		{
+			double rate = m_currentRate.tv_sec + ((double)m_currentRate.tv_usec / 1000000);
+			rate *= (1.0 - ((double)SOUTH_THROTTLE_PERCENT / 100.0));
+			m_currentRate.tv_sec = (long)rate;
+			m_currentRate.tv_usec = (rate - m_currentRate.tv_sec) * 1000000;
```
AmandeepArora: normalization may be required here also.

C/services/south/south.cpp
```diff
@@ -356,6 +374,7 @@ void SouthService::start(string& coreAddress, unsigned short corePort)
 						pollCount += (int) vec->size();
 						delete vec; 	// each reading object inside vector has been allocated on heap and moved to Ingest class's internal queue
 					}
+					throttlePoll();
```
AmandeepArora: This would be called on each poll interval? That could be a lot for higher rates. May be we can call this less frequently? May be every 5 seconds (Greatest common divisor of SOUTH_THROTTLE_DOWN_INTERVAL and SOUTH_THROTTLE_UP_INTERVAL)

PR #1627, opened 2019-07-30
FOGL-3081 update source repo list configuration before fetch or install available packages

ashish-jabble:
As discussed on friday scrum, we should add `sudo apt update` before fetching and on installing available packages.
1)  curl -X GET http://localhost:8081/foglamp/plugins/available
2) curl -sX POST http://localhost:8081/foglamp/plugins -d '{"format":"repository", "name": "foglamp-south-sinusoid"}'

**Use case:**
nightly [repo](http://archives.dianomic.com/nightly/ubuntu1804/x86_64/) gets updated over night

In case of any failure REST API throws a bad request 400 with an error message for below scenarios

1. Any network issues with configured source/repo list
2. On incorrect sources/repo list URL
3. Either  `sudo apt update` or `sudo apt list | grep foglamp`  failed which is used internally in code, now it will capture the standard output and standard error of the apt or or yum command

**NOTE:** As GUI dialog disappears with error messge after some time. So we have another story to store persistent log history and this will be taken care in FOGL-3032 and will probably saved under path `FOGLAMP_ROOT/data/plugins/`

Also as discussed with @praveen-garg filed a JIRA story FOGL-3086 for adding audit trail information on install/update package

### DIFF
--- a/python/foglamp/services/core/api/plugins/common.py
+++ b/python/foglamp/services/core/api/plugins/common.py
@@ -117,7 +117,7 @@ def load_and_fetch_c_hybrid_plugin_info(plugin_name: str, is_config: bool, plugi
 
 
 def fetch_available_packages(package_type: str = "") -> list:
-    plugins = []
+    log_output = []
     plugin_dir = '/plugins/'
     _PATH = _FOGLAMP_DATA + plugin_dir if _FOGLAMP_DATA else _FOGLAMP_ROOT + '/data{}'.format(plugin_dir)
     stdout_file_name = "output.txt"
@@ -144,12 +144,12 @@ def fetch_available_packages(package_type: str = "") -> list:
     with open("{}".format(stdout_file_path), 'r') as fh:
         for line in fh:
             line = line.rstrip("\n")
-            plugins.append(line)
+            log_output.append(line)
 
     # Remove stdout file
     os.remove(stdout_file_path)
 
     if ret_code != 0:
-        raise ValueError(plugins)
+        raise ValueError(log_output)
 
-    return plugins
+    return log_output
--- a/python/foglamp/services/core/api/plugins/common.py
+++ b/python/foglamp/services/core/api/plugins/common.py
@@ -123,36 +123,36 @@ def fetch_available_packages(package_type: str = "") -> list:
     _PATH = _FOGLAMP_DATA + plugin_dir if _FOGLAMP_DATA else _FOGLAMP_ROOT + '/data{}'.format(plugin_dir)
     stdout_file_name = "output.txt"
     stdout_file_path = "/{}/{}".format(_PATH, stdout_file_name)
-
     if not os.path.exists(_PATH):
         os.makedirs(_PATH)
 
     _platform = platform.platform()
-
     pkg_type = "" if package_type is None else package_type
-    if 'centos' in _platform or 'redhat' in _platform:
-        cmd = "sudo yum list available foglamp-{}\* | grep foglamp | cut -d . -f1 > {} 2>&1".\
-            format(pkg_type, stdout_file_path)
-    else:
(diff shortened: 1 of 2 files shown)


python/foglamp/services/core/api/plugins/common.py
```diff
@@ -123,36 +123,36 @@ def fetch_available_packages(package_type: str = "") -> list:
...
-    subprocess.run([cmd], stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
+    if ret_code != 0:
+        # Remove stdout file in case of error
+        os.remove(stdout_file_path)
+        raise ValueError(plugins)
+
+    # Remove stdout file in case of success
+    os.remove(stdout_file_path)
 
     return plugins
```
praveen-garg: is this the installation_logs?

python/foglamp/services/core/api/plugins/common.py
```diff
@@ -123,36 +123,36 @@ def fetch_available_packages(package_type: str = "") -> list:
...
-    arg2 = '/data{}'.format(plugin_dir)
-    cmd = "{} rm {}{}".format(arg1, arg2, stdout_file_name)
-    subprocess.run([cmd], stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
+    if ret_code != 0:
+        # Remove stdout file in case of error
+        os.remove(stdout_file_path)
+        raise ValueError(plugins)
+
+    # Remove stdout file in case of success
+    os.remove(stdout_file_path)
```
praveen-garg: We are removing in any case? I see it in `if ret_code != 0:` block as well

python/foglamp/services/core/api/plugins/common.py
```diff
@@ -123,36 +123,36 @@ def fetch_available_packages(package_type: str = "") -> list:
...
-            format(pkg_type, stdout_file_path)
+    pkg_mgt = 'yum' if 'centos' in _platform or 'redhat' in _platform else 'apt'
 
+    cmd = "sudo {} update > {} 2>&1".format(pkg_mgt, stdout_file_path)
     ret_code = os.system(cmd)
-    if ret_code != 0:
-        raise ValueError
+    # sudo apt/yum -y install only happens when update is without any error
+    if ret_code == 0:
+        if 'centos' in _platform or 'redhat' in _platform:
```
praveen-garg: optimized check is ` if pkg_mgt == 'yum'`

python/foglamp/services/core/api/plugins/common.py
ashish-jabble: yeah `apt/yum list` standard output will write first in this file and then return in API response.

python/foglamp/services/core/api/plugins/common.py
ashish-jabble: yes package list execution only happens if `sudo apt update` executes without any error, otherwise it will stop here

python/foglamp/services/core/api/plugins/install.py
```diff
@@ -246,20 +246,29 @@ def copy_file_install_requirement(dir_files: list, plugin_type: str, file_name:
...
     msg = ""
-    with open("{}".format(_FOGLAMP_ROOT + stdout_file_path), 'r') as fh:
+    plugin_dir = '/plugins/'
+    stdout_file_name = "output.txt"
+    stdout_file_path = "/{}/{}".format(_PATH, stdout_file_name)
+
+    if not os.path.exists(_PATH):
+        os.makedirs(_PATH)
+
+    cmd = "sudo {} update > {} 2>&1".format(pkg_mgt, stdout_file_path)
```
ashish-jabble: `update` command is not needed here as we are calling [fetch_available_package](https://github.com/foglamp/FogLAMP/pull/1627/files#diff-fdf85c87517aa764a4ae1b7bf9e8e427R75) before plugin install to ensure that requested plugin is not available for the given repository.

PR #1629, opened 2019-07-31
FOGL-3032 package error handling | FOGL-3033 package logs

ashish-jabble:
- [x] fetch_available service and plugins packages
- [x] install
- [x] update
- [x] FOGL-3033
- [x] Unit tests fixes and new ones

**NOTE:** -- 
~Response return without **".log"** extensions for log files in fetch, install, update endpoints ONLY. For package logs it will remain as it is~


python/foglamp/services/core/api/package_log.py
```diff
@@ -0,0 +1,72 @@
...
+    for root, dirs, files in os.walk(_get_logs_dir()):
+        if str(file_name) not in files:
+            raise web.HTTPNotFound(reason='{} file not found'.format(file_name))
+
+    file_content = []
+    with open("{}{}".format(_get_logs_dir(), file_name), 'r') as fh:
+        for line in fh:
+            line = line.rstrip("\n")
+            file_content.append(line)
+    return web.json_response({"result": file_content})
```
praveen-garg: Please return file Response, so that we have the flexibility to show it any of the required way a_ browser new tab b_ download c_ parse content and show preview / in right pane under logs. This will also remove the overhead of open and parse file, and convert to JSON.

a_ will be a good option when alert will appear with link in GUI.

/cc @monika-Dianomic

python/foglamp/services/core/api/plugins/common.py
```diff
@@ -133,23 +131,57 @@ def fetch_available_packages(package_type: str = "") -> list:
...
+
+    # Create empty log file name
+    open(_PATH + log_file_name, "w").close()
+
+    # A maximum of _NO_OF_FILES_TO_RETAIN log files will be maintained.
+    # When it exceeds the limit the very first log file will be removed on the basis of creation time
+    files = glob.glob("{}*.log".format(_PATH))
+    files.sort(key=os.path.getctime)
+    if len(files) >= _NO_OF_FILES_TO_RETAIN:
+        for f in files[:-(_NO_OF_FILES_TO_RETAIN - 1)]:
```
praveen-garg: I think it should be:

```
 if len(files) > _NO_OF_FILES_TO_RETAIN:
        for f in files[:-(_NO_OF_FILES_TO_RETAIN)]:

```

python/foglamp/services/core/api/package_log.py
```diff
@@ -0,0 +1,73 @@
...
+
+    :Example:
+        curl -sX GET http://localhost:8081/foglamp/package/log
+    """
+    # Get logs directory path
+    logs_root_dir = _get_logs_dir()
+    found_files = []
+
+    for root, dirs, files in os.walk(logs_root_dir):
+        found_files = [f for f in files if f.endswith(valid_extension)]
```
praveen-garg: Please return a list of objects with { timestamp, plugin/service name  or "update", logFile }

python/foglamp/services/core/api/package_log.py
```diff
@@ -0,0 +1,73 @@
...
+
+    return web.json_response({"logs": found_files})
+
+
+async def get_log_by_name(request: web.Request) -> web.FileResponse:
+    """ GET for a particular log file will return the content of the log file.
+
+    :Example:
+        a) Download file
+        curl -O http://localhost:8081/foglamp/package/log/190802-11-45-28-foglamp-south-sinusoid.log
```
praveen-garg: for update only action, I think we should name it `190802-11-45-28-update.log`

PR #1630, opened 2019-08-01
FOGL-3026: move pyobject_reading_parser.cpp under common-plugin-inter…

pintomax:
…faces/python/

FOGL-3026: move pyobject_reading_parser.cpp under common-plugin-interfaces/python/


pintomax:
This is a pre-requisite before adding filter-plugin-interfaces support

PR #1634, opened 2019-08-06
FOGL-3026: allow Python filter to be loaded when South plugin module …

pintomax:
…is a Python one

FOGL-3026: allow Python filter to be loaded when South service plugin module is a Python one

Note: currenlty only one Python Filter can be added to filters pipeline, no matter if South service plugin is C++ or Python.


AmandeepArora:
Looks fine.

C/services/common-plugin-interfaces/python/include/python_plugin_common_interface.h
```diff
@@ -10,15 +10,18 @@
...
-extern PyObject* pModule;
+// This is the map of Python object initialised in:
+// South, Notification, Filter  plugin interfaces
+// Note: map allocation is done in PluginInterfaceInit of South or Filter or Notification
+map<string, PyObject*> *pythonModules = NULL;
 
-// PluginName set by PluginInterfaceInit
-extern string sPluginName;
+// Global variable gPluginName set by PluginInterfaceInit
+string gPluginName;
```
AmandeepArora: Please add a comment, what this string would hold.

C/services/common-plugin-interfaces/python/include/python_plugin_common_interface.h
```diff
@@ -293,18 +317,21 @@ static PLUGIN_HANDLE plugin_init_fn(ConfigCategory *config)
 static void plugin_reconfigure_fn(PLUGIN_HANDLE* handle,
 				  const std::string& config)
 {
-	if (!pModule)
+	// Look for Python module for gPluginName key
+	auto it = pythonModules->find(gPluginName);
```
AmandeepArora: pythonModules check against "NULL" is required here?

C/services/common-plugin-interfaces/python/include/python_plugin_common_interface.h
```diff
@@ -413,31 +440,34 @@ static void logErrorMessage()
  */
 static void plugin_shutdown_fn(PLUGIN_HANDLE handle)
 {
-	if (!pModule)
+	// Look for Python module for gPluginName key
+	auto it = pythonModules->find(gPluginName);
```
AmandeepArora: Check pythonModules for null.

C/services/filter-plugin-interfaces/python/python_plugin_interface.cpp
```diff
@@ -52,22 +56,24 @@ extern PyObject* createReadingsList(const vector<Reading *>& readings);
  */
 void filter_plugin_ingest_fn(PLUGIN_HANDLE handle, READINGSET *data)
 {
-	if (!pModule)
+	auto it = pythonModules->find(fPluginName);
```
AmandeepArora: Same check against NULL here also

C/services/filter-plugin-interfaces/python/python_plugin_interface.cpp
```diff
@@ -221,8 +229,10 @@ PLUGIN_HANDLE filter_plugin_init_fn(ConfigCategory* config,
  */
 void *PluginInterfaceInit(const char *pluginName, const char * pluginPathName)
 {
-        // Set plugin name
-        sPluginName = pluginName;
+        // Set plugin name for common-interfaces
+        gPluginName = pluginName;
```
AmandeepArora: South plugin interface code also writes gPluginName, so it would be over-written here?

pintomax: this variable gets the right value depending on the cpp file that sets it

PR #1635, opened 2019-08-07
Enable discovery and loading of notificationRule Python plugin

pintomax:
Enable discovery and loading of notificationRule Python plugin


AmandeepArora (approved):
Worked well for me when I was testing engine failure ML rule plugin.

MarkRiddoch (approved):

PR #1636, opened 2019-08-07
FOGL-3034 upgrade on installation package

ashish-jabble:
1) **Installation** category added in core startup and also as a child of **General** category
2) upgrade before package install added on the basis of  _upgradeOnInstall_ config item value
3) unit tests added and modified

@MarkRiddoch Should we consider this category as a reserved category? See FOGL-2014


MarkRiddoch:
> @MarkRiddoch Should we consider this category as a reserved category? See FOGL-2014

I don't think we need to do anything special here for delete.

We certainly need to start thinking about security and who has access to categories like this, but this is a much bigger subject that needs to be addressed.

PR #1639, opened 2019-08-12
FOGL-2583: Test added for fft filter south and threshold filter north

Singhal-Vaibhav:


praveen-garg:
@ashish-jabble foglamp_nightly_build_ub16/356/consoleFull 
I did the dev on ub18. We need to ensure to run this in system test stage for RPi buster and stretch etc. Please take care of that post merge.

PR #1651, opened 2019-08-20
FOGL-3026: plugin_reconfigure updates the content of handle

pintomax:
plugin_reconfigure updates the content of handle pointer, which is a Python Dict object initially set by plugin_init.

Similar change will be added for notification rule Python plugin.

This solution might also be applied to South Python plugins, removing the need of passing PLUGIN_HANDLE *


C/services/filter-plugin-interfaces/python/python_plugin_interface.cpp
```diff
@@ -120,21 +120,34 @@ static void filter_plugin_reconfigure_fn(PLUGIN_HANDLE handle,
...
-		handle = PyObject_Init(pReturn, Py_TYPE(pReturn));
-		// Remove old handle
-		Py_CLEAR(tmp);
-
-		Logger::getLogger()->debug("plugin_handle: plugin_reconfigure(): "
-					   "got updated handle from python plugin=%p, plugin '%s'",
-					   handle,
-					   gPluginName.c_str());
+		PyObject* tmp = (PyObject *)handle;
+		// Check current handle is Dict and pReturn is a Dict too
```
AmandeepArora: It is not a good idea to assume that handle is a python dict type object. We should not assume anything about its type or contents. It may very well be a class object for some plugin(s).

pintomax: The idea, is to force a Dict as the output of plugin_init and plugin_reconfigure.

The dict can be as complex as the user wants, also with other objects in it.

I think it's feasible.

pintomax: Just tested this:

class ComplexNumber:
    def __init__(self,r = 0,i = 0):
        self.real = r
        self.imag = i

    def getData(self):
        _LOGGER.info("{0}+{1}j".format(self.real,self.imag))


plugin_init

 global c1
    data['complex'] = c1
    return data


 plugin_triggers(handle):

 handle['complex'].getData()

Log:

Aug 20 04:12:53 ubuntu16-server-virt FogLAMP[18769] INFO: foo: module.name: 2+3j

pintomax: Also tested reconfigure:

plugin_reconfigure(handle, new_config):

saved_class = handle['complex']
new_handle['complex'] = saved_class
new_handle['complex'].real = 33
new_handle['complex'].imag = 17

Aug 20 04:34:53 ubuntu16-server-virt FogLAMP[19971] INFO: foo: module.name: >>> foo >>> Triggers = {'triggers': [{'asset': 'Random_B'}]}
Aug 20 04:34:53 ubuntu16-server-virt FogLAMP[19971] INFO: foo: module.name: 33+17j

MarkRiddoch: The API spec says it is a JSON object represented as a Python dict, so I think we are safe here.

The C++ interface specifies an opaque pointer, so indeed in that case we should assume nothing.

AmandeepArora: Other alternative approach is to make a copy of handle object in plugin_handle in shim layer and that (copy) is returned to C side. And at plugin_reconfigure time, we can just overwrite the handle (copy) variable in shim layer to new handle. So handle from C code perspective still remains the same, but its contents may have changed and underlying handle object could be allocated at a different memory location. All this happens transparently from C code's perspective and hence it won't need to assume python dict type.

PR #1656, opened 2019-08-21
FOGL-3164 : statistics collector is very CPU intensive on RPi Zero 

praveen-garg:
* FOGL-3164: limit cpu usage by statistics  python process

* FOGL-3164: cpu limit to 40

* FOGL-3164: added OS check


MarkRiddoch:
As a workaround I don't see any problem with this.

PR #1657, opened 2019-08-22
FOGL-3170: Add pipeline ready check

pintomax:
FOGL-3170: Add pipeline ready check


C/services/south/ingest.cpp
```diff
@@ -330,6 +330,14 @@ vector<Reading *>* newQ = new vector<Reading *>();
 		FilterPlugin *firstFilter = filterPipeline->getFirstFilterPlugin();
 		if (firstFilter)
 		{
+			// Check whether filters are set before calling ingest
+			while (!filterPipeline->isReady())
+			{
+				Logger::getLogger()->warn("Ingest called before "
+							  "filter pipeline is ready");
+				std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
praveen-garg: @MarkRiddoch @pintomax  would it not be better to give retries;  May be 3 consecutive attempts of 1*50 or 2*50 and 3*50 ms?
Or we have stressed enough that 100ms is always good.

pintomax:
Just created a new PR

PR #1658, opened 2019-08-22
FOGL-3170: Add pipeline ready check

pintomax:
FOGL-3170: Add pipeline ready check


praveen-garg:
I was expecting some retries; Looks we have stressed enough that 100ms is always good.

pintomax:
@praveen-garg log message and milliseconds sleep only applies to Raspberry Pi Zero.

In other platforms we don't see any message.

PR #1660, opened 2019-08-22
Fogl 3189: Minor improvements on 1.7.0

sstefano69:


sstefano69:
> Let's make it:
> 
> ```
> 
> 	if [[ $os_name == *"Raspbian"*  ]]; then
> 
> 		cpulimit -l 40 -- python3 -m foglamp.tasks.statistics "$@"
> 	else
> 		# Standard execution on other platforms
> 		python3 -m foglamp.tasks.statistics "$@"
> 	fi
> ```
> 
> i.e. remove the redundant complexity of Redhat checks and nested if

done, thanks for the good suggestion

sstefano69:
> Also remove line
> `os_version=`(grep -o '^VERSION_ID=.*' /etc/os-release | cut -f2 -d" | sed 's/"/`

done, thanks for the good suggestion

scripts/foglamp
```diff
@@ -485,7 +485,7 @@ foglamp_status() {
 
                 # Show Tasks
                 foglamp_log "info" "=== FogLAMP tasks:" "outonly" "pretty"
-                ps -ef | grep -o 'python3 -m foglamp.tasks.*' | grep -o 'foglamp.tasks.*' | grep -v 'foglamp.tasks\.\*' || true
+                ps -ef | grep -v 'cpulimit*' | grep -o 'python3 -m foglamp.tasks.*' | grep -o 'foglamp.tasks.*'    | grep -v 'foglamp.tasks\.\*' || true
```
praveen-garg: Are we sure, we want to invert-match?

sstefano69: yes, we want all the rows apart from the ones elated to cpulimit

praveen-garg: So we shall stop showing stats collector task  in foglamp status on Raspbian now? Would not it be better to `| sed 's/cpulimit -l 40 --//g'` and retain the current behavior!

scripts/tasks/statistics
```diff
@@ -18,11 +18,11 @@ fi
 cd "${FOGLAMP_ROOT}/python"
 
 os_name=`(grep -o '^NAME=.*' /etc/os-release | cut -f2 -d\" | sed 's/"//g')`
-os_version=`(grep -o '^VERSION_ID=.*' /etc/os-release | cut -f2 -d\" | sed 's/"//g')`
 
-if [[ ( $os_name == *"Red Hat"* || $os_name == *"CentOS"* ) &&  $os_version == *"7"* ]]; then
+if [[ $os_name == *"Raspbian"*  ]]; then
```
praveen-garg: With this, we can remove `cpulimit` from control files for aarch64 and x86_64 https://github.com/foglamp/foglamp-pkg/tree/develop/packages/Debian. 

Also I'm not sure, we should replace the check based on arch=`uname -m` Or arch=`arch`?!

PR #1665, opened 2019-08-23
1.7.0 chnage-log: typo fixes and removed redundant stuff

praveen-garg:


praveen-garg:
the script is a one time job that works when RC branch is cut off. 
And that also needs  review / fixes for proper arrangement according to fixVersion.

PR #1672, opened 2019-08-30
FOGL-3243 Ensure script type config item file content handling with CacheManager

ashish-jabble:
`Squashed all commits` as its candidate of 1.7.0RC

### DIFF
--- a/python/foglamp/common/configuration_manager.py
+++ b/python/foglamp/common/configuration_manager.py
@@ -74,6 +74,7 @@ class ConfigurationCache(object):
         self.miss += 1
         return False
 
+    # TODO: FOGL-3246 Add description
     def update(self, category_name, category_val, display_name=None):
         """Update the cache dictionary and remove the oldest item"""
         if category_name not in self.cache and len(self.cache) >= self.max_cache_size:
@@ -335,6 +336,7 @@ class ConfigurationManager(ConfigurationManagerSingleton):
                                               value=new_category_val, display_name=display_name).payload()
             result = await self._storage.insert_into_tbl("configuration", payload)
             response = result['response']
+            # TODO: FOGL-3246 Add description in cache
             self._cacheManager.update(category_name, new_category_val, display_name)
         except KeyError:
             raise ValueError(result['message'])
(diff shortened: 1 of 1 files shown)


pintomax:
@praveen-garg In my Pi3 Buster I still see the issue

pintomax:
Retested in a clean environment on Buster: ok

python/foglamp/common/configuration_manager.py
```diff
@@ -601,13 +601,16 @@ def get_entry_val(k):
...
+                category_value = self._handle_script_type(category_name, self._cacheManager.cache[category_name]['value'])
+                self._cacheManager.update(category_name, category_value, display_name=self._cacheManager.cache[category_name]['displayName'])
+                return category_value
 
             category_value = await self._read_category_val(category_name)
 
             if category_value is not None:
-                self._cacheManager.update(category_name, category_value)
                 category_value = self._handle_script_type(category_name, category_value)
+                self._cacheManager.update(category_name, category_value)
```
praveen-garg: this will force here to set `display_name = category_name`, we should test that and create a ticket to check `                 self._cacheManager.update(category_name, category_value)` calls and handle this; And also a story to handle category-> description as well.

python/foglamp/common/configuration_manager.py
```diff
@@ -635,9 +638,9 @@ def get_entry_val(k):
                 if cat_item is not None:
                     category_value = await self._read_category_val(category_name)
                     if category_value is not None:
-                        self._cacheManager.update(category_name, category_value)
-                        self._cacheManager.cache[category_name]['value'].update({item_name: cat_item})
                         category_value = self._handle_script_type(category_name, category_value)
+                        self._cacheManager.update(category_name, category_value)
```
praveen-garg: as above.

> this will force here to set display_name = category_name in cache, we should create a ticket to check self._cacheManager.update(category_name, category_value) calls and handle this; And also a story to handle category-> description as well. And add them here with FIXME and TODO

python/foglamp/common/configuration_manager.py
ashish-jabble: FOGL-3246

python/foglamp/common/configuration_manager.py
ashish-jabble: FOGL-3246

PR #1675, opened 2019-09-02
FOGL-3243 Ensure script type config item file content handling with CacheManager (1.7.0RC)

ashish-jabble:
ensure script type config item file content handling &

displayName fixes now retain after create category; as it is setting to NONE while caching on update & FOGL-3246 created

https://github.com/foglamp/FogLAMP/pull/1672

This will also solve problem mentioned in FOGL-3111


ashish-jabble:
cherry-picked from https://github.com/foglamp/FogLAMP/pull/1672

PR #1678, opened 2019-09-03
FOGL-3251: fix for filter setup at service creation time

pintomax:
FOGL-3251: fix for filter setup at service creation time


pintomax:
@shariq can you please update your python35 filter from

https://github.com/foglamp/foglamp-filter-python35/pull/40

PR #1683, opened 2019-09-05
FOGL-3194 test - default asset prefix of enviro-pHAT changed; check stats etc. as per the new value

praveen-garg:
https://github.com/foglamp/foglamp-south-envirophat/pull/32

### DIFF
--- a/tests/system/python/rpi/test_e2e_rpi_ephat.py
+++ b/tests/system/python/rpi/test_e2e_rpi_ephat.py
@@ -28,7 +28,7 @@ __version__ = "${VERSION}"
 SOUTH_PLUGIN = "envirophat"
 SVC_NAME = "Room-1"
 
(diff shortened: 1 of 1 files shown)


tests/system/python/rpi/test_e2e_rpi_ephat.py
```diff
@@ -28,7 +28,7 @@
 SOUTH_PLUGIN = "envirophat"
 SVC_NAME = "Room-1"
 
-ASSET_PREFIX = "envirophat/"  # default for envirophat South plugin
+ASSET_PREFIX = "e_"  # default for envirophat South plugin
```
ashish-jabble: This was only due to as we changed the default asset name value for plugin in https://github.com/foglamp/foglamp-south-envirophat/pull/32/files

PR #1688, opened 2019-09-06
FOGL-3179 upload script file attribute fixes when configuration run call back

ashish-jabble:


praveen-garg:
Tested and ran unit tests locally withOUT updating test-requirements; all looks good!

PR #1690, opened 2019-09-06
FOGL-3039 patch / FOGL-3089 HTTPInternalServerError instead of HTTPException and  response reason not None handling

praveen-garg:
* HTTPInternalServerError instead of HTTPException for http server error
* In case of HTTP Errors if response reason is not there - it should be asserted with empty string instead of `None`

**Note:** if there is valid StatusText / response.reason that should be asserted as raised.


praveen-garg:
https://github.com/foglamp/FogLAMP/pull/1614

praveen-garg:
see https://github.com/aio-libs/aiohttp/commit/f590bfd966e2dd440680cf7f1cb7c76886b41cb0

praveen-garg:
Ran multiple iterations of unit, api and system test with stretch, buster, ub16/18 and RHEL. looks good!

PR #1698, opened 2019-09-11
Fogl 3264: C south service - anticipated/moved lock

sstefano69:


praveen-garg:
@sstefano69 @pintomax  Please don't directly merge into 1.7.0RC at this stage.

PR #1701, opened 2019-09-12
FOGL-3264: anticipated/moved lock

pintomax:
FOGL-3264: anticipated/moved lock


pintomax:
@MarkRiddoch @sstefano69 This is a new PR based on develop

PR #1728, opened 2019-10-03
FOGL-3326 allow noninteractive debian install in lab tests install script 

praveen-garg:
**update upgrade has challenges!** Need more testing and investigation...


tests/system/python/packages/test_available_and_install_api.py
```diff
@@ -20,8 +20,9 @@
 __version__ = "${VERSION}"
 
 available_pkg = []
-# plugins installed by default 4
 counter = 4
+"""  By default 4 plugins are installed i.e. all north
+"""
```
ashish-jabble, line 25: Better to add `TODO: FOGL-3168`, once resolved we need to update the test as by default only 2 north plugins

ashish-jabble (approved):
LGTM!

tests/system/python/packages/test_available_and_install_api.py
praveen-garg: I would rather like to see it failing, when the real behavior changes. And check and act w.r.t. corresponding story/ tasks etc.

YashTatkondawar (approved):

PR #1729, opened 2019-10-03
FOGL-3026: add support for native Python filter plugins

pintomax:
FOGL-3026: enable discovery of native Python filter plugins


python/foglamp/common/plugin_discovery.py
```diff
@@ -41,11 +42,14 @@ def get_plugins_installed(cls, plugin_type=None, is_config=False):
...
             plugins_list.extend(plugins_list_south)
             plugins_list.extend(plugins_list_c_south)
+            plugins_list.extend(plugins_list_filter)
             plugins_list.extend(plugins_list_c_filter)
             plugins_list.extend(plugins_list_c_notify)
             plugins_list.extend(plugins_list_c_rule)
         elif plugin_type in ['filter', 'notificationDelivery', 'notificationRule']:
             plugins_list = cls.fetch_c_plugins_installed(plugin_type, is_config)
+            plugins_list_filter = cls.fetch_plugins_installed("filter", is_config)
+            plugins_list.extend(plugins_list_filter)
```
praveen-garg: @pintomax please check 
https://github.com/foglamp/FogLAMP/pull/1732/files#diff-84ad01acf8d68fab247cedc5084a7a16L49-R51

if plugin_type in request is `filter` which now are expected to discover python based  plugins too `else` part will take care of it. 

**Improved the the unit tests too for new scenario.**

```
tests.unit.python.foglamp.common.test_plugin_discovery.TestPluginDiscovery.test_get_c_notify_plugins_installed

tests.unit.python.foglamp.common.test_plugin_discovery.TestPluginDiscovery.test_get_c_rules_plugins_installed
```
 
these 2 were genuine failure, which helped to identify issue in the code itself!

python/foglamp/common/plugin_discovery.py
```diff
@@ -41,11 +42,14 @@ def get_plugins_installed(cls, plugin_type=None, is_config=False):
             plugins_list.extend(plugins_list_c_north)
             plugins_list.extend(plugins_list_south)
             plugins_list.extend(plugins_list_c_south)
+            plugins_list.extend(plugins_list_filter)
             plugins_list.extend(plugins_list_c_filter)
             plugins_list.extend(plugins_list_c_notify)
             plugins_list.extend(plugins_list_c_rule)
         elif plugin_type in ['filter', 'notificationDelivery', 'notificationRule']:
```
praveen-garg: this elif block is only for those types which have only C plugins.

python/foglamp/services/core/api/filters.py
```diff
@@ -74,8 +76,16 @@
         if category_info is not None:
             raise ValueError("This '{}' filter already exists".format(filter_name))
 
-        # Load C filter plugin info
-        loaded_plugin_info = apiutils.get_plugin_info(plugin_name, dir='filter')
+        # Load C/Python filter plugin info
+        #loaded_plugin_info = apiutils.get_plugin_info(plugin_name, dir='filter')
+        try:
+            # Try fetching Python filter
+            plugin_module_path = "{}/python/foglamp/plugins/filter/{}".format(_FOGLAMP_ROOT, plugin_name)
```
praveen-garg, line 83: @pintomax  https://github.com/foglamp/FogLAMP/commit/46a78659c7dbc91089bfc4dde236085d62bdcc32

Otherwise `foglamp/plugins/installed` API endpoint will result in errors with default install (and until python filter plugin is installed)! 

Make sure to run `tests/system/python/api/test_plugin_discovery.py` and actually
 * api and 
 * e2e system tests

python/foglamp/services/core/api/filters.py
```diff
@@ -74,8 +76,16 @@
         if category_info is not None:
             raise ValueError("This '{}' filter already exists".format(filter_name))
 
-        # Load C filter plugin info
-        loaded_plugin_info = apiutils.get_plugin_info(plugin_name, dir='filter')
+        # Load C/Python filter plugin info
+        #loaded_plugin_info = apiutils.get_plugin_info(plugin_name, dir='filter')
```
praveen-garg, line 80: remove these 2 commented statements!

pintomax:
Next commits will add support for multiple instances of native Python plugins (filters and notification rules)

C/services/common/plugin_manager.cpp
```diff
@@ -399,6 +400,10 @@ char		buf[MAXPATHLEN];
 	{
 		pluginHandle = new NotificationPythonPluginHandle(name.c_str(), buf);
 	}
+	else if (type.compare("filter") == 0)
```
MarkRiddoch: plugin_api.h has defines for the filter type strings, it would be nicer to use these.

MarkRiddoch (dismissed):
Just a minor point, otherwise ok

pintomax:
@AmandeepArora It would be good to test the foglamp-rule-ML-engine-failure with this branch

pintomax:
Just created a new repository:

https://github.com/foglamp/foglamp-filter-ema/tree/FOGL-3362

It's the EMA Python native filter plugin

C/services/common-plugin-interfaces/python/include/python_plugin_common_interface.h
```diff
@@ -10,18 +10,61 @@
...
+		string    m_name;
+		string    m_type;
+		PyThreadState*	m_tState;
+};
+
 extern "C" {
 // This is the map of Python object initialised in each 
 // South, Notification, Filter  plugin interfaces
-static map<string, PyObject*> *pythonModules = new map<string, PyObject *>();;
+static map<string, PythonModule*> *pythonModules = new map<string, PythonModule*>();
```
MarkRiddoch, line 60: Is it intended these are shared between all instances of the interface?

C/services/common-plugin-interfaces/python/include/python_plugin_common_interface.h
```diff
@@ -10,18 +10,61 @@
...
+// Map of PLUGIN_HANDLE objects, updated by plugin_init calls
+static map<PLUGIN_HANDLE, PythonModule*> *pythonHandles = new map<PLUGIN_HANDLE, PythonModule*>();
 
 // Global variable gPluginName set by PluginInterfaceInit:
 // it has a different memory address when set/read by
 // PluginInterfaceInit in South, Filter or Notification
+// Only used in plugin_info_fn calls
 static string gPluginName;
```
MarkRiddoch, line 68: Is it not possible to wrap this in a class? I'm concerned by the sharing semantics for these static variables.

If there really only should be one in any given process then I think wrapping them in a single class would make it much clearer. Maybe something like a PythonPluginManager class.

C/services/common-plugin-interfaces/python/include/python_plugin_common_interface.h
```diff
@@ -254,71 +363,201 @@ static PLUGIN_INFORMATION *plugin_info_fn()
...
+	// Acquire GIL
+	PyGILState_STATE state = PyGILState_Ensure();
+
+	// Import Python module using a new interpreter
+	if (loadModule || reloadModule)
 	{
-		// Failure
-		if (PyErr_Occurred())
+		// Start a new interpreter
+		newInterp = Py_NewInterpreter();
```
MarkRiddoch, line 468: Can this fail?

C/services/common-plugin-interfaces/python/include/python_plugin_common_interface.h
```diff
@@ -254,71 +363,201 @@ static PLUGIN_INFORMATION *plugin_info_fn()
...
+					   pName.c_str());
+
+		// Import Python script
+		PyObject *newObj = PyImport_ImportModule(name.c_str());
+
+		// Check result
+		if (newObj)
+		{
+			// Create a new PythonModule
+			PythonModule* newModule = new PythonModule(newObj,
```
MarkRiddoch: This could fail

C/services/filter-plugin-interfaces/python/python_plugin_interface.cpp
```diff
@@ -161,32 +170,44 @@ static void filter_plugin_reconfigure_fn(PLUGIN_HANDLE handle,
...
+					   "handle is NULL");
 		return;
 	}
 
-	auto it = pythonModules->find(gPluginName);
-	if (it == pythonModules->end() ||
+	if (!pythonHandles)
+	{
+		// Plugin name can not be logged here
+		Logger::getLogger()->error("pythonHandles map is NULL "
```
MarkRiddoch, line 183: Can we say something more meaningful to an end user here? Or is this a truly unexpected event?

C/services/filter-plugin-interfaces/python/python_plugin_interface.cpp
```diff
@@ -161,32 +170,44 @@ static void filter_plugin_reconfigure_fn(PLUGIN_HANDLE handle,
...
+
+	auto it = pythonHandles->find(handle);
+	if (it == pythonHandles->end() ||
 	    !it->second)
 	{
+		// Plugin name can not be logged here
 		Logger::getLogger()->fatal("plugin_handle: plugin_ingest(): "
-					   "pModule is NULL for plugin '%s'",
-					   gPluginName.c_str());
+					   "pModule is NULL");
```
MarkRiddoch, line 194: We seem to have reduce the information here, is the deliberate?

C/services/filter-plugin-interfaces/python/python_plugin_interface.cpp
```diff
@@ -259,62 +281,159 @@ PLUGIN_HANDLE filter_plugin_init_fn(ConfigCategory* config,
...
+	}
+
+	// Acquire GIL
+	PyGILState_STATE state = PyGILState_Ensure();
+
+	// Import Python module using a new interpreter
+	if (loadModule || reloadModule)
+	{
+		// Start a new interpreter
+		newInterp = Py_NewInterpreter();
```
MarkRiddoch, line 356: failure condition?

C/services/filter-plugin-interfaces/python/python_plugin_interface.cpp
```diff
@@ -259,62 +281,159 @@ PLUGIN_HANDLE filter_plugin_init_fn(ConfigCategory* config,
...
+					   string(reloadModule ? "re-" : "").c_str(),
+					   pName.c_str());
+
+		// Import Python script
+		PyObject *newObj = PyImport_ImportModule(name.c_str());
+
+		// Check for NULL
+		if (newObj)
+		{
+			PythonModule* newModule = new PythonModule(newObj,
```
MarkRiddoch: Failure condition?

MarkRiddoch (changes requested):
Since this has been a problematic area do you think you could add some documentation, i.e. comments, to explain the policy for creating interpreters, modules etc and when things are shared and when they are not.

Generally a few comments on failure conditions and making error messages more user meaningful if possible.

C/services/filter-plugin-interfaces/python/python_plugin_interface.cpp
pintomax: Plugin name cannot be logged if pythonHandles map is not available or handle does not exist un the map

C/services/filter-plugin-interfaces/python/python_plugin_interface.cpp
pintomax: Unexpected event

C/services/common-plugin-interfaces/python/include/python_plugin_common_interface.h
```diff
@@ -254,71 +363,224 @@ static PLUGIN_INFORMATION *plugin_info_fn()
...
-					   "in loaded python module '%s'",
-					   gPluginName.c_str());
+
+		// Plugin name not previously loaded: check current Python module
+		// pName is the key
+		auto it = pythonModules->find(pName);
+		if (it == pythonModules->end())
+		{
+			Logger::getLogger()->debug("plugin_handle: plugin_init(): "
+						   "pModule nof found for plugin '%s': "
```
AmandeepArora, line 422: typo : 'nof'

C/services/common-plugin-interfaces/python/include/python_plugin_common_interface.h
```diff
@@ -254,71 +363,224 @@ static PLUGIN_INFORMATION *plugin_info_fn()
...
+			// Set plugin type
+			PluginManager* pMgr = PluginManager::getInstance();	
+			PLUGIN_HANDLE tmp = pMgr->findPluginByName(pName);
+			if (tmp)
+			{
+				PLUGIN_INFORMATION* pInfo = pMgr->getInfo(tmp);
+				if (pInfo)
+				{
+					loadPluginType = string(pInfo->type);
+				}
```
AmandeepArora, line 435: else case?

C/services/common-plugin-interfaces/python/include/python_plugin_common_interface.h
```diff
@@ -254,71 +363,224 @@ static PLUGIN_INFORMATION *plugin_info_fn()
...
+			}
```
AmandeepArora, line 436: 'else' case is required?

C/services/common-plugin-interfaces/python/include/python_plugin_common_interface.h
```diff
@@ -418,19 +701,34 @@ static void plugin_reconfigure_fn(PLUGIN_HANDLE* handle,
...
+		if (pythonHandles)
+		{
+			// Remove current handle from the pythonHandles map
+			pythonHandles->erase(it);
+
+			// Add the handle to handles map as key, PythonModule object as value
+			std::pair<std::map<PLUGIN_HANDLE, PythonModule*>::iterator, bool> ret;
+			ret = pythonHandles->insert(pair<PLUGIN_HANDLE, PythonModule*>
+				((PLUGIN_HANDLE)*handle, currentModule));
+		}
```
AmandeepArora: 'else' case is required here? I think atleast a debug log.

AmandeepArora (commented):
Looks good. Some minor comments. Also I think many segments of code look quite similar, we should probably create some helper functions and call them rather than repeating these code segments.

praveen-garg:
Tested this with https://github.com/foglamp/foglamp-filter-ema :

1) Add sinusoid 
a) add ema in **disabled** mode 

sinusoid_graph

**expected:** sinusoid reads should not be affected.

b. warnings in syslog

```
Oct 16 12:15:00 raspberrypi s3[9358]: WARNING: FilterPluginInterface currently does not support 'plugin_start', plugin 'ema'
Oct 16 12:15:00 raspberrypi s3[9358]: WARNING: PythonPluginHandle::ResolveSymbol returning NULL for sym=plugin_start
Oct 16 12:15:00 raspberrypi s3[9358]: WARNING: FilterPluginInterface currently does not support 'plugin_start', plugin 'ema'
Oct 16 12:15:00 raspberrypi s3[9358]: WARNING: PythonPluginHandle::ResolveSymbol returning NULL for sym=plugin_start
3. Add ema in disable mode then enable, Getting FATAL in syslog
Oct 16 12:19:44 raspberrypi s3[9358]: ERROR: Called python script method plugin_reconfigure : error while getting result object, plugin 'ema'
Oct 16 12:19:44 raspberrypi s3[9358]: FATAL: logErrorMessage: Error 'NameError("name 'config' is not defined")', plugin 'ema'
```

2. enable ema now, FATAL errors in syslog **and filter does not apply**

```
Oct 16 12:19:44 raspberrypi s3[9358]: ERROR: Called python script method plugin_reconfigure : error while getting result object, plugin 'ema'
Oct 16 12:19:44 raspberrypi s3[9358]: FATAL: logErrorMessage: Error 'NameError("name 'config' is not defined")', plugin 'ema'
```

3. Add ema (fresh) in enabled mode, it works!  **but** 

a) disabling does not change anything. i.e. ema datapoint will contiune to be caluclated for sinusoid
b) same warnings as in 1(b)

**note**..   May be these fixes are needed are on filter side only or core side; or both!


**Reconfig of ema rate and datpoint & regression testing to follow...** /cc @YashTatkondawar

pintomax:
@praveen-garg warning messages 1.b) are always  logged. I'll change that

Following your steps on Ubuntu 16 right now

pintomax:
@praveen-garg When filter is added, even if not enabled, the readings are converted from C++ to Python and from Python to C++.

The latter conversion sets a new timestamp, so the sinusoid graph is not ok.

I'll check that conversion with @AmandeepArora

MarkRiddoch:
> The latter conversion sets a new timestamp, so the sinusoid graph is not ok.

We should never create a new timestamp for a reading once it has been set, regardless of the enabled or disabled state. The timestamp may be converted between languages, but it should always be the same timestamp.

pintomax:
There was just a missing check for "user_ts" in setReadingAttr()

pintomax:
@praveen-garg Could you please check again your above steps?

praveen-garg (approved):
lgtm! 

https://github.com/foglamp/foglamp-filter-ema/pull/4 for typo and config item order fixes for EMA plugin

AmandeepArora (approved):

MarkRiddoch (approved):

PR #1740, opened 2019-10-11
FOGL-3332: fix etc dir ownership doing sudo make install

sstefano69:
Signed-off-by: Stefano Simonelli <stefano@dianomic.com>


praveen-garg (changes requested):
I believe this should be relative to either a)`$FOGLAMP_DATA` if set or b) `$FOGLAMP_ROOT/data`. We should not copy it from source when `sudo make install`.

May be we should let it to the user to create and keep the file there manually, until we get an API feature to allow to upload `<x>.keytab` file.

The data/etc/kerberos/README.rst documentation needs to be updated in https://foglamp.readthedocs.io/en/develop/KERBEROS.html

i.e. `${FOGLAMP_ROOT}/data/etc/kerberos` or `${FOGLAMP_DATA}/etc/kerberos`


---
if the decision is to move forward the existing way; Cannot we put the fix for parent `$(DATA_SRC_DIR)` before  https://github.com/foglamp/FogLAMP/blob/develop/Makefile#L473 instead of creating a new target?

sstefano69:
the true permission problem is related to the dir data/etc  so I maintained its reference in the readme.

image

MarkRiddoch (dismissed):

Makefile
```diff
@@ -470,6 +470,11 @@ install_data : $(DATA_INSTALL_DIR) $(DATA_SRC_DIR)
 # directories must be set to be owned by the calling user.
 ifdef SUDO_USER
 ifeq ("$(USER)","root")
+
+	# generate_selfcertificate creates the etc directory owned by root if it doesn't exists
+	# the data and etc directories must be set to be owned by the calling user.
```
praveen-garg: We already have this comment for main ifdef block

praveen-garg (approved):

README.rst
```diff
@@ -196,3 +196,20 @@ This version of FogLAMP relies on SQLite to run. SQLite is embedded into the Sto
...
+FogLAMP version 1.7.0
+---------------------
+
+$FOGLAMP_ROOT/data/etc directory ownership
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The execution of the *sudo make install* immediately after *git clone* will create a *data/etc* directory owned by the *root* user,
+it should be owned by the user that will run FogLAMP, to fix it:
+::
+    chown -R <user>:<user> $FOGLAMP_ROOT/data
```
praveen-garg, line 212: May be this can simply be `chown -R $(whoami):$(whoami) $FOGLAMP_ROOT/data`

praveen-garg (dismissed):

README.rst
sstefano69: I have followed the same structure of the text above : 

```
If You Use PostgreSQL: Creating the Database Repository
=======================================================

This version of FogLAMP relies on SQLite to run. SQLite is embedded into the Storage service, but you may want to use PostgreSQL as a buffer and metadata storage (refer to the documentation on `ReadTheDocs <http://foglamp.readthedocs.io>`_ for more info. With a version of PostgreSQL installed via *apt-get* first you need to create a new database user with:
::
   sudo -u postgres createuser -d <user>

where *user* is the name of the Linux user that will run FogLAMP. The FogLAMP database user must have *createdb* privileges (i.e. the *-d* argument).
|br| |br|
```

Makefile
sstefano69: done

README.rst
praveen-garg: Okay. Consistency is a good thing to have. 👍

sstefano69: done

praveen-garg (approved):

PR #1741, opened 2019-10-14
FOGL-3347 update dependecies for aiohttp and cchardet

praveen-garg:
Signed-off-by: Praveen Garg <praveen.garg@nerdapplabs.com>


ashish-jabble (approved):
LGTM!! As per the CI results. More info available on JIRA itself.

PR #1745, opened 2019-10-16
FOGL 3331 Fixed the exit code status for lab test script

YashTatkondawar:


tests/system/lab/run_until_fails
```diff
@@ -20,12 +20,16 @@ fi
...
-	./reset
-	./test || break
+  echo "***************"
+  echo "***************"
+  echo "Run $i"
+  echo "***************"
+  echo "***************"
+  ./reset
+  ./test
+  if [[ $? -eq 1 ]]
```
praveen-garg: Better is to check `$? -ne 0` as anything other than a 0 status is undesirable.

praveen-garg (approved):

PR #1753, opened 2019-10-22
FOGL-3165 - Caching mechanism added for update/upgrade commands

ashish-jabble:
**Fixed items**
- [x] New configuration items added for the installation category with some optional attributes
- [x] fetch_available_packages
- [x] update
- [x] upgrade
- [x] unit tests
- [x] system tests - Checked on local environment. All good

**NOTE:**
1) fetch available caching added irrespective of the type, once the argument type is changed; cache refreshed
2) Also in fetch available package when update limit does not exceed for the day & available package cache is NOT expired yet. So on next call of this definition reset the get available cache and then do update->actual list package call and then do caching
3) Known issue: When caching is there; package log file is still created under data directory with same response on every hit.


ashish-jabble:
@MarkRiddoch Please have a look and also I have listed out the items in description itself.

MarkRiddoch (commented):
Looks good as far as it goes. The cache_clear is definitely needed. I would surround the cached function with another function that is what everything else calls. In the new function issue a cache_clear if it is more than 15 minutes since the last time you called cache_clear and then always call the cached function.

MarkRiddoch (dismissed):
Looks good to me

python/foglamp/services/core/server.py
```diff
@@ -292,12 +292,43 @@ class Server:
...
+            'maximum': '8',
+            'validity': 'upgradeOnInstall == "true"'
+        },
         'upgradeOnInstall': {
             'description': 'Run upgrade prior to installing new software',
             'type': 'boolean',
             'default': 'false',
-            'displayName': 'Upgrade on Install'
+            'displayName': 'Upgrade on Install',
+            'order': '3'
```
praveen-garg: swap the order with `maxUpgrade` as that depends on this.

python/foglamp/services/core/api/plugins/common.py
```diff
@@ -119,39 +120,81 @@ def load_and_fetch_c_hybrid_plugin_info(plugin_name: str, is_config: bool, plugi
...
+    else:
+        _get_available_packages.cache_clear()
+        pkg_cache_mgr['list']['last_accessed_time'] = ""
+
+    ret_code, available_packages = _get_available_packages(ret_code, tmp_log_output_fp, pkg_mgt, pkg_type)
 
     # combine above output in logs file
     with open("{}".format(stdout_file_path), 'a') as fh:
-        fh.write(" \n".join(log_output))
+        fh.write(" \n".join(available_packages))
```
praveen-garg, line 198: when `len(available_packages) is 0 && exit code is 0`, May be we should a) write a message  in log file `No plugin available to install` or b) don't create any and send link value as empty string in API response. (but in case b) there can be logs for update command)

billehunt (dismissed):
Looks good to me. I haven't tested extensively in potential failure cases, but general usage looks good.

python/foglamp/services/core/api/plugins/common.py
```diff
@@ -119,39 +120,84 @@ def load_and_fetch_c_hybrid_plugin_info(plugin_name: str, is_config: bool, plugi
...
+        if not len(available_packages):
+            fh.write("No package available to install")
```
praveen-garg, line 200: removed check and appending to available_packages; otherwise  _get_available_packages shall result API to return  {"plugins": ["No package available to install"], "link": "log/191025-16-47-43-list.log"}

praveen-garg (approved):

PR #1755, opened 2019-10-24
FOGL-2349: remove scripts dir under FOGLAMP_DATA

pintomax:
FOGL-2349: remove scripts dir under FOGLAMP_DATA


praveen-garg:
@mshariq-nerd  can you please check this resolves FOGL-2132?

praveen-garg:
@pintomax the scope of FOGL-2349 is to handle complete reset. 

We need to verify and remove all which was created post fresh install. there might be few more things created at runtime including the below:

- [ ] backup  
- [ ] core.err  
- [ ] etc and its child directories  content
- [x] extras  (it has fogbench only)
- [ ] logs  e.g. package logs 
- [x] scripts  
- [ ] support  
- [ ] var
- [ ] snapshots - plugins and tables

PR #1758, opened 2019-10-30
FOGL-3406 - As per FOGL-3061 multiple asset time bucket API support added

ashish-jabble:


shariq1501:
@MarkRiddoch , @ashish-jabble  It looks fine as per FOGL-3406 description. Able to get buckets for multiple assets but 

1.  We need to pass something for **{readings}** in cURL request to get api 
`curl -sX GET http://localhost:8081/foglamp/asset/{asset_code_1},{asset_code_2}/{reading}/bucket/{bucket_size}` response. 
it can be anything say **blah**,  but need to pass in the url otherwise it will not work.  
`curl -sX GET 
http://localhost:8081/foglamp/asset/system%2FcpuUsage_all,system%2FdiskTraffic_loop0/**blah**/bucket/30 | jq `
IMO it should not be mandatory to pass **reading** value because in case of multiple asset that is of no use. 

2. This implementation does not serve the purpose for GUI reading graph work because we have separate api calls for separate asset. Check
Screen Shot 2019-10-31 at 5 47 06 PM
In current GUI design we don't need to get all asset bucket in one go. We need endpoint without datapoint/{reading} for single asset to get all **datapoints** in single bucket for single asset
Screen Shot 2019-10-31 at 5 48 22 PM

check [https://scaledb.atlassian.net/browse/FOGL-3374](FOGL-3374).

Right now because of  mandatory **{reading}** param in api call we are not able to get all datapoints in a bucket
`curl -sX GET http://localhost:8081/foglamp/asset/{asset_code}/{reading}/bucket/{bucket_size}`

ashish-jabble:
@mshariq-nerd Added new REST API endpoint for multiple datapoints per asset. Kindly have a look.

MarkRiddoch (approved):
Looks good t one, thanks Ashish

shariq1501 (approved):

PR #1759, opened 2019-10-31
FOGL-3131 Fixes for north instance name reused after task deletion

ashish-jabble:


praveen-garg (dismissed):
Unable to reuse north instance name on upgrade after task deletion, schedule_id column is there but value is empty. 


This (module) add task and delete task is for north instance only. i..e process_name = 'north'  ||  process_name = north_c  if the schedule name is not unique, better to delete based on composite value of north instance name (i.e. schedule name) and process_name available in tasks table. And we can avoid all this upgrade / downgrade schema issues.

praveen-garg:
We already have checks for schedule name and category name. Hence no duplicate name can be found for process north and/ or north_c

     count = await check_schedules(storage, name)
     if count != 0:
          raise web.HTTPBadRequest(reason='A north instance with this name already exists')

ashish-jabble:
> Unable to reuse north instance name on upgrade after task deletion, schedule_id column is there but value is empty.

yeah. Do we need update sch_id for the rest of the entries? Is that possible?

> This (module) add task and delete task is for north instance only. i..e process_name = 'north' || process_name = north_c if the schedule name is not unique, better to delete based on composite value of north instance name (i.e. schedule name) and process_name available in tasks table. And we can avoid all this upgrade / downgrade schema issues

Please check https://github.com/foglamp/FogLAMP/compare/FOGL-3131_2?expand=1

But for upgrade user we still can't reuse north instance if someone has tried delete north instance in previous release and then try same name with new release it still says "Unable to reuse name, already used by a previous task"

praveen-garg:
> But for upgrade user we still can't reuse north instance if someone has tried delete north instance in previous release and then try same name with new release it still says "Unable to reuse name, already used by a previous task"

We should delete tasks for non-existent schedule name on upgrade for process name in north and north_c. 

Note: this is the case, when you upgraded after deleting an instance which had some tasks. the delete after upgrade should not be a problem in FOGL-3131_2 as we have in this branch.

praveen-garg:
> Do we need update sch_id for the rest of the entries? Is that possible?

we can find matching schedule id, based on schedule name and process_name; and update them into new table. If we go this long route.

MarkRiddoch (dismissed):
Thanks Ashish

praveen-garg (approved):

PR #1761, opened 2019-11-06
FOGL-3246 - description KV pair is added to configuration cache manager & other fixes

ashish-jabble:
Custom Job ran over [ub1804_with_poatgres](http://jenkins.dianomic.com:8080/view/custom%20jobs/job/foglamp_postgres_ub1804/70/) and don't see any regression with an existing unit & system tests. 4 tests are failing which is expected and due to FOGL-3125 with Postgres engine Only.

### DIFF
--- a/python/foglamp/common/configuration_manager.py
+++ b/python/foglamp/common/configuration_manager.py
@@ -367,9 +367,9 @@ class ConfigurationManager(ConfigurationManagerSingleton):
         # SELECT configuration.key, configuration.description, configuration.value, configuration.display_name, configuration.ts FROM configuration
         payload = PayloadBuilder().SELECT("key", "description", "value", "display_name", "ts") \
             .ALIAS("return", ("ts", 'timestamp')) \
-            .FORMAT("return", ("ts", "YYYY-MM-DD HH24:MI:SS.MS")).WHERE(["key", "=", cat_name]).payload()
+            .FORMAT("return", ("ts", "YYYY-MM-DD HH24:MI:SS.MS")).WHERE(["key", "=", cat_name]).LIMIT(1).payload()
         result = await self._storage.query_tbl_with_payload('configuration', payload)
-        return result['rows'] if result['rows'] else None
+        return result['rows'][0] if result['rows'] else None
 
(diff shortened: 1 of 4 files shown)


python/foglamp/common/configuration_manager.py
```diff
@@ -364,6 +363,14 @@ def get_entry_val(k):
             category_info.append((row['key'], row['description'], row["display_name"]))
         return category_info
 
+    async def _read_category(self, cat_name):
+        # SELECT configuration.key, configuration.description, configuration.value, configuration.display_name, configuration.ts FROM configuration
+        payload = PayloadBuilder().SELECT("key", "description", "value", "display_name", "ts") \
+            .ALIAS("return", ("ts", 'timestamp')) \
+            .FORMAT("return", ("ts", "YYYY-MM-DD HH24:MI:SS.MS")).WHERE(["key", "=", cat_name]).payload()
+        result = await self._storage.query_tbl_with_payload('configuration', payload)
+        return result['rows'] if result['rows'] else None
```
praveen-garg: I suggest to make it to avoid fetching `category[0]` every-time where this returned value is being used; & also add handling of keyError, if result is malformed.

python/foglamp/common/configuration_manager.py
```diff
@@ -364,6 +363,14 @@ def get_entry_val(k):
             category_info.append((row['key'], row['description'], row["display_name"]))
         return category_info
 
+    async def _read_category(self, cat_name):
+        # SELECT configuration.key, configuration.description, configuration.value, configuration.display_name, configuration.ts FROM configuration
+        payload = PayloadBuilder().SELECT("key", "description", "value", "display_name", "ts") \
+            .ALIAS("return", ("ts", 'timestamp')) \
+            .FORMAT("return", ("ts", "YYYY-MM-DD HH24:MI:SS.MS")).WHERE(["key", "=", cat_name]).payload()
```
praveen-garg: better to add LIMIT 1 to let query engine return as early as possible

praveen-garg (approved):

MarkRiddoch (approved):

PR #1766, opened 2019-11-14
FOGL-3387 Handling added on PUT configuration item as per optional attribute value

ashish-jabble:
- [x] length optional attribute is added in C-files as required; as it is NOT appearing without this change
- [x] Value handling with optional attribute added in set config item & bulk config item
i.e fixed for minimum, maximum, length, (mandatory, rule has already covered before)
- [x] unit tests added

Custom CI ub1804postgres report [here](http://jenkins.dianomic.com:8080/view/custom%20jobs/job/foglamp_postgres_ub1804/74/) against this branch. So Unit (4 audit unit tests are failing as expected on Postgres; already have a JIRA for the same)& system (api, e2e, smoke) tests are okay.

### DIFF
--- a/C/common/config_category.cpp
+++ b/C/common/config_category.cpp
@@ -570,6 +570,25 @@ string ConfigCategory::getDisplayName(const string& name) const
 }
 
 /**
+ * Return the length value of the configuration category item
+ *
+ * @param name	The name of the configuration item to return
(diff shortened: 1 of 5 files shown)


praveen-garg (approved):

MarkRiddoch (approved):
Looks good to me, thanks

PR #1771, opened 2019-11-21
FOGL-3482 fixed reset when dir is not available; i.e. make only

praveen-garg:
Signed-off-by: Praveen Garg <praveen.garg@nerdapplabs.com>


pintomax (approved):

ashish-jabble (approved):
As per custom Job CI results against this branch. All good now.

PR #1774, opened 2019-11-27
FOGL-3346_patch

ashish-jabble:
As there is No way to parameterized dynamic input values with fixture and [NOT a good practice](https://stackoverflow.com/questions/3145720/how-to-accumulate-state-across-tests-in-py-test/3145756#3145756). Hence I have used with below approach.

The original assertions are replaced by if statements which append messages to an errors list in case condition are not met. Then you assert the errors list is empty (an empty list is False) and make the assertion message contains each message of the errors list.

Check the package installation [CI-Report ](http://jenkins.dianomic.com:8080/view/custom%20jobs/job/foglamp_packages_installation_ub1804/106/testReport/) and you may see now it attempts for other plugins as well NOT stop on digiducer as its plugin discovery failed

### DIFF
--- a/tests/system/python/packages/test_available_and_install_api.py
+++ b/tests/system/python/packages/test_available_and_install_api.py
@@ -142,7 +142,7 @@ class TestPackages:
         if package_build_source_list.lower() == 'true':
             for pkg_name in available_pkg:
                 self._verify_and_install_package(foglamp_url, pkg_name)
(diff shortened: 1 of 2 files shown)


praveen-garg (approved):
LGTM!

YashTatkondawar (approved):

PR #1782, opened 2019-12-03
FOGL-3064 update API added for service package

ashish-jabble:
*Note*: on buster service endpoints are calling thrice. (NO clue yet)

### DIFF
--- a/python/foglamp/services/core/api/plugins/update.py
+++ b/python/foglamp/services/core/api/plugins/update.py
@@ -102,7 +102,7 @@ async def update_plugin(request: web.Request) -> web.Response:
                     if sch_info[0]['enabled'] == 't':
                         status, reason = await server.Server.scheduler.disable_schedule(uuid.UUID(sch_info[0]['id']))
                         if status:
-                            _logger.warning("{} {} instance is disabled as {} plugin is updating..".format(
+                            _logger.warning("Disabling {} {} instance, as {} plugin is updating..".format(
                                 p['service'], _type, name))
                             sch_list.append(sch_info[0]['id'])
 
@@ -146,8 +146,7 @@ async def _get_sch_id_and_enabled_by_name(name) -> list:
 def update_repo_sources_and_plugin(_type: str, name: str) -> tuple:
     # Below check is needed for python plugins
     # For Example: installed_plugin_dir=wind_turbine; package_name=wind-turbine
-    if "_" in name:
-        name = name.replace("_", "-")
+    name = name.replace("_", "-")
 
     # For endpoint curl -X GET http://localhost:8081/foglamp/plugins/available we used
     # sudo apt list command internal so package name always returns in lowercase,
@@ -172,7 +171,7 @@ def update_repo_sources_and_plugin(_type: str, name: str) -> tuple:
 
 
 def do_update(request):
-    _logger.info("{} plugin update starts...".format(request._name))
+    _logger.info("{} plugin update started...".format(request._name))
     code, link = update_repo_sources_and_plugin(request._type, request._name)
     if code != 0:
         _logger.error("{} plugin update failed. Logs available at {}".format(request._name, link))
@@ -181,9 +180,7 @@ def do_update(request):
         # PKGUP audit log entry
         storage_client = connect.get_storage_async()
         audit = AuditLogger(storage_client)
-        if "_" in request._name:
-            request._name = request._name.replace("_", "-")
-        audit_detail = {'packageName': "foglamp-{}-{}".format(request._type, request._name)}
+        audit_detail = {'packageName': "foglamp-{}-{}".format(request._type, request._name.replace("_", "-"))}
         asyncio.ensure_future(audit.information('PKGUP', audit_detail))
 
     # Restart the services which were disabled before plugin update
--- a/python/foglamp/services/core/api/plugins/update.py
+++ b/python/foglamp/services/core/api/plugins/update.py
@@ -102,7 +102,7 @@ async def update_plugin(request: web.Request) -> web.Response:
                     if sch_info[0]['enabled'] == 't':
                         status, reason = await server.Server.scheduler.disable_schedule(uuid.UUID(sch_info[0]['id']))
                         if status:
-                            _logger.warning("Disabling {} {} instance, as {} plugin is updating..".format(
+                            _logger.warning("Disabling {} {} instance, as {} plugin is updating...".format(
                                 p['service'], _type, name))
                             sch_list.append(sch_info[0]['id'])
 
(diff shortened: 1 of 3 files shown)


MarkRiddoch (dismissed):

python/foglamp/services/core/api/service.py
```diff
@@ -40,6 +41,7 @@
     | GET POST            | /foglamp/service                                      |
     | GET                 | /foglamp/service/available                            |
     | GET                 | /foglamp/service/installed                            |
+    | PUT                 | /foglamp/service/{type}/{name}/update                 |
```
praveen-garg, line 44: this can be `/foglamp/service/{name}/update`, We can identify type if type based on process name, and check with a set of ['notification]`

python/foglamp/services/core/api/service.py
```diff
@@ -398,11 +410,82 @@ def load_c_plugin(plugin: str, service_type: str) -> Dict:
...
-
+    services = get_service_installed()
     return web.json_response({"services": services})
+
+
+async def update_service(request: web.Request) -> web.Response:
+    """ update service
+
+    :Example:
+        curl -sX PUT http://localhost:8081/foglamp/service/Notification/notification/update
```
praveen-garg: Better to say Example:
`curl -sX PUT http://localhost:8081/foglamp/service/Notification/FL-Notifications/update`  or 
`curl -sX PUT http://localhost:8081/foglamp/service/FL-Notifications/update`

python/foglamp/services/core/api/service.py
```diff
@@ -398,11 +410,82 @@ def load_c_plugin(plugin: str, service_type: str) -> Dict:
...
+    """
+    _type = request.match_info.get('type', None)
+    name = request.match_info.get('name', None)
+    try:
+        _type = _type.lower()
+        if _type != 'notification':
+            raise ValueError("Invalid plugin type. Must be 'notification'")
```
praveen-garg: `Invalid service type. Should be 'notification'"`

python/foglamp/services/core/api/service.py
```diff
@@ -398,11 +410,82 @@ def load_c_plugin(plugin: str, service_type: str) -> Dict:
...
+
+        # Check requested service name is installed or not
+        installed_services = get_service_installed()
+        if name not in installed_services:
+            raise KeyError("{} service is not yet installed. So update is not possible.".format(name))
```
praveen-garg: `{} service is not installed yet. Hence update is not possible.`

python/foglamp/services/core/api/service.py
```diff
@@ -398,11 +410,82 @@ def load_c_plugin(plugin: str, service_type: str) -> Dict:
...
+
+        storage_client = connect.get_storage_async()
+        # TODO: process_name ends with "_c" suffix
+        payload = PayloadBuilder().SELECT("id", "enabled", "schedule_name").WHERE(['process_name', '=', '{}_c'.format(
+            _type)]).payload()
+        result = await storage_client.query_tbl_with_payload('schedules', payload)
+        sch_info = result['rows']
```
praveen-garg, line 440: what if there multiple schedules with prcoess_name = 'notification_c'? move `schedule_name` to WHERE clause And if no rows then service exists but not added (hence no disable required)

python/foglamp/services/core/api/service.py
```diff
@@ -398,11 +410,82 @@ def load_c_plugin(plugin: str, service_type: str) -> Dict:
...
+        sch_list = []
+        if sch_info and sch_info[0]['enabled'] == 't':
+            status, reason = await server.Server.scheduler.disable_schedule(uuid.UUID(sch_info[0]['id']))
+            if status:
+                _logger.warning("{} service is disabled as {} service is updating..".format(
```
praveen-garg: I suggest `Disabled schedule for {}, as {} service of type {} is being updated...`
Don;t we need audit log?

python/foglamp/services/core/api/service.py
```diff
@@ -398,11 +410,82 @@ def load_c_plugin(plugin: str, service_type: str) -> Dict:
...
+        raise web.HTTPBadRequest(reason=str(ex))
+    except Exception as ex:
+        raise web.HTTPInternalServerError(reason=str(ex))
+
+    return web.json_response({"message": "{} service update in process. Wait for few minutes to complete.".format(
+        name)})
+
+
+def do_update(request):
+    _logger.info("{} service update starts...".format(request._name))
```
praveen-garg: `starts...` > `started...`

python/foglamp/services/core/api/service.py
```diff
@@ -398,11 +410,82 @@ def load_c_plugin(plugin: str, service_type: str) -> Dict:
...
+    name = "foglamp-service-{}".format(request._name.lower())
+    _platform = platform.platform()
+    stdout_file_path = common.create_log_file(action="update", plugin_name=name)
```
praveen-garg: if this accepts only 2 args, avoid key-worded args. especially as  `plugin_name` is there, this log action is for.

praveen-garg (changes requested):

python/foglamp/services/core/api/service.py
ashish-jabble: name -> Package name (`foglamp-service-notification`) NOT a notification instance. Hence only `notification` value will be considered

praveen-garg: Sorry, I got confused! `foglamp/service/notification/notification` is good to go as /type/name

python/foglamp/services/core/api/service.py
praveen-garg: Ignore this!

YashTatkondawar (dismissed):

python/foglamp/services/core/api/plugins/update.py
```diff
@@ -102,7 +102,7 @@
                     if sch_info[0]['enabled'] == 't':
                         status, reason = await server.Server.scheduler.disable_schedule(uuid.UUID(sch_info[0]['id']))
                         if status:
-                            _logger.warning("{} {} instance is disabled as {} plugin is updating..".format(
+                            _logger.warning("Disabling {} {} instance, as {} plugin is updating..".format(
```
praveen-garg: `updating..` should be `updating...`

python/foglamp/services/core/api/service.py
praveen-garg: this is not a plugin but service

praveen-garg (changes requested):

praveen-garg (approved):

PR #1783, opened 2019-12-04
FOGL-3064 notification plugins API update added

ashish-jabble:
Signed-off-by: ashish-jabble <ashish@dianomic.com>

### DIFF
--- a/python/foglamp/services/core/api/plugins/update.py
+++ b/python/foglamp/services/core/api/plugins/update.py
@@ -82,7 +82,7 @@ async def update_plugin(request: web.Request) -> web.Response:
                             notification_name, name, _type))
                         await config_mgr.set_category_item_value_entry(notification_name, "enable", "false")
                         notification_list.append(notification_name)
-            _type = "notify"
+            _type = "notify" if _type == 'notificationDelivery' else "rule"
         else:
             # Tracked plugins from asset tracker
             tracked_plugins = await _get_plugin_and_sch_name_from_asset_tracker(_type)
--- a/python/foglamp/services/core/api/plugins/update.py
+++ b/python/foglamp/services/core/api/plugins/update.py
@@ -58,7 +58,7 @@ async def update_plugin(request: web.Request) -> web.Response:
             raise KeyError("{} plugin is not yet installed. So update is not possible.".format(name))
 
         sch_list = []
-        notify_list = []
+        notification_list = []
         if _type in ['notificationDelivery', 'notificationRule']:
             # Check Notification service is enabled or not
             payload = PayloadBuilder().SELECT("id", "enabled", "schedule_name").WHERE(['process_name', '=',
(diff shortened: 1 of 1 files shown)


MarkRiddoch (dismissed):

python/foglamp/services/core/api/plugins/update.py
```diff
@@ -39,48 +40,76 @@
...
+                config_mgr = ConfigurationManager(storage_client)
+                all_notifications = await config_mgr._read_all_child_category_names("Notifications")
+                for notification in all_notifications:
+                    notification_config = await config_mgr._read_category_val(notification['child'])
+                    notification_name = notification_config['name']['value']
+                    channel = notification_config['channel']['value']
+                    rule = notification_config['rule']['value']
+                    is_enabled = True if notification_config['enable']['value'] == 'true' else False
+                    if (channel == name and is_enabled) or (rule == name and is_enabled):
+                        _logger.warning("{} notification instance is disabled as {} {} plugin is updating..".format(
```
praveen-garg: `Disabling {} notification instance, as {} {} plugin is updating...`

python/foglamp/services/core/api/plugins/update.py
```diff
@@ -39,48 +40,76 @@
...
 
-        # Tracked plugins from asset tracker
-        tracked_plugins = await _get_plugin_and_sch_name_from_asset_tracker(_type)
         sch_list = []
-        filters_used_by = []
-        if _type == 'filter':
-            # In case of filter, for asset_tracker table we are inserting filter category_name in plugin column
-            # instead of filter plugin name by Design
-            # Hence below query is required to get actual plugin name from filters table
+        notify_list = []
```
praveen-garg: this should be `notification_instances_list`

python/foglamp/services/core/api/plugins/update.py
```diff
@@ -39,48 +40,76 @@
...
+        if _type in ['notificationDelivery', 'notificationRule']:
+            # Check Notification service is enabled or not
+            payload = PayloadBuilder().SELECT("id", "enabled", "schedule_name").WHERE(['process_name', '=',
```
praveen-garg, line 65: can not there be multiple schedules of notification_c, (may be disabled)? This is important to fix as we are checking sch_info[0].

praveen-garg (changes requested):

python/foglamp/services/core/api/plugins/update.py
ashish-jabble: Nope, we have added a restriction of notification service with one instance only when hit  add_service endpoint

praveen-garg: Okay

praveen-garg (approved):
lgtm!

PR #1790, opened 2019-12-06
FOGL-2684: Added return statements

YashTatkondawar:
Test Result: <base_uri>/nightly%20jobs/job/foglamp_nightly_build_ub16/460/


ashish-jabble (commented):
Looks good. It would be nice to have custom nightly CI Job against it. So that we are sure no regression introduce from it.

ashish-jabble (approved):

PR #1791, opened 2019-12-09
FOGL-3519: decrease default start by the user provided length

pintomax:
FOGL-3519: decrease default start by the user provided length

curl -s 'http://127.0.0.1:8081/foglamp/asset/sinusoid/bucket/1?length=600'

now returns more than 60 results.

If the change is OK we could add a similar fix to another API  call:

curl -sX GET 'http://localhost:8081/foglamp/asset/{asset_code}/{reading}/bucket/{bucket_size}?start=<start point>&length=<length>'


pintomax:
Branch needs approval

billehunt (approved):

shariq1501 (approved):

PR #1792, opened 2019-12-09
FOGL-2718 - Error in syslog after south service deletion and re-creation 

ashish-jabble:
Custom Job ran `<BASE_URI>/nightly%20jobs/job/foglamp_nightly_build_ub18/347/`

**Result:** No regression as such found for our unit & system tests with this change


praveen-garg (approved):
LGTM! thanks for deep investigation on this.

MarkRiddoch (approved):

PR #1794, opened 2019-12-11
FOGL-3530: decrease default start by the user provided length

pintomax:
FOGL-3530: decrease default start by the user provided length

Also removed a query that is time consuming and it's useless as there could be empty result set with provided parameters


pintomax:
@mshariq-nerd You can test this branch even if the unit tests are failing

shariq1501 (dismissed):

shariq1501 (dismissed):

tests/unit/python/foglamp/services/core/api/test_browser_assets.py
```diff
@@ -317,6 +317,7 @@ def q_result(*args):
             # FIXME: ordering issue and add tests for datetimeunits request param
             # assert '{"aggregate": [{"operation": "min", "json": {"column": "reading", "properties": "humidity"}, "alias": "min"}, {"operation": "max", "json": {"column": "reading", "properties": "humidity"}, "alias": "max"}, {"operation": "avg", "json": {"column": "reading", "properties": "humidity"}, "alias": "average"}], "where": {"column": "asset_code", "condition": "=", "value": "fogbench_humidity"}, "limit": 20}' in args1
 
+    @pytest.mark.skip(reason='Test skip')
```
praveen-garg: As these tests needs a rewrite once api is mature enough for the requirements. We should create a JIRA for this and `@pytest.mark.skip(reason='TODO: FOGL-xxx summary of JIRA')` for all these bucket api tests.

pintomax: FOGL-3541 created

python/foglamp/services/core/api/browser.py
```diff
@@ -458,12 +458,7 @@ def where_clause(request, where):
         start = ts - length
         asset_code_list = asset_code.split(',')
         _readings = connect.get_readings_async()
-        for code in asset_code_list:
```
praveen-garg, line 461: We need to find a way for validation. Looks we are giving an empty set now if an invalid assetName or assetName/assetDatapoint or assetAttribute is supplied in request. Ideally it should be 404.

praveen-garg (approved):

PR #1804, opened 2020-01-03
FOGL-3578: System test for foglamp-north-gcp gateway 

YashTatkondawar:


tests/system/python/packages/test_gcp_gateway.py
```diff
@@ -0,0 +1,134 @@
...
+import subprocess
+import http.client
+import json
+import time
+import pytest
+from pathlib import Path
+import utils
+
+__author__ = "Yash Tatkondawar"
+__copyright__ = "Copyright (c) 2019 Dianomic Systems Inc."
```
MarkRiddoch: Should be 2020!

tests/system/python/packages/test_gcp_gateway.py
```diff
@@ -0,0 +1,134 @@
...
+
+        copy_certs()
+
+        payload = {"name": task_name,
+                   "plugin": "{}".format(north_plugin),
+                   "type": "north",
+                   "schedule_type": 3,
+                   "schedule_repeat": 5,
+                   "schedule_enabled": True,
+                   "config": {"project_id": {"value": "foglamp"},
```
MarkRiddoch: We should probably have some notes at least as to the GCP setup we are assuming and the certs etc.

MarkRiddoch (changes requested):
There is no check that data actually gets to GCP, only that we have statistics to say we sent it. Can we investigate if there is a simple way to retrieve a statistic count or something of the kind from GCP itself so that we can be more confident that the data was seen by GCP.

tests/system/python/packages/test_gcp_gateway.py
YashTatkondawar: I have added this as [prerequisite](https://github.com/foglamp/FogLAMP/pull/1804/files#diff-2cae579d734ab10093cc71381e01aff2R9). Do you want me to add anywhere else?

MarkRiddoch (dismissed):

tests/system/python/packages/test_gcp_gateway.py
```diff
@@ -0,0 +1,209 @@
...
+        assert False, "remove package script failed!"
+
+    try:
+        subprocess.run(["cd {}/tests/system/python/scripts/package/ && ./setup {}"
+                       .format(PROJECT_ROOT, package_build_version)], shell=True, check=True)
+    except subprocess.CalledProcessError:
+        assert False, "setup package script failed"
+
+    try:
+        subprocess.run(["sudo apt install -y foglamp-north-gcp-gateway foglamp-south-sinusoid"], shell=True, check=True)
```
praveen-garg: We should make it `foglamp-north-gcp` from `nightly/buster/armv7l/` (atleast as long as we have it with foglamp/ org; and then think again once its migrated....)

praveen-garg (dismissed):

praveen-garg (approved):

MarkRiddoch (approved):

PR #1805, opened 2020-01-06
FOGL-3511

ashish-jabble:
**Fixed items**
- [x] south-phidget references are removed from system tests
- [x] foglamp-sqlite-pkg package download from dianomic org in requirements.sh for RPM based platforms build 
- [x] filter-rms-trigger references are removed from system tests
- [x] filter-simple-python references are removed from system tests
- [x] filter-ema references are removed from system tests

**Note** - @MarkRiddoch What is the plan for the backup of deleted system tests? May be in plugin repo itself under path like `{Repo_Name}/tests/system/...`


YashTatkondawar (approved):
I have tested the changes with regards to this PR:
1. Tested the requirements.sh on CentOS platform.
2. Tested the test_lab.py.
3. Tested the test_filters.py with updated code.
4. Tested test_available_and_install_api.py with updated package_list.json file.

After testing, this changes looks good to me.

MarkRiddoch (approved):

PR #1806, opened 2020-01-07
FOGL-3553 Added new configuration item type northTask

rajeshkumar426:
Signed-off-by: rajeshkumar426 <rajeshkumar426@outlook.com>

- modification in unit testcases

- Added new type in configuration manager


tests/unit/python/foglamp/common/test_configuration_manager.py
```diff
@@ -2509,13 +2511,26 @@ def mock_coro():
...
+    @pytest.mark.parametrize("input, expected", [
+        ('northTask', 'true'),
+        ('north', 'false'),
+        ('bool', 'false'),
+        ('boolean', 'true')
+    ])
+    async def test_valid_type(self, input, expected):
+        storage_client_mock = MagicMock(spec=StorageClientAsync)
+        c_mgr = ConfigurationManager(storage_client_mock)
+        assert expected == c_mgr._validate_type(input)
```
ashish-jabble: There is NO _validate_type definition in configuration manager. Remove this test as it is NOT required

python/foglamp/common/configuration_manager.py
```diff
@@ -1361,6 +1361,8 @@ def _str_to_ipaddress(item_val):
                 return False
         elif _type == 'string':
             return isinstance(_value, str)
+        elif _type == "northTask":
+            return isinstance(_value, str)
```
ashish-jabble, line 1363: Seems like it is same as of **string** type. Though I am not sure about its  data validation w.r.t doc i.e `allow for items to contain the name of a north task configuration category`

tests/unit/python/foglamp/common/test_configuration_manager.py
```diff
@@ -2541,7 +2556,9 @@ def mock_coro():
         ("JSON", None),
         ("URL", "blah"),
         ("URL", "example.com"),
-        ("URL", "123:80")
+        ("URL", "123:80"),
+        ("northTask", 123),
+        ("northTask", 1_1)
```
ashish-jabble: Syntactical error

ashish-jabble (changes requested):

MarkRiddoch (commented):
Nothing to add to Ashish's comments. Will approve when these are dealt with.

python/foglamp/common/configuration_manager.py
rajeshkumar426: As per discussed with @MarkRiddoch `Correct, the main difference is that it is a hint to the GUI that this is a pick list of north task names. I don't think we can easily tell if the entry is that of a task, which would be a good check to be able to make. However it we can't easily do that then I would skip this for the time.`

ashish-jabble (approved):

MarkRiddoch (approved):

PR #1809, opened 2020-01-09
FOGL-3602 Always add asset tracker data

MarkRiddoch:
Signed-off-by: Mark Riddoch <mark@dianomic.com>


ashish-jabble (approved):
Now system tests are good with the change.

PR #1825, opened 2020-01-31
FOGL-3688: Updated the system test based on changes in FOGL-3673

YashTatkondawar:


tests/system/python/conftest.py
```diff
@@ -285,12 +283,98 @@ def _read_data_from_pi(host, admin, password, pi_database, asset, sensor):
...
+                        if el["Name"] == _head:
+                            elx = el["Items"]
+                            for _el in elx:
+                                _recoded_value_list.append(_el["Value"])
+                            _data_pi[_head] = _recoded_value_list
+
+                # Delete recorded elements
+#                conn.request("DELETE", '/piwebapi/elements/{}'.format(web_id), headers=headers)
+#                res = conn.getresponse()
+#                res.read()
```
praveen-garg: Has this been kept commented to verify on CI?

tests/system/python/e2e/test_e2e_coap_pi_web_api.py
```diff
@@ -53,7 +53,7 @@ def _verify_egress(read_data_from_pi, pi_host, pi_admin, pi_passwd, pi_db, wait_
     data_from_pi = None
 
     # See C/plugins/common/omf.cpp
-    af_hierarchy_level = "foglamp_data_piwebapi"
+    af_hierarchy_level = "data_piwebapi"
```
praveen-garg: `foglamp/data_piwebapi` and treat the same way as for tests/system/python/packages/test_pi_webapi.py

tests/system/python/conftest.py
```diff
@@ -285,12 +283,98 @@ def _read_data_from_pi(host, admin, password, pi_database, asset, sensor):
...
+                r = json.loads(res.read().decode())
+                for el in r["Items"]:
+                    if el["Name"] == pi_database:
+                        elements = el["Links"]["Elements"]
+
+            if elements is not None:
+                conn.request("GET", elements, headers=headers)
+                res = conn.getresponse()
+                r = json.loads(res.read().decode())
+                url_elements_list = r["Items"][0]["Links"]["Elements"]
```
praveen-garg: I would suggest to make it like:

               items = r["Items"]
               url_elements_list = None
               for el in items:
                    if el["Name"] == af_hierarchy_list[0]:
                        url_elements_list = el["Links"]["Elements"]

praveen-garg: And we SHOULD put a  multi-line comment, if hierarchy is `foglamp/tenant1/data_piwebapi` then we need to keep finding `["Links"]["Elements"]` recursively for `foglamp` then `tenant1`, and next code block will read last node `data_piwebapi` Elements link.

praveen-garg (changes requested):

praveen-garg (approved):
lgtm!

PR #1834, opened 2020-02-12
FOGL-3718 Added API endpoint to remove plugin from FogLAMP

rajeshkumar426:
Signed-off-by: rajeshkumar426 <rajeshkumar426@outlook.com>

### DIFF
--- a/python/foglamp/services/core/api/plugins/remove.py
+++ b/python/foglamp/services/core/api/plugins/remove.py
@@ -1,0 +1,78 @@
+# -*- coding: utf-8 -*-
+
+# FOGLAMP_BEGIN
+# See: http://foglamp.readthedocs.io/
+# FOGLAMP_END
+
+import platform
+import os
+from aiohttp import web
+import logging
+from foglamp.common import logger
+from foglamp.common.plugin_discovery import PluginDiscovery
+from foglamp.services.core.api.plugins import common
+
+__author__ = "Rajesh Kumar"
+__copyright__ = "Copyright (c) 2019, Dianomic Systems Inc."
+__license__ = "Apache 2.0"
+__version__ = "${VERSION}"
+
+_logger = logger.setup(__name__, level=logging.INFO)
+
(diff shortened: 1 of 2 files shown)


python/foglamp/services/core/api/plugins/remove.py
```diff
@@ -0,0 +1,78 @@
...
+
+async def plugin_delete(request):
+    """
+    Remove plugin from foglamp
+
+    '''
+    EndPoint: curl -X DELETE http://host-ip:port/foglamp/plugins/{type}/{name}/remove
+    '''
+    Example:
+        curl -X DELETE http://host-ip:port/foglamp/plugins/south/sinusoid/remove
```
MarkRiddoch: There is no need to have the /remove on the end of the URL, this is a DELETE operation

MarkRiddoch (changes requested):

python/foglamp/services/core/routes.py
```diff
@@ -164,11 +165,12 @@ def setup(app):
     app.router.add_route('GET', '/foglamp/package/log', package_log.get_logs)
     app.router.add_route('GET', '/foglamp/package/log/{name}', package_log.get_log_by_name)
 
-    # Plugins (install, discovery, update)
+    # Plugins (install, discovery, update, delete)
     app.router.add_route('GET', '/foglamp/plugins/installed', plugins_discovery.get_plugins_installed)
     app.router.add_route('GET', '/foglamp/plugins/available', plugins_discovery.get_plugins_available)
     app.router.add_route('POST', '/foglamp/plugins', plugins_install.add_plugin)
     app.router.add_route('PUT', '/foglamp/plugins/{type}/{name}/update', plugins_update.update_plugin)
+    app.router.add_route('DELETE', '/foglamp/plugins/{type}/{name}/remove', remove.plugin_delete)
```
pintomax: We don't need "/remove" at the end of URL

pintomax (changes requested):

PR #1835, opened 2020-02-12
Fix memory leaks in statistics handling

MarkRiddoch:
FOGL-3694: adds EDS compatibility


pintomax (dismissed):

C/services/south/ingest.cpp
```diff
@@ -86,6 +86,7 @@ int Ingest::createStatsDbEntry(const string& assetName)
 				return -1;
```
AmandeepArora, line 87: I think result should be freed here also.

AmandeepArora (dismissed):
One minor comment, looks good otherwise.

C/services/south/ingest.cpp
MarkRiddoch: Yes, thanks for picking that additional one up

AmandeepArora (approved):

pintomax (approved):

ashish-jabble:
@MarkRiddoch Anything pending on the pull request? fyi, I ran a custom CI JOB with the branch on platforms (ub1804, RPI with buster) and I couldn't find any regression with it. I think we can merge this PR under foglamp organization.

